PR #50, opened 2021-07-09
feat(pipettes): control motor via spi using uart

Laura-Danielle:
## Overview
This is a WIP. When this PR is complete, you will be able to send commands over serial via uart to control the motor.

Once this PR is complete, we can discuss a better layout for the entire ot3 project and that should be implemented immediately after this so that we can more easily split up work.

## TODO
1. Return the status/data via uart
2. figure out the pulse per revolution to move the motor at and how to sync the clocks
3. Make sure that you are static casting the enum class of motor registers
4. Test w/ a motor 
5. Add tests


gantry/firmware/main.cpp
```diff
@@ -66,7 +66,7 @@ auto main() -> int {
     /** Initializes the peripherals clocks
      */
     RCC_Peripheral_Clock_Select();
-    MX_GPIOA_Init();
+    MX_SPI2_Init();
 
     //    xTaskCreateStatic(uartTask, "USART Task", stack.size(), nullptr, 1,
```
sfoster1, line 71: Let's take the moment to clean up some commented out code

include/firmware/common/spi_comms.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include <span>
+#include "stm32g4xx_hal_conf.h"
+
+namespace spi {
+    class Spi {
+      public:
+        static constexpr auto BUFFER_SIZE = 5;
+        using BufferType = std::array<uint8_t, BUFFER_SIZE>;
+        Spi();
+        void send_command(BufferType aTxBuffer, uint32_t& data, uint8_t& status);
```
sfoster1: This almost certainly wants to either take `aTxBuffer` by reference, e.g. `BufferType& aTxBuffer`, or take a view like a `span`. Defining this as a pass-by-value requires copying the entirety of the buffer every call.

include/pipettes/core/communication.hpp
```diff
@@ -41,6 +41,15 @@ auto MessageReader::read(Reader &communication)
         case static_cast<uint32_t>(pipette_messages::MessageType::stop):
             r = pipette_messages::Stop{};
             break;
+        case static_cast<uint32_t>(pipette_messages::MessageType::setup):
```
sfoster1, line 44: we should consider changing this code so that each switch case `return`s. while it can sometimes be good practice to maintain a single point of return, in a case like this removing the return from line 70 and having it in each switch case means compilation will fail if you forget one, whereas right now everything will work until runtime if you forget an `r =...` line.

include/pipettes/tests/test_motor_control.hpp
```diff
@@ -0,0 +1,71 @@
+#pragma once
+
+#include <array>
+#include <cstdint>
+#include <iostream>
+#include <span>
+
+#include "common/core/bit_utils.hpp"
+//#include "pipettes/firmware/motor_control.hpp"
```
sfoster1, line 9: commented code

include/pipettes/tests/test_motor_control.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+/*
+ * Motor Control and namespace.
```
sfoster1, line 12: kind of a confusing block comment - maybe mostly just mention this is a test mock for the real thing?

pipettes/firmware/motor_control.cpp
```diff
@@ -0,0 +1,95 @@
...
+ * get_status, get the status of the GSTAT register
+ */
+
+/*
+ * Private Functions
+ */
+void MotorControl::build_command(uint8_t command, const uint32_t& command_data,
+                                 BufferType& txBuffer) {
+    // need to pass in data parameter and use int_to_bytes here
+    auto* output = txBuffer.begin();
```
sfoster1, line 37: not sure if this is from lint, but `auto*` is appropriate when you're declaring something that _is_ a pointer, not just has pointer semantics. Sometimes it will be - if `BufferType` is an array - but sometimes it won't, sometimes it will just be an iterator, and then `auto*` will be incorrect. We should make it just `auto`.

pipettes/firmware/motor_control.cpp
```diff
@@ -0,0 +1,95 @@
...
+ * set_speed, sets a global variable called speed
+ * get_speed, gets a global variable called speed
+ * move, moves the motor for an unlimited amount of time
+ * get_status, get the status of the GSTAT register
+ */
+
+/*
+ * Private Functions
+ */
+void MotorControl::build_command(uint8_t command, const uint32_t& command_data,
```
sfoster1, line 34: are commands always 5 bytes? if they can ever be different sizes we should probably have this return the output iterator.

pipettes/firmware/motor_control.cpp
```diff
@@ -0,0 +1,95 @@
...
+ * Public Functions
+ */
+
+void MotorControl::move() {
+    // Here we need to sync the clock line and pulse
+    // still not 100% how to do that.
+
+    Set_Enable_Pin();
+    Set_Direction();
+    while (1) {
```
sfoster1, line 56: Uh can we make this maybe some big number of iterations bit uncomfortable with it infinite

pipettes/firmware/motor_control.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+void MotorControl::setup() {
+    // GCONF 0x01
+    // IHOLD_IRUN 0x1010
+    // CHOPCONF 0x8008
+    auto txBuffer = std::array<uint8_t, BUFFER_SIZE>{};
+    constexpr uint32_t gconf_data = 0x01;
+    constexpr uint32_t ihold_irun_data = 0x1010;
+    constexpr uint32_t chopconf = 0x8008;
+    build_command(WRITE | static_cast<uint8_t>(MotorRegisters::GCONF),
```
sfoster1, line 72: one thing we could do that would make this a little easier is a free constexpr function like
```
enum class Mode = { WRITE = 0x80, READ = 0x0};
constexpr uint8_t command_byte(Mode mode, MotorRegisters register) { return static_cast<uint8_t>(mode) | static_cast<uint8_t>(register); }
```
which is kind of like a helper macro because it's constexpr, but typesafe. Could make it cleaner

pipettes/firmware/spi_comms.cpp
```diff
@@ -0,0 +1,56 @@
...
+ * send_command takes a transmit buffer (with the command and data
+ * already formatted), a uint8_t empty data which will be modified by
+ * the information received from the stepper driver and status which will also
+ * be modified by information received from the driver.
+ */
+
+/*
+ * Private Functions
+ */
+void Spi::transmit_receive(BufferType& transmit, BufferType& receive) {
```
sfoster1, line 32: `transmit` could and should be a const ref

pipettes/firmware/spi_comms.cpp
```diff
@@ -0,0 +1,56 @@
...
+    Set_CS_Pin();
+}
+
+/*
+ * Public Functions
+ */
+
+Spi::Spi() { handle = MX_SPI2_Init(); }
+
+void Spi::send_command(BufferType aTxBuffer, uint32_t& data, uint8_t& status) {
```
sfoster1, line 45: `aTxBuffer` should be a const ref

sfoster1 (approved):
Looks good in content; some form things we could fix up.

pipettes/firmware/motor_control.cpp
ahiuchingau: If we stick with TMC2130, I think they are always going to be 5 bytes.

include/pipettes/core/communication.hpp
ahiuchingau: Am I fixing this correctly? Do I keep the `default` case the same or does it need to raise an error?

sfoster1: this is fine although at some point we should probably put more thought into what happens when an unmatched message is received

include/pipettes/core/uart_message_handler.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include "pipettes/core/pipette_messages.hpp"
+
+template <typename T, motor_protocol::MotorProtocol Motor>
+requires io::WriterProtocol<T> class MessageHandler {
+  public:
+    explicit MessageHandler(T &writer, Motor &motor,
+                            communication::MessageWriter &message_writer)
+        : writer(writer), motor(motor), message_writer(message_writer) {}
+
+    void handle(const pipette_messages::ReceivedMessage &message) {
```
sfoster1, line 17: Could we make this function named something different than the ones that are called by the visitor? I understand what's going on here and the compiler handles it fine probably but to me, it's too easy to go overboard with overloads in c++ and we should really restrict ourselves to only having overloads that
- Provide the same semantic functionality
- Have essentially the same contract (constrained by their inputs)
- Basically, have any other differences between them than what is required to transform different inputs

For instance, it's usually pretty bad to have something like
```c++
class Foo {
public:
    bool bar;
    bool gotBar;
    bool getBar() { gotBar = true; return bar;}
    bool getBar() const { return bar;}
}
```
because they don't do the same thing (aside from handling different kinds of inputs). And here, the `handle` defined on line 17 does something completely different than the `handle`s defined on line 23 and down; it's in its own semantic category.

include/pipettes/core/uart_message_handler.hpp
```diff
@@ -0,0 +1,61 @@
+#pragma once
+
+#include <variant>
+
+#include "common/hal/io.hpp"
+#include "common/hal/motor.hpp"
+#include "pipettes/core/communication.hpp"
+#include "pipettes/core/pipette_messages.hpp"
+
+template <typename T, motor_protocol::MotorProtocol Motor>
```
sfoster1, line 10: Let's give the first template protocol a better name than T. I think we can also qualify it in the template as we do with motor protocol. Like maybe `template<io::WriterProtocol Writer, motor_procotol::MotorProtocol Motor>`? `T` is the common case for examples because it's so generic and thus useful as a parameter for something like a container, but here I think it's just confusing.

include/pipettes/tests/test_motor_control.hpp
```diff
@@ -0,0 +1,51 @@
+#pragma once
+
+#include <array>
+#include <cstdint>
+#include <iostream>
```
sfoster1, line 5: stray include?

pipettes/firmware/motor_control.cpp
sfoster1: Honestly if this is all going to be 5 byte commands we could just build the array in this function and return it on the stack:
```c++
auto MotorControl::build_command(uint8_t command, uint32_t command_data) -> BufferType {
   auto tx_buffer = BufferType{0};
   ...
   return tx_buffer;
}
```
It's normally a no-no in cpp to return objects by-value but especially if the object is small, whatever.

Could also get a little fancy with returning via move which would make the compiler possibly emit zero-copy code but that too is pointless. 5 bytes is very small

pipettes/firmware/motor_control.cpp
sfoster1: Really should just make this auto

include/firmware/common/spi_comms.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include "stm32g4xx_hal_conf.h"
+
+namespace spi {
+class Spi {
+  public:
+    static constexpr auto BUFFER_SIZE = 5;
+    using BufferType = std::array<uint8_t, BUFFER_SIZE>;
+    Spi();
+    void send_command(const BufferType& aTxBuffer, uint32_t& data,
```
sfoster1, line 12: I think it may be a better idea to not have this interface differentiate between `data` and `status` and just take and provide buffers of bytes. We want the things that put the bytes in buffers to be in the testable layers anyway, and regardless of where the serialization/deserialization happens these functions aren't actually parsing or building the spi messages themselves (i.e., this class doesn't know what the various registers or their content formats are) so I think we should just go all the way.

pipettes/firmware/system_stm32g4xx.c
```diff
@@ -86,7 +85,7 @@
 #endif /* HSE_VALUE */
 
 #if !defined  (HSI_VALUE)
-  #define HSI_VALUE    16000000U /*!< Value of the Internal oscillator in Hz*/
+  #define HSI_VALUE    32000000U /*!< Value of the Internal oscillator in Hz*/
```
sfoster1, line 88: let's do a big note to self to revisit the system clocking at some point

sfoster1 (changes requested):
Couple little things but the structure looks great

sfoster1 (approved):
Looks great!

PR #54, opened 2021-07-20
fix(common): bit_utils: handle size checking

sfoster1:
bit_utils didn't have a way to check that its inputs and outputs matched
in size, and had some other safety stuff around appropriate inputs that
wasn't there.

Now, we check the sizes of the available containers (in the for loops);
check the appropriateness of the contained int with an (annoying)
template requirement; and have an overload of bytes_to_int for iterator
pairs as well as a generic container that fulfills range, which should
work with any listlike container or view.


amitlissack (approved):
Awesome. Thanks!

PR #55, opened 2021-07-21
chore: Update cmake-utils

sfoster1:
Pulls recent changes from cmake-utils/main, including formatter updates, and applies the new format rules.


include/pipettes/core/communication.hpp
```diff
@@ -69,21 +69,32 @@ class MessageWriter {
     void write(Writer &communication, const pipette_messages::SentMessage &m);
 
   private:
+<<<<<<< HEAD
```
amitlissack, line 72: Unmerged stuff here.

amitlissack (changes requested):

include/pipettes/core/communication.hpp
sfoster1: Crap

amitlissack (approved):

PR #58, opened 2021-08-02
feat(can): canlib message dispatching

amitlissack:
Sorry for the monstrosity of a PR.  I seemed to just keep working on it without finding a good stopping point.

I'll happily split it up if people want. We can also review it together.

I plan to add a README although I'm not sure if I'll do that before we review the code and patterns.


can/simulator/FreeRTOSConfig.h
```diff
@@ -0,0 +1,175 @@
+/*
+ * FreeRTOS Kernel V10.0.1
+ * Copyright (C) 2017 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
```
amitlissack, line 4: This file is just here to use the posix FreeRTOS port.

can/simulator/freertos_idle_timer_task.cpp
```diff
@@ -0,0 +1,27 @@
+/*
+ * Configuration for the FreeRTOS idle task, which is necessary when we told it
+ * we're using static allocation. Provides the same configuration as the other
+ * stacks, but in callback form (vApplicationGetIdleTaskMemory is called by the
+ * RTOS internals)
+ */
+
+#include <array>
+
```
amitlissack, line 9: Also just here for free rtos port

can/simulator/rtos_main.cpp
```diff
@@ -0,0 +1,94 @@
+#include <iostream>
+#include <variant>
```
amitlissack, line 2: This "simulator" is here to illustrate interacting with the CAN message buffer/dispatching.

can/simulator/rtos_main.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    void visit(StopRequest &m) { std::cout << "HandlerB stop" << std::endl; }
+};
+
+// Handlers for the parsed messages
+static auto a_handler = HandlerA{};
+static auto b_handler = HandlerB{};
+
+// The dispatch targets.
+static auto a = freertos_can_dispatch::FreeRTOSCanDispatcherTarget<
```
amitlissack, line 51: Build a message buffer that parses GetSpeedRequest messages and sends them to `a_handler`

can/simulator/rtos_main.cpp
```diff
@@ -0,0 +1,94 @@
...
+    256, HandlerA, GetSpeedRequest>{a_handler};
+static auto b =
+    freertos_can_dispatch::FreeRTOSCanDispatcherTarget<256, HandlerB,
```
amitlissack, line 54: Build a message buffer that parses StopRequest messages and sends them to `b_handler`

can/simulator/rtos_main.cpp
```diff
@@ -0,0 +1,94 @@
...
+                                                       StopRequest>{b_handler};
+
+// Buffer poller tasks
+static auto dispatcher_task_a =
```
amitlissack, line 58: Create tasks to poll the message buffers we just created.

can/simulator/rtos_main.cpp
```diff
@@ -0,0 +1,94 @@
...
+    FreeRTOSTask<256, 5, decltype(a.poller)>("dispatcher_a", a.poller);
+static auto dispatcher_task_b =
+    FreeRTOSTask<256, 5, decltype(b.poller)>("dispatcher_b", b.poller);
+
+static auto dispatcher = Dispatcher(&a.buffer_target, &b.buffer_target);
```
amitlissack, line 63: Create a dispatcher to the two message buffers we just created

can/simulator/rtos_main.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+static auto buffer = FreeRTOMessageBuffer<1024>{};
+static auto poller =
+    freertos_can_dispatch::FreeRTOSCanBufferPoller(buffer, dispatcher);
```
amitlissack, line 67: Now build the message buffer that simulates the message coming from the bus and connect it to the dispatcher.

can/simulator/rtos_main.cpp
```diff
@@ -0,0 +1,94 @@
...
+static auto dispatcher_task =
+    FreeRTOSTask<256, 5, decltype(poller)>("dispatcher", poller);
```
amitlissack, line 69: Create a task to poll the CAN bus message buffer.

can/simulator/rtos_main.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+void vATaskFunction(void *pvParameters) {
+    auto ids = std::array<MessageId, 3>{StopRequest::id, GetSpeedRequest::id,
```
amitlissack, line 72: A task that puts messages into the CAN BUS message buffer. 

It cycles through three message types. Two have interested handlers and one does not.

pipettes/firmware/can_task.cpp
```diff
@@ -1,82 +1,74 @@
 #include "common/firmware/can_task.hpp"
```
amitlissack, line 1: This file sets up a phony Pipette can message routing path

pipettes/firmware/can_task.cpp
```diff
@@ -1,82 +1,74 @@
...
-static FDCAN_HandleTypeDef can;
-static constexpr auto buffsize = 64;
-// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
-static auto buff = std::array<uint8_t, buffsize>{};
+static auto can_bus_1 = HalCanBus(&fdcan1);
+static auto message_writer_1 = MessageWriter(can_bus_1);
 
-static void run(void *parameter) {
-    parameter = nullptr;
+struct MotorHandler {
```
amitlissack, line 22: A handler for motor related messages

pipettes/firmware/can_task.cpp
```diff
@@ -1,82 +1,74 @@
...
-            TxHeader.DataLength = RxHeader.DataLength;
-            TxHeader.FDFormat = RxHeader.FDFormat;
-
-            if (HAL_FDCAN_AddMessageToTxFifoQ(&can, &TxHeader, buff.data()) !=
-                HAL_OK) {
-                // Transmission request Error
-                Error_Handler();
-            }
+/** Dispatcher to the various handlers */
+static auto dispatcher = Dispatcher(&motor_dispatch_target.parse_target);
```
amitlissack, line 54: Notice that the Dispatcher will write directly to the parse target and bypass the message buffer target. 

That means that the Task that polls the ROOT CAN message buffer will handle the motor messages.

The idea is that `DispatchParseTarget` and `DispatchMessageBufferTarget` have the same interface. We can either chain a `DispatchMessageBufferTarget` into a `DispatchParseTarget` which requires a task to poll the MessageBuffer or go directly to a `DispatchParseTarget`

I'll make a picture.

pipettes/firmware/CMakeLists.txt
```diff
@@ -29,7 +34,9 @@ add_executable(pipettes
 
 target_link_libraries(pipettes
         PUBLIC STM32G491RETx
-        STM32G4xx_Drivers_Pipettes STM32G4xx_FreeRTOS_Pipettes)
+        STM32G4xx_Drivers_Pipettes STM32G4xx_FreeRTOS_Pipettes
```
Laura-Danielle, line 37: I was kind of under the impression that the firmware part of CAN would also be a static library (like can-core). Is the reason we can't do this because of FreeRTOS and/or the HAL?

amitlissack: I hoped to do that and had some success, but in the end decided to follow your pattern with the firmware files in `common`.

include/can/core/dispatch.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * A CanMessageBufferListener that will dispatch messages to other
+ * CanMessageBufferListeners
+ * @tparam Listener CanMessageBufferListener objects
+ */
+template <CanMessageBufferListener... Listener>
+class Dispatcher {
+  public:
+    explicit Dispatcher(Listener*... listener) : registered{listener...} {}
```
sfoster1, line 115: what's the expected lifetime of these listener pointers? should they be references instead?

sfoster1 (commented):
Looks awesome

can/firmware/hal_can_bus.cpp
```diff
@@ -0,0 +1,45 @@
...
+    HAL_FDCAN_ConfigFilter(handle, &filter_def);
+}
+
+void HalCanBus::send(uint32_t arbitration_id, uint8_t* buffer,
+                     CanFDMessageLength buffer_length) {
+    tx_header.Identifier = arbitration_id;
+    tx_header.DataLength = hal_can_utils::length_to_hal(buffer_length);
+    tx_header.MessageMarker = 0;
+
+    HAL_FDCAN_AddMessageToTxFifoQ(handle, &tx_header, buffer);
```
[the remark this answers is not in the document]
amitlissack: Excellent point. I have not fully fleshed out whether transfers will happen in this method or a separate task. Either way, I'll add a mutex here.

include/common/core/freertos_task.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include <array>
+
+#include "FreeRTOS.h"
+#include "task.h"
+
+namespace freertos_task {
+
+template <uint32_t StackDepth, UBaseType_t Priority, typename Entry>
+class FreeRTOSTask {
```
sfoster1, line 11: let's make this explicitly not copyable and not movable by doing
```
FreeRTOSTask& operator=(FreeRTOSTask&) = delete;
FreeRTOSTask&& operator=(FreeRTOSTask&&) = delete;
FreeRTOSTask(FreeRTOSTask&) = delete;
FreeRTOSTask(FreeRTOSTask&&) = delete;
```
This is because this class at initialization time holds data structures by-value stored in the class and then passes pointers to them to a third-party system; those pointers will become invalid if the class instance changes its location in memory. So we should prevent that from happening.

include/can/core/freertos_can_dispatch.hpp
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * @tparam BufferSize The size of the FreeRTOMessageBuffer in bytes
+ * @tparam HandlerType The type of the parsed message callback. Must conform to
+ * HandlesMessages concept.
+ * @tparam MessageTypes The CanMessage types this MessageBuffer is interested
+ * in.
+ */
+template <std::size_t BufferSize, typename HandlerType,
+          CanMessage... MessageTypes>
+requires HandlesMessages<HandlerType, MessageTypes...>
```
sfoster1, line 61: i think we might also want to require `!std::is_move_constructible<HandlerType> && !std::is_move_assignable<HandlerType>` to prevent the handler, which will be captured by a long-lived reference, from getting changed out from under us. We could also require `!std::is_copy_{assignable,constructible}` to prevent people from accidentally making a copy that isn't actually associated with this anymore but that wouldn't be invalid behavior just probably wrong behavior.

include/can/firmware/hal_can_bus.hpp
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * HAL FD CAN wrapper. Matches the CanBus concept.
+ */
+class HalCanBus {
+  public:
+    /**
+     * Construct
+     * @param handle A pointer to an initialized FDCAN_HandleTypeDef
+     */
+    explicit HalCanBus(FDCAN_HandleTypeDef* handle);
```
sfoster1, line 22: by forward declaring `FDCAN_HandleTypeDef` you could avoid having to include the hal_fdcan here. You'd lose the ability to handle HAL_StatusTypeDef but whatever

include/can/firmware/hal_can_message_buffer.hpp
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "can/firmware/hal_can_bus.hpp"
+#include "common/core/freertos_message_buffer.hpp"
+
+namespace hal_can_message_buffer {
+
+constexpr auto read_message_buffer_size = 1024;
```
sfoster1, line 8: feels like this should be a defaulted template argument

include/common/core/freertos_message_buffer.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+template <std::size_t BufferSize>
+FreeRTOMessageBuffer<BufferSize>::FreeRTOMessageBuffer() {
+    handle = xMessageBufferCreateStatic(backing.size(), backing.data(),
+                                        &message_buffer);
+}
+
+template <std::size_t BufferSize>
+FreeRTOMessageBuffer<BufferSize>::~FreeRTOMessageBuffer() {
+    vMessageBufferDelete(handle);
```
sfoster1, line 69: i think it probably might not come up but are you actually allowed to call `vMessageBufferDelete` on something created with `CreateStatic`?

include/common/firmware/can.h
```diff
@@ -7,6 +7,12 @@
 extern "C" {
 #endif  // __cplusplus
 
+
+/**
+ * The global can 1 handle.
+ */
+extern FDCAN_HandleTypeDef fdcan1;
```
sfoster1, line 14: this doesn't feel like a great idea. i would hope we can really strictly limit where we use globals like this and i think adding some friction to using it by requiring it to be declared extern everywhere it's used will help.

include/can/core/freertos_can_dispatch.hpp
```diff
@@ -0,0 +1,91 @@
...
+     * @param listener The listener to be called back when a new message is in
+     * the buffer.
+     */
+    FreeRTOSCanBufferPoller(BufferType& buffer, Listener& listener)
+        : buffer{buffer}, listener{listener}, reader{buffer, listener} {}
+
+    /**
+     * Task entry point.
+     */
+    void operator()() {
```
sfoster1, line 37: is this stoppable by calling some method on `reader`? if not we should probably annotate it [`[[noreturn]]`](https://en.cppreference.com/w/cpp/language/attributes/noreturn)

include/can/core/can_message_buffer.hpp
```diff
@@ -0,0 +1,134 @@
...
+                backing.begin(), backing.end(), arbitration_id);
+            listener.handle(arbitration_id, start_of_message, end);
+        }
+    }
+
+  private:
+    static auto constexpr buffer_size =
+        message_core::MaxMessageSize + sizeof(uint32_t);
+    std::array<uint8_t, buffer_size> backing{};
+    Buffer& message_buffer;
```
sfoster1, line 130: I think long term captures-by-reference like this probably want to have their types required not-movable so they can't accidentally be invalidated.

include/can/core/can_message_buffer.hpp
```diff
@@ -0,0 +1,134 @@
...
+        for (; iter != backing.end() && buffer != limit; iter++, buffer++) {
+            *iter = *buffer;
+        }
+        return iter - backing.begin();
+    }
+
+    static auto constexpr buffer_size =
+        message_core::MaxMessageSize + sizeof(uint32_t);
+    std::array<uint8_t, buffer_size> backing{};
+    Buffer& message_buffer;
```
sfoster1, line 78: long term reference, make it non-movable

can/simulator/FreeRTOSConfig.h
sfoster1: wait this stuff all works on the posix port? a lot of it looks extremely cortex-m specific

sfoster1 (changes requested):
Couple #justcppthings

include/can/firmware/hal_can_message_buffer.hpp
amitlissack: I don't think so. This is a specific message buffer. The one populated by ISR. I will add comments.

include/can/firmware/hal_can_bus.hpp
amitlissack: But I'm also using `FDCAN_TxHeaderTypeDef` in this class. I can't forward declare that.

sfoster1 (approved):
Nice! looks great!

PR #59, opened 2021-08-03
refactor(motor-control): move motor-control out of pipettes

ahiuchingau:
Move motor control class out of the pipettes subsystem.


include/motor-control/core/spi.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+#include <array>
+#include <concepts>
+#include <cstdint>
+
+namespace spi {
+
+template <class Spi>
+concept SpiProtocol = requires(Spi spi_comms,
+                               const std::array<uint8_t, 5>& transmit,
```
sfoster1, line 11: Depending on how general we want to be, it may be a good idea to make the `5` actually a template argument; even if we don't though we should probably make it depend on the `BUFFER_SIZE` value in the motor_control.hpp (or the other way around) so it's not possible to change them independently.

I think it's probably fine to keep it specific to a number; but if we do I think we should name this concept something like `TMC2130Spi` to make it clear it's not just a generic SPI interface, it's for this specific driver with this specific packet size.

include/motor-control/core/motor_control.hpp
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    [[nodiscard]] auto get_current_status() const -> uint8_t { return status; }
+
+    [[nodiscard]] auto get_current_data() const -> uint32_t { return data; }
+
+  private:
+    uint8_t status = 0x0;
+    uint32_t data = 0x0;
+    uint32_t speed{0};
```
sfoster1, line 91: [nit] let's initialize this the same way

include/motor-control/core/motor_control.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+namespace motor_control {
+
+const uint8_t WRITE = 0x80;
+/**
+ * Implementation of the Motor concept.
+ *
+ */
+
+enum class MotorRegisters : uint8_t {
```
sfoster1, line 15: to me the definitions belong in some other Controller type of class; it makes sense to keep the two concepts
- Motion controller, which takes values like velocity, distance, etc and turns them into pulse trains
- Motor driver controller, which sets values like chopper style, etc
separate, since they actually do pretty separate things.

sfoster1 (approved):
Couple small things but they don't necessarily have to be in this pr

include/motor-control/core/motor_control.hpp
Laura-Danielle: I think that change should be made in a follow up PR ^

Laura-Danielle (approved):
lgtm, aside from seth's comments. You'll probably need to remove the `/` from the title so that the semantic title passes.

PR #61, opened 2021-08-09
refactor(motor-control): create motion controller and motor driver components

ahiuchingau:
This PR separates the current motor control library into two components: (1) motion controller, and (2) motor driver.


sfoster1 (approved):
I really like this! I think in the future we might add a higher-level interface rather than having callers poke into the struct, once we start adding higher-level behaviors; but this I think is a really good framework for how the system should work and has a good presence for adding more, like a limit switch endstop or something.

Laura-Danielle (approved):
nice, I like it

PR #62, opened 2021-08-10
refactor(common): move comm protocols and motor control to common

Laura-Danielle:
## Overview
Rather than copying over firmware files into each subproject, we should put common firmware files in the common subproject so that other subprojects can simply add those files to their individual executables.

## Next Steps
- We should add message types for the gantry subsystem

## Review Requests
Does the structure seem OK? Any change requests or files to be moved around?


DerekMaggio:
A few questions from the noob.
1. How do we decide if something needs to be in the `common` project? Just if more than one other subproject uses it?
2. It looks like `can_task.cpp` is a straight copy and paste of `can_task.cpp` in `/gantry/firmware`. Can we pull this file out into common to be used by multiple pieces of firmware?

Laura-Danielle:
> A few questions from the noob.
> 
> 1. How do we decide if something needs to be in the `common` project? Just if more than one other subproject uses it?
> 2. It looks like `can_task.cpp` is a straight copy and paste of `can_task.cpp` in `/gantry/firmware`. Can we pull this file out into common to be used by multiple pieces of firmware?

1. Yep, if more than one subproject uses a file _and_ it can be separated out then it should go in common. Things that cannot be separated out typically rely heavily on parts of the BSP for either FreeRTOS or an ST board.
2. `can_task.cpp` falls into the category that it cannot be separated out due to it relying directly on FreeRTOS.

sfoster1 (approved):
Looks good!

amitlissack:
Is the idea that the `common/firmware` files are the default configurations for the various peripherals? These can be overriden by not compiling them into the sub-project?

Laura-Danielle:
> Is the idea that the `common/firmware` files are the default configurations for the various peripherals? These can be overriden by not compiling them into the sub-project?

Yep that's correct

PR #63, opened 2021-08-12
refactor(motor-control): add GPIO config file in each individual subsystem

ahiuchingau:
We can now configure each subsystem with a unique set of GPIO pins for motor control.


sfoster1 (changes requested):
Let's not introduce more header-definition based configuration. We already have some that we have to use because of other people's code, but let's not add more.

Instead, we should define a struct that holds the configuration so that it can be passed at runtime. As long as we declare the port as a pointer, we can forward-declare it to C and be just fine without having to include the device header; and then the pin is just a number.

There's two or three basic ways we could handle this struct:
- We could have an API like this, with set-direction/set-step/etc, which takes as its arguments the port and pin. This would be the quickest method but it might provide for some programmer error since you could accidentally pass the step pin into set_direction.
- We could have an API that looks more like just set and reset, which takes as its arguments the port and pin; this would be more general. It would look basically exactly like that HAL class, and would probably get eliminated by the compiler, but would continue to serve the purpose of avoiding including the hal headers in the c++
- We could make this C interface into an object. What we'd do here is we'd take that struct and make these set_direction/set_step/etc functions take a pointer to that struct as their first argument; the firmware layer for the end executable would create it, pass it into the c++ motor controller, and that would hold it by-value and pass it down to the c calls. We could even take this a step further and store pointers to these functions in the struct, but that's kind of nasty.

Either way, let's not do this configuration via #define.

common/firmware/motor_driver/motor.c
```diff
@@ -4,19 +4,26 @@
  */
 
 #include "FreeRTOS.h"
+#include "motor_gpio_conf.h"
 #include "stm32g4xx_hal_conf.h"
 #include "task.h"
 
-void Set_Direction() { HAL_GPIO_WritePin(GPIOB, GPIO_PIN_1, GPIO_PIN_SET); }
+void Set_Step() { HAL_GPIO_WritePin(STEP_GPIO_PORT, STEP_PIN, GPIO_PIN_SET); }
```
amitlissack, line 11: Minor point, but we use snake case for function names. Lower case.

pipettes/firmware/uart_task.cpp
```diff
@@ -8,6 +8,7 @@
 #include "common/firmware/uart_comms.hpp"
 #include "motor-control/core/motor.hpp"
 #include "pipettes/core/communication.hpp"
+//#include "pipettes/firmware/hardware_config.hpp"
```
sfoster1, line 11: remove

sfoster1 (approved):
Nice! this is pretty much what we were thinking about. One commented include to remove though.

pipettes/firmware/uart_task.cpp
ahiuchingau: Oh man!

PR #64, opened 2021-08-13
fix(motor): clocking part2

Laura-Danielle:
## Overview
We thought sending pulses to the motor required a higher clock speed, but I think that was a red herring? Also adding back in a delay so the motor can actually step properly.

In other news, anyone have resources for figuring out the correct clock speed for different applications? 😢


ahiuchingau (approved):
🕐

PR #65, opened 2021-08-16
refactor(common): pull RCC clock select into the common directory

Laura-Danielle:
## Overview

We shouldn't have the RCC clock select function be dependent on UART. This PR separates that function, and also error handler functions into their own C files that can be used as needed.

# Review Requests
Are we OK with the file locations/naming/functionality as is? Anything you'd like changed?

### DIFF
--- a/common/firmware/errors/errors.c
+++ b/common/firmware/errors/errors.c
@@ -1,3 +1,7 @@
+#pragma once
+
+#include "stm32g4xx_hal_conf.h"
+
 #include "common/firmware/errors.h"
 
 void Error_Handler(void) {
--- a/common/firmware/errors/errors.c
+++ b/common/firmware/errors/errors.c
@@ -1,0 +1,11 @@
+#include "common/firmware/errors.h"
+
+void Error_Handler(void) {
+    /* USER CODE BEGIN Error_Handler_Debug */
+    /* User can add his own implementation to report the HAL error return state
+     */
+    __disable_irq();
+    while (1) {
+    }
+    /* USER CODE END Error_Handler_Debug */
+}
--- a/common/firmware/uart/uart.c
+++ b/common/firmware/uart/uart.c
@@ -1,6 +1,7 @@
 #include "common/firmware/uart.h"
+#include "common/firmware/errors.h"
 
-static void Error_Handler();
+
 /**
  * Initializes the Global MSP.
  */
@@ -137,23 +138,4 @@ UART_HandleTypeDef MX_LPUART1_UART_Init() {
 
     /* USER CODE END USART1_Init 2 */
     return huart1;
-}
-
-void RCC_Peripheral_Clock_Select() {
-    RCC_PeriphCLKInitTypeDef PeriphClkInit = {0};
-
-    PeriphClkInit.PeriphClockSelection = RCC_PERIPHCLK_USART1 | RCC_PERIPHCLK_FDCAN;
-    PeriphClkInit.Usart1ClockSelection = RCC_USART1CLKSOURCE_PCLK2;
-    PeriphClkInit.FdcanClockSelection = RCC_FDCANCLKSOURCE_PCLK1;
-    if (HAL_RCCEx_PeriphCLKConfig(&PeriphClkInit) != HAL_OK) {
-        Error_Handler();
-    }
-
-}
-
-static void Error_Handler() {
-    for(;;) {
-        HAL_GPIO_TogglePin(GPIOA, GPIO_PIN_5);
-        HAL_Delay(100);
-    }
 }
--- a/include/common/firmware/clocking.h
+++ b/include/common/firmware/clocking.h
@@ -1,11 +1,14 @@
-#include "stm32g4xx_hal_conf.h"
-
+#ifndef __CLOCKING_H__
+#define __CLOCKING_H__
 
 #ifdef __cplusplus
 extern "C" {
 #endif  // __cplusplus
+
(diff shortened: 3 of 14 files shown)


common/firmware/errors/errors.c
```diff
@@ -0,0 +1,11 @@
+#include "common/firmware/errors.h"
+
+void Error_Handler(void) {
```
amitlissack, line 3: Nice

include/common/firmware/errors.h
```diff
@@ -0,0 +1,9 @@
+#include "stm32g4xx_hal_conf.h"
```
amitlissack, line 1: Is this needed?

include/common/firmware/clocking.h
```diff
@@ -0,0 +1,11 @@
+#include "stm32g4xx_hal_conf.h"
```
amitlissack, line 1: Is this include needed?

Laura-Danielle: I can either include it in this header file or the `.c` file. We need the hal_conf because of the HAL functions we are using in both `errors.c` and `clocking.c`.

include/common/firmware/errors.h
Laura-Danielle: See above comment.

include/common/firmware/clocking.h
amitlissack: I understand. 

I always try to limit includes to `c` files whenever possible. Purely to reduce compilation times. If a header file changes then all files that include that file will need to be rebuilt.  That can have a cascading effect.

common/firmware/errors/errors.c
```diff
@@ -0,0 +1,11 @@
+#include "common/firmware/errors.h"
```
amitlissack, line 1: Need an include guard (#pragma once)

common/firmware/uart/uart.c
```diff
@@ -1,6 +1,7 @@
 #include "common/firmware/uart.h"
```
amitlissack, line 1: Need an include guard (#pragma once)

include/common/firmware/errors.h
amitlissack: Need an include guard (#pragma once)

include/common/firmware/clocking.h
amitlissack: Need an include guard (#pragma once)

sfoster1: include guard comment in the wrong place?

I agree with amit, for a couple other reasons. In general, having include files that include others that they do not _directly_ require to function bloats compilation time but also hides bugs where some .c(pp) file doesn't include something but does include this, and it gets papered over.

Having specifically the hal files included here also makes this basically not usable from c++ because the hal does stuff that isn't really good to do in c++ (volatile increment-assignment). We should avoid it if possible.

sfoster1 (changes requested):
Is this something that can actually be shared?

As implemented right now this is controlling peripheral clock gating for every peripheral on the system. For it to be shared means either
- every project gates on the clocks of the exact same peripherals, no matter whether they need them or not
- this only handles some core set of peripherals, or just creates the periphclkinit struct with that data and returns it for a later call or something, or takes a struct with uncommon stuff and fills it in.

Some of this is definitely common, like the can; some of it may not be, like that specific i2c bus driver. 

Maybe it's something that we share for now and take back later I guess but this is going to become really painful as soon as these places diverge unless it becomes something that either takes or provides the clock init struct.

include/common/firmware/clocking.h
Laura-Danielle: I see. That makes sense, will definitely change it.

Laura-Danielle:
> Is this something that can actually be shared?
> 
> As implemented right now this is controlling peripheral clock gating for every peripheral on the system. For it to be shared means either
> 
> * every project gates on the clocks of the exact same peripherals, no matter whether they need them or not
> * this only handles some core set of peripherals, or just creates the periphclkinit struct with that data and returns it for a later call or something, or takes a struct with uncommon stuff and fills it in.
> 
> Some of this is definitely common, like the can; some of it may not be, like that specific i2c bus driver.
> 
> Maybe it's something that we share for now and take back later I guess but this is going to become really painful as soon as these places diverge unless it becomes something that either takes or provides the clock init struct.

What about if we make this function a little more versatile and allow each project to pass in the type of clocks it needs by using a struct?

sfoster1:
> > Is this something that can actually be shared?
> > As implemented right now this is controlling peripheral clock gating for every peripheral on the system. For it to be shared means either
> > 
> > * every project gates on the clocks of the exact same peripherals, no matter whether they need them or not
> > * this only handles some core set of peripherals, or just creates the periphclkinit struct with that data and returns it for a later call or something, or takes a struct with uncommon stuff and fills it in.
> > 
> > Some of this is definitely common, like the can; some of it may not be, like that specific i2c bus driver.
> > Maybe it's something that we share for now and take back later I guess but this is going to become really painful as soon as these places diverge unless it becomes something that either takes or provides the clock init struct.
> 
> What about if we make this function a little more versatile and allow each project to pass in the type of clocks it needs by using a struct?

Sure we definitely could but also... what does this function now do? It's a function that takes a struct and then just immediately passes that struct to another function. What does it gain us?

sfoster1 (approved):
Nice!

ahiuchingau (approved):
I like it!

PR #66, opened 2021-08-18
Add i2c support

Laura-Danielle:
## Overview

These are the first two steps in pipette recognition. We should next add in messaging support to
read and write from the eeprom using CANBus. The hope, with this PR, is to use the structure laid out by the EEPROM concept for other things. This means that when we need to support things such as the pressure sensor etc, we can simply write a pressure sensor policy concept with all the functionality required and pass in a i2c writer to it (along with implementing CAN bus messages).

closes #RET-89 and closes #RET-90.

## Review Requests
The tests kind of suck, but I couldn't think of other things to test. Do you like the locations of the files? Do you like the structure?

## Notes

The EEPROM can also be used to store CAN IDs, do we want to support this?


include/common/firmware/i2c_comms.hpp
```diff
@@ -0,0 +1,20 @@
...
+#include <span>
+
+#include "stm32g4xx_hal_conf.h"
+
+namespace i2c {
+class I2C {
+  public:
+    uint8_t DEVICE_ADDRESS = 0x1;
+    static constexpr auto BUFFER_SIZE = 1;
+    using BufferType = std::array<uint8_t, BUFFER_SIZE>;
```
Laura-Danielle, line 11: We might want to make a BufferType concept in a follow-up PR? I decided not to do that here.

include/motor-control/core/motor_driver.hpp
```diff
@@ -5,7 +5,6 @@
 
 namespace motor_driver {
 
-const uint8_t WRITE = 0x80;
```
Laura-Danielle, line 8: Random leftover WRITE command not used anywhere so I removed it here.

include/pipettes/core/eeprom.hpp
```diff
@@ -0,0 +1,39 @@
...
+/*
+ * A template and helper functions describing
+ */
+
+namespace i2c {
+template <class I2C>
+concept EEPromPolicy =
+    requires(I2C i2c_comms, const uint8_t transmit,
+             std::array<uint8_t, I2C::BUFFER_SIZE>& receive) {
+    {i2c_comms.BUFFER_SIZE};
```
Laura-Danielle, line 18: Rather than having to dig further for the buffer size parameter, I am making BUFFER_SIZE a member of the EEPromPolicy concept. This allows me to have the line on 30.

include/pipettes/tests/test_eeprom.hpp
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <array>
+#include <cstdint>
+
+namespace test_eeprom {
+
+class TestI2C {
```
Laura-Danielle, line 8: Test I2C Comms class used in `test_eeprom.cpp`

include/pipettes/core/eeprom.hpp
```diff
@@ -0,0 +1,39 @@
...
+    {i2c_comms.transmit(transmit)};
```
Laura-Danielle, line 19: On lines 19-20, I am specifying that my concept requires a transmit/receive function. This allows me do `writer.transmit` and `writer.receive` where writer is my EEPromPolicy concept. @sotaoverride

common/firmware/i2c/i2c.c
```diff
@@ -0,0 +1,67 @@
...
+        GPIO_InitStruct.Alternate = GPIO_AF4_I2C1;
+        HAL_GPIO_Init(
+            GPIOB,  // NOLINT(cppcoreguidelines-pro-type-cstyle-cast)
+            &GPIO_InitStruct);  // NOLINT(cppcoreguidelines-pro-type-cstyle-cast)
+    }
+}
+
+I2C_HandleTypeDef MX_I2C1_Init()
+{
+    I2C_HandleTypeDef hi2c1;
```
sfoster1, line 26: Just as we discussed in some other PRs, if this is a common library it really can't do things like directly configure a specific I2C peripheral instance. Using this code would require every program to use one specific I2C peripheralk, and to map its outputs to the same specific GPIO pins. It should either be part of `pipette`, because I think that's where it's used, or have a lot of the extreme low level configuration pulled out - at least the I2C unit assignment and the specific GPIO pins.

include/common/firmware/i2c_comms.hpp
sfoster1: or make it a span or something

sfoster1 (changes requested):
We definitely can't have specific-peripheral-instance configuration in `common` but the rest looks pretty good

common/firmware/i2c/i2c.c
```diff
@@ -0,0 +1,66 @@
+#include "common/firmware/i2c.h"
```
sfoster1, line 1: I think this should probably be left up to the individual executables, or at least allow them to customize things like pins, etc.

sfoster1: The same reasons that cause the HAL to have this kind of apply here - this code should be general and do only a minimal amount of autoconfiguration.

sfoster1 (changes requested):

include/pipettes/core/eeprom.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include "common/core/bit_utils.hpp"
+
+/*
+ * A template and helper functions describing
+ */
+
+namespace i2c {
+template <class I2C>
+concept EEPromPolicy =
+    requires(I2C i2c_comms, const uint8_t transmit,
```
sfoster1, line 16: shouldn't this allow a larger buffer for the serial to write? i guess it doesn't matter much right now but It might save us some pain down the road

include/pipettes/core/eeprom.hpp
```diff
@@ -0,0 +1,39 @@
+#pragma once
```
sfoster1, line 1: really like that this file exists and is in pipettes, i think that's a great separation of concerns

include/common/firmware/i2c_comms.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
+#include <span>
```
sfoster1, line 2: don't need this I think

include/common/firmware/i2c_comms.hpp
```diff
@@ -0,0 +1,20 @@
...
+#include "stm32g4xx_hal_conf.h"
+
+namespace i2c {
+class I2C {
+  public:
+    uint8_t DEVICE_ADDRESS = 0x1;
+    static constexpr auto BUFFER_SIZE = 1;
+    using BufferType = std::array<uint8_t, BUFFER_SIZE>;
+    I2C();
+    void transmit(uint8_t value);
```
sfoster1, line 13: I feel like this is where the span or something should be, since this is common code might as well make it a buffer

include/common/firmware/i2c.h
```diff
@@ -0,0 +1,29 @@
...
+extern "C" {
+#endif  // __cplusplus
+struct I2CConfig {
+    I2C_TypeDef* instance;
+    uint8_t address;
+    uint16_t address_mode;
+};
+I2C_HandleTypeDef MX_I2C_Init(struct I2CConfig* conf);
+
+#define REGISTER1 I2C1
```
sfoster1, line 19: This, I don't think is a good idea. using #define to point to other defines tends to muddy things up a lot, and is really only worth it to kind of move the same number between entirely separate semantic concepts. For instance, if this was `eeprom.h` and this was `#define EEPROM_I2C I2C1`, that would be one thing. But here, 
- `REGISTER1` is actually _less_ specific a name than `I2C1`
- You have to include this file to get `REGISTER1`, which means you're transitively including the hal, so you're not really gaining anything out of it

include/common/firmware/i2c.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
```
sfoster1, line 1: The more I read this the more I think the common code should just be hands off on specific peripheral initialization. This uses the same weird static-callback style that compilers are really bad at emitting diagnostics for as the hal, it makes you include the hal anyway if you include this file... I honestly think that if you had the c++ stuff exactly the same but just removed i2c.h and made the i2c_comms firmware object (this doesn't have to go in the concept or anything) need
- a fully constructed handle, by-reference (since i'm pretty sure it has internal pointers)
- and an associated address and address mode, which rather than being `#define`s were enum or enum class elements
then I think things would be better.

The other benefit to that is that it brings more of the static free-store initialization closer to itself. I think that sort of thing really should be kept in the end executables, those are the only places that should ship around references to static free store objects.

sfoster1: In fact the other benefit of this is that you could even like forward declare your config element struct and then i2c_comms.hpp wouldn't have to include the hal at all, that'd be cool

sfoster1 (changes requested):
Definitely looking better, love the eeprom vs i2c distinction, but I think we should take the initialization stuff one step farther and just make it live in the end executable.

include/pipettes/core/eeprom.hpp
Laura-Danielle: The HAL Transmit function only allows 8 bits of data to be sent at a time. If we want to write more, we would probably need to implement something in the individual write function that takes in a buffer and parses that buffer appropriately. I think we should keep it as uint8_t for now until we have a better idea of the serial number size we'll be sending.

PR #67, opened 2021-08-19
feat(can): socket can integration

amitlissack:
## Background

This PR adds socket-can integration into the can-simulator project.

In the future, `can-simulator` will be a shared-lib. For now it is an application that runs, accepts two different kinds of messages (SetSpeedRequest and GetSpeedRequest). It simply loops back what it receives.

No filtering is implemented.

This is primarily to give @DerekMaggio something to play with while i'm on vacation.

Important to note. `can-simulator` is linux only.

## Testing

This screen grab shows the `can-simulator` in action. On the left is a python script using running. All it does is send a `SetSpeedRequest` message , read the response, and repeat. On the right is the output of the can-simulator. 

Screen Shot 2021-08-18 at 8 10 31 PM

This is the python script used for testing.

```
import can
import time


bus = "socketcan"
channel = "vcan0"


b = can.interface.Bus(channel=channel, bustype=bus)

for i in range(15):
	msg = can.Message(arbitration_id=3 << 15, data=[1, 2, 3, 4])
	b.send(msg)
	print(b.recv(1))
	time.sleep(1)

```


can/simulator/socket_can.cpp
```diff
@@ -0,0 +1,81 @@
...
+    taskENTER_CRITICAL();
+    auto read_len = ::read(handle, &frame, sizeof(struct can_frame));
+    taskEXIT_CRITICAL();
+
+    if (read_len > 0) {
+        arb_id = frame.can_id;
+        buff_len = frame.can_dlc;
+        ::memcpy(buff, frame.data, buff_len);
+
+        std::cout << "arb_id: " << std::hex << arb_id << " "
```
Laura-Danielle, line 69: Not a real concern since this will be running on a laptop, but cout tends to be much slower than printf. See this [stack overflow article](https://stackoverflow.com/questions/2872543/printf-vs-cout-in-c) about it.

include/can/simulator/sim_canbus.hpp
```diff
@@ -0,0 +1,93 @@
...
+#include "can/core/can_message_buffer.hpp"
+#include "can/core/message_core.hpp"
+#include "common/core/message_buffer.hpp"
+
+using namespace can_bus;
+
+namespace sim_canbus {
+
+template <class T>
+concept BusTransport = requires(T t, uint32_t arb_id, uint32_t& out_arb_id,
```
Laura-Danielle, line 15: Should these types also be unsigned?

Laura-Danielle (commented):
Just some random comments/questions. In general looks good.

can/simulator/FreeRTOSConfig.h
```diff
@@ -71,7 +71,7 @@ extern uint32_t SystemCoreClock;
 #define configCPU_CLOCK_HZ (SystemCoreClock)
 #define configTICK_RATE_HZ ((TickType_t)1000)
 #define configMINIMAL_STACK_SIZE ((uint16_t)128)
-#define configTOTAL_HEAP_SIZE ((size_t)(15 * 1024))
+#define configTOTAL_HEAP_SIZE ((size_t)(64 * 1024))
```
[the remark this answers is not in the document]
amitlissack: 64k? That's not so big really.

I really struggled to get this code to work and this change may have been some random blind experimentation.

can/simulator/socket_can.cpp
amitlissack: When we get into simulation we should create a `log` function so we can change this in one place.

include/can/simulator/sim_canbus.hpp
amitlissack: Definitely.

can/simulator/socket_can.cpp
sfoster1: https://www.boost.org/doc/libs/1_63_0/libs/log/doc/html/index.html ?

sfoster1 (approved):
Looks good to me!

Laura-Danielle (approved):

PR #75, opened 2021-09-07
feat(eeprom): add eeprom messages to CANBus

Laura-Danielle:
## Overview
Closes ticket RET-110 in jira. With this PR you should be able to read and write to the eeprom via CAN messaging. I randomly picked the arbitration IDs, if you don't like them I'm very open to changing it.

## Review requests
Is the size of messages OK, or will it be a problem in CANBus? Anything else I didn't account for?


amitlissack (approved):
Nice job!

sfoster1 (approved):

PR #76, opened 2021-09-07
feat(motor-control): use timer interrupt to move motor

ahiuchingau:
We are using the basic timer interrupt register 7 to move motor instead of using `HAL_Delay()` or `vTaskDelay()`


common/firmware/motor_driver/motor.c
```diff
@@ -18,3 +19,7 @@ void reset_pin(struct PinConfig config) {
 }
 
 void delay(const int seconds) { vTaskDelay(seconds); }
+
+void start_it() { TIM7_Start_IT(); }
```
amitlissack, line 23: Nit: Does `IT` stand for something (interrupt timer)? I read this as "start it". Can we call it by a full name?

include/motor-control/core/motor_driver.hpp
```diff
@@ -34,12 +42,16 @@ class MotorDriver {
...
 
     void setup() {
-        // GCONF 0x01
-        // IHOLD_IRUN 0x1010
-        // CHOPCONF 0x8008
         constexpr uint32_t gconf_data = 0x01;
         constexpr uint32_t ihold_irun_data = 0x1010;
         constexpr uint32_t chopconf = 0x8008;
+
+        //        constexpr uint32_t gconf_data = 0x04;
```
amitlissack, line 49: Do we need these commented out lines?

amitlissack (commented):
Some little nits.

common/firmware/motor_driver/motor.c
Laura-Danielle: It stands for interrupt timer. We can change it to be a full name for non-HAL functions.

include/motor-control/core/motor_driver.hpp
Laura-Danielle: I believe so. I will change it.

gantry/firmware/tim7.c
```diff
@@ -0,0 +1,76 @@
...
+    }
+}
+
+void HAL_TIM_Base_MspInit(TIM_HandleTypeDef *htim) {
+    if (htim == &htim7) {
+        /* Peripheral clock enable */
+        __HAL_RCC_TIM7_CLK_ENABLE();
+
+        /* TIM7 interrupt Init */
+        HAL_NVIC_SetPriority(TIM7_IRQn, 0, 0);
```
sfoster1, line 64: This priority will be a problem as soon as you interact with anything freertos based in the isr. might be fine for now but as a note for the future we'll have to drop it below the values in freertosconfig.

gantry/firmware/tim7.c
```diff
@@ -0,0 +1,76 @@
...
+    GPIO_InitStruct.Pull = GPIO_NOPULL;
+    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
+    HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
+
+    /* EXTI interrupt init*/
+    HAL_NVIC_SetPriority(EXTI15_10_IRQn, 0, 0);
+    HAL_NVIC_EnableIRQ(EXTI15_10_IRQn);
+}
+
+void MX_TIM7_Init(void) {
```
sfoster1, line 32: Might make sense to have this take the desired interrupt frequency as an argument and do whatever math is required there; failing that, somewhere the frequency should be set in a #define or something with the macro based math from the constants in the system clocking file to derive the prescaler value.

include/common/firmware/tim7.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
```
sfoster1, line 1: This lives in common, but the .c file lives in gantry. I think it's a pretty good idea - it's the static callbacks again basically - but if we're going to do it this way I think the header here should not specify tim7, just specify an api for starting and stopping a general purpose timer, like
```
void motor_hardware_init(); // implies MX_GPIO_Init() and MX_TIM7_Init() and TIM7_Init()
void motor_hardware_start(); // implies TIM7_Start_IT
void motor_hardware_stop(); // implies TIM7_Stop_IT
```

and probably should have some way of indicating what the interrupt should call.

include/motor-control/core/motor_driver.hpp
```diff
@@ -34,12 +42,12 @@ class MotorDriver {
     MotorDriver(SpiDriver& spi) : spi_comms(spi) {}
 
     void setup() {
-        // GCONF 0x01
-        // IHOLD_IRUN 0x1010
-        // CHOPCONF 0x8008
-        constexpr uint32_t gconf_data = 0x01;
-        constexpr uint32_t ihold_irun_data = 0x1010;
-        constexpr uint32_t chopconf = 0x8008;
+        constexpr uint32_t gconf_data = 0x04;
```
sfoster1, line 45: These values at some point need some documenting

pipettes/firmware/stm32g4xx_it.c
```diff
@@ -135,7 +136,22 @@ void FDCAN1_IT0_IRQHandler(void)
     /* USER CODE END FDCAN1_IT0_IRQn 1 */
 }
 
+/**
+  * @brief This function handles EXTI line[15:10] interrupts.
+  */
+void EXTI15_10_IRQHandler(void)
```
sfoster1, line 142: what's this for?

pipettes/firmware/system_stm32g4xx.c
```diff
@@ -227,11 +227,16 @@ void SystemClock_Config(void) {
                                   RCC_CLOCKTYPE_PCLK1 | RCC_CLOCKTYPE_PCLK2;
     RCC_ClkInitStruct.SYSCLKSource = RCC_SYSCLKSOURCE_PLLCLK;
     RCC_ClkInitStruct.AHBCLKDivider = RCC_SYSCLK_DIV1;
-    RCC_ClkInitStruct.APB1CLKDivider = RCC_HCLK_DIV1;
+    RCC_ClkInitStruct.APB1CLKDivider = RCC_HCLK_DIV2;
     RCC_ClkInitStruct.APB2CLKDivider = RCC_HCLK_DIV1;
 
     HAL_RCC_ClockConfig(&RCC_ClkInitStruct, FLASH_LATENCY_4);
-
+//    PeriphClkInit.PeriphClockSelection = RCC_PERIPHCLK_LPUART1;
```
sfoster1, line 234: let's get rid of the commented code here

sfoster1 (changes requested):
Some stuff we should tune up bust looking good!

include/motor-control/core/motor_driver.hpp
Laura-Danielle: They're still pretty influx, but will definitely document once we have solidified setup values.

gantry/firmware/tim7.c
Laura-Danielle: I think we'll do this in the next PR. I think we'll be fine-tuning the frequency and setting up the calculations needed to determine whether we should take a step or not.

pipettes/firmware/stm32g4xx_it.c
Laura-Danielle: We don't need it. It was something added by the ST Cube IDE.

sfoster1 (approved):
Nice!

PR #78, opened 2021-09-13
feat(motor-control): command motor to move specific steps

ahiuchingau:
We can now specify the number of steps to move the motor through UART.


sfoster1 (changes requested):
There are two problems with this: a specific technical one, and a larger conceptual one. Luckily, the fix is the same for both.

The specific technical one is that `set_steps()` and the handling of `steps` in general will cause a data race, because it is being read and written by an ISR and a freertos task, and the ISR can preempt at any time. That needs some freertos aware synchronization primitive.

The larger conceptual one, though, is that this is a level-setting state. There's always a "steps", and either it's 0 or not-0, and when it's not-0 you're moving. And there's a message that sets the current number of steps. But what we want is to handle discrete, chunked, moves, with distances and times and velocities and accelerations all associated with them.

To that end, I think we should add a small queue in between the motor control task and the timer interrupt. This queue represents moves that have been parsed and prepped and are ready for execution. The motor control task submits them. The ISR reads them (and since it's reading from a freertos queue with appropriate methods, this is now safe).

The ISR, in turn, has a local state of "current move" - steps left, velocity, acceleration (eventually). It executes this move until it's done. When it's done, it tries to pull a move off the queue, and when it pulls one, it executes it. Now we're back in the conceptual space of executing a queue of moves, which is what we want.

For now these moves can just be step counts; but we should definitely handle them with the "move" command rather than a "set steps" command which carries the concept of setting an everlasting state.

include/common/core/freertos_message_queue.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+#include "FreeRTOS.h"
+#include "queue.h"
+
+namespace freertos_message_queue {
+
+template <typename Message, size_t queue_size = 10>
+class FreeRTOSMessageQueue {
+  public:
+    explicit FreeRTOSMessageQueue(uint8_t notification_bit, const char* name)
```
amitlissack, line 13: Looks like `notification_bit` isn't really used, am i right?

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace motor_handler {
+
+using Message = pipette_messages::Move;
+
+template <template <class> class QueueImpl>
+requires MessageQueue<QueueImpl<Message>, Message>
+class MotorInterruptHandler {
+  public:
+    using GenericQueue = QueueImpl<Message>;
```
amitlissack, line 15: I would like to see some doc strings explaining what these methods are doing.

gantry/firmware/step_motor.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "motor-control/core/motor_messages.hpp"
+
+using namespace motor_messages;
+
+static auto handler_class = motor_handler::MotorInterruptHandler<
+    freertos_message_queue::FreeRTOSMessageQueue>();
+
+void step_motor() {
+    if (handler_class.can_step()) {
+        toggle_step_pin();
```
sfoster1, line 13: worth noting that this will step half as fast as you think since it can toggle once per interrupt. the two common approaches to this are to have the pin go on and off in the same interrupt, which costs you time since there's a minimum time the step pin must be high to make the driver work; or to always set the step pin low at the beginning of the interrupt, then set it high again if necessary, which might have a similar timing problem. i think this is fine for now since we're not setting velocity, but something to keep in mind.

include/common/core/freertos_message_queue.hpp
```diff
@@ -0,0 +1,74 @@
...
+        BaseType_t higher_woken = pdFALSE;
+        auto sent = xQueueSendFromISR(queue, &message, &higher_woken);
+        portYIELD_FROM_ISR(  // NOLINT(cppcoreguidelines-pro-type-cstyle-cast)
+            higher_woken);
+        return sent;
+    }
+
+    auto try_read_isr(Message* message) const -> bool {
+        BaseType_t higher_woken = pdFALSE;
+        return xQueueReceiveFromISR(queue, message, &higher_woken) == pdTRUE;
```
sfoster1, line 51: see `try_write_isr`, might need to do another `portYIELD` depending on `higher_woken`. should be in the freertos docs.

include/motor-control/core/motion_controller.hpp
```diff
@@ -34,12 +47,12 @@ class MotionController {
...
     void set_acceleration(uint32_t a) { acc = a; }
-    void set_distance();
+    void set_distance(uint32_t d) { dist = d; }
 
-    void move() {
+    void move(const Move& msg) {
         set_pin(hardware_config.enable);
         set_pin(hardware_config.direction);
-        timer_interrupt_start();
+        queue.try_write(msg);
```
sfoster1, line 55: this is configured to drop the message if it fails. it's worth considering that this is in a freertos task and it therefore may in fact be fine to use a direct `write` that blocks until the message is sent, because that way we won't lose a message (or might lose it farther back where it's easier to send the error).

include/pipettes/core/pipette_messages.hpp
```diff
@@ -32,14 +35,17 @@ struct GetStatusResult {
     uint32_t data;
 };
 
-struct Status {};
+struct SetDistance {
+    uint32_t mm;
```
sfoster1, line 39: Whats the difference between "set distance" and "move"?

pipettes/firmware/step_motor.cpp
```diff
@@ -0,0 +1,29 @@
+#include "motor-control/core/step_motor.hpp"
```
sfoster1, line 1: kinda seems like this may be able to be shared?

sfoster1 (approved):
This looks good and we can merge it but I'm still concerned about set distance and move. What's the difference between the two? Why is "set distance" necessary?

include/pipettes/core/pipette_messages.hpp
ahiuchingau: Move replaces set distance, so we can remove that

Laura-Danielle: Yeah, I was planning to refactor that when we flesh out the move command.

gantry/firmware/step_motor.cpp
Laura-Danielle: Yep, we're aware. We'll be making tweaks in following PRs. The main goal of the fixup was to mainly add in the queue.

include/common/core/freertos_message_queue.hpp
Laura-Danielle: yep it seems like it does need to be called.

include/motor-control/core/motion_controller.hpp
Laura-Danielle: I see what you're saying. I think I would rather consider this in a follow-up PR because we aren't even really taking into account any overflow in the queue at the moment. Will definitely keep in mind.

PR #79, opened 2021-09-13
feat(can): Can device id message

amitlissack:
- Fill out `DeviceInfoRequest` and `DeviceInfoResponse` messages
- Add `DeviceInfoHandler` to handle device info messages. 
- Pipette sub-project handles `DeviceInfoRequest` messages.

closes RET-164 in Jira.

### DIFF
--- a/can/tests/test_messages.cpp
+++ b/can/tests/test_messages.cpp
@@ -99,7 +99,7 @@ SCENARIO("message serializing works") {
 
     GIVEN("a device info response message") {
         auto message = DeviceInfoResponse{NodeId::pipette, 0x00220033};
-        auto arr = std::array<uint8_t, 5>{0,0, 0, 0, 0};
+        auto arr = std::array<uint8_t, 5>{0, 0, 0, 0, 0};
         auto body = std::span{arr};
         WHEN("serialized") {
             auto size = message.serialize(arr.begin(), arr.end());
--- a/can/tests/test_messages.cpp
+++ b/can/tests/test_messages.cpp
@@ -34,6 +34,17 @@ SCENARIO("message deserializing works") {
             }
         }
     }
+
+    GIVEN("a device info response body") {
+        auto arr = std::array<uint8_t, 5>{1, 2, 3, 4, 5};
+        WHEN("constructed") {
+            auto r = DeviceInfoResponse::parse(arr.begin(), arr.end());
+            THEN("it is converted to a the correct structure") {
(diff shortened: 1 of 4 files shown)


include/can/core/messages.hpp
```diff
@@ -45,6 +45,44 @@ struct HeartbeatResponse {
...
+    auto serialize(Output body, Limit limit) const -> uint8_t {
+        return 0;
+    }
+};
+
+struct DeviceInfoResponse {
+    static const auto id = MessageId::device_info_response;
+
+    NodeId node_id;
+    uint32_t version;
```
amitlissack, line 66: Preliminary values for the response. Enough to id the sub system and its version.

is uint32 good enough for version?

can/tests/test_messages.cpp
```diff
@@ -34,6 +34,17 @@ SCENARIO("message deserializing works") {
...
         }
     }
+
+    GIVEN("a device info response body") {
+        auto arr = std::array<uint8_t, 5>{1, 2, 3, 4, 5};
+        WHEN("constructed") {
+            auto r = DeviceInfoResponse::parse(arr.begin(), arr.end());
+            THEN("it is converted to a the correct structure") {
+                REQUIRE(r.node_id == NodeId::pipette);
+                REQUIRE(r.version == 0x02030405);
```
DerekMaggio, line 44: Can these hex values be pulled out into constants to give them some context?

include/can/core/messages.hpp
sfoster1: IMO we should set this up for a couple more things than just version, and then can care about version in general a little less. I don't think it's necessarily critical to get right the first time, but we could do something like
- two bits of build type (dev, testing, release)
- a byte or two of like message schema version
- a four byte incrementing version seems fine i suppose
- a byte or two for well-known-id system identification through some centrally defined enum
- at some point we'll want a serial number probably
- a hardware revision unless we want to fold that into the well-known-id

sfoster1 (approved):
Some thoughts on the particulars of the device identification packet, but I don't think the details are critical. It looks good, prior work really paying off

can/tests/test_messages.cpp
amitlissack: It's just meaningless test data

include/can/core/messages.hpp
amitlissack: Good thoughts. I'll add this comment as todo and merge as is.

PR #81, opened 2021-09-16
refactor(freertos): allow FreeRTOS to be used in both core and firmware

Laura-Danielle:
## Overview

I'm putting this up before the motor task refactor because I figured people might have opinions on it. I want to be able to eventually have all the freertos tasks living in core (including CAN tasks).

## Changelog
- Fixed a small issue with the generic FreeRTOS task class
- Pulled FreeRTOS building out of the `pipettes/firmware`
- Removed a lot of un-used include headers in `pipettes/firmware/main.cpp`

## Review Requests
Everything compiled, but will this break anything to your knowledge? This also should be done in the gantry subproject. Do you want me to do it in this PR or separately?


sfoster1 (approved):
This is good for now but we should try and figure out a way that keeps a bit of built-in pressure to keep core logic separate from rtos config.

Laura-Danielle:
Putting this off for a little while until we have a simulator up in the gantry/pipettes subprojects and we feel comfortable with separation of concerns between core and firmware.

PR #85, opened 2021-09-17
refactor(pipettes): reduce some of the warnings generated in the pipettes project

Laura-Danielle:
## Overview

There are quite a lot of warnings, related to different things. I have tried to reduce warnings that
were unrelated to the HAL layer of things. Some of these files were touched in the FreeRTOS PR, but
I plan to open a separate PR for that once the motor controller task stuff is in a good place. There
were a few CAN related warnings that I did not know how to resolve.

RET-191

### DIFF
--- a/include/can/core/message_writer.hpp
+++ b/include/can/core/message_writer.hpp
@@ -34,8 +34,8 @@ class MessageWriter {
 
   private:
     Writer& writer;
-    std::array<uint8_t, 64> buffer;
-    can_ids::ArbitrationId arbitration_id;
+    std::array<uint8_t, 64> buffer{};
+    can_ids::ArbitrationId arbitration_id{};
 };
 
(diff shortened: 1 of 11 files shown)


include/can/core/message_writer.hpp
```diff
@@ -34,8 +34,8 @@ class MessageWriter {
 
   private:
     Writer& writer;
-    std::array<uint8_t, 64> buffer;
```
amitlissack, line 37: This might be a merge error. I believe I changed 64 to a constant earlier today.

amitlissack (changes requested):

include/can/core/message_writer.hpp
amitlissack: I was wrong.

amitlissack (approved):

PR #86, opened 2021-09-17
refactor(can): refactor day!

amitlissack:
- Move arbitration id data structures to their own file.
- Reduce the verbosity of messages.hpp by defining some base classes.
- Separate CanBus concept into two. There was no reason to mix `send` with `add_filter`.
- Fix a data race in MessageWriter.


Laura-Danielle (commented):
I think I could use a PR demo on the separation of `send` and `filtering`.

sfoster1 (approved):
Sweet

PR #87, opened 2021-09-17
refactor(spi): remove HAL includes where possible

Laura-Danielle:
# Overview

To reduce warnings emitted from macros in the st bsp files, we should try to ensure that we are not
importing hal files unnecessarily in our SPI comms module.

RET-193

# Review requests. Let me know if any syntax or other things seem weird. Will need to test on Monday.


sfoster1 (approved):
Noice

PR #88, opened 2021-09-20
refactor(gantry): add uart 

ahiuchingau:
This PR adds the UART feature that we implemented in the pipettes subproject to gantry.


sfoster1 (approved):
Works for me!

PR #90, opened 2021-09-21
fix(simulation): Can sim uses fd

amitlissack:
@DerekMaggio was integrating our python hardware with our firmware emulation and found that our use of FD in python caused the C++ side to ignore messages.

This PR remedies that.


DerekMaggio:
Successfully tested against emulator using CANFD message.

DerekMaggio (approved):

PR #92, opened 2021-09-21
refactor(pipettes): Can pipettes handlers

amitlissack:
- move eeprom and motor handlers out of can_task.cpp and into `include/pipette/core/message_handlers`. 
- fix a race condition in `MessageWriter`
- add a minimal Move command structure.

This PR is part of a pairing with @Laura-Danielle to address comments in https://github.com/Opentrons/ot3-firmware/pull/78.

More to come in a future PR.


sfoster1 (approved):
Nice, looks good to me. A bit unclear what the "increment" is but it's a WIP so whatever.

PR #93, opened 2021-09-22
fix(cansim): add CAN_CHANNEL environment variable.

amitlissack:
We had been hard coding the channel to `vcan0`. This allows changing it using an environment variable.

### DIFF
--- a/can/simulator/rtos_main.cpp
+++ b/can/simulator/rtos_main.cpp
@@ -85,8 +85,8 @@ static auto dispatcher_task =
  * The socket can reader. Reads from socket and writes to message buffer.
  */
 void can_bus_poll_task(void *) {
-    const char *channel = std::getenv(ChannelEnvironmentVariableName);
-    channel = channel ? channel : DefaultChannel;
+    const char * env_channel_val = std::getenv(ChannelEnvironmentVariableName);
+    auto channel = env_channel_val ? env_channel_val : DefaultChannel;
 
     transport.open(channel);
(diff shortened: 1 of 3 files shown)


can/simulator/rtos_main.cpp
```diff
@@ -81,7 +85,10 @@ static auto dispatcher_task =
  * The socket can reader. Reads from socket and writes to message buffer.
  */
 void can_bus_poll_task(void *) {
-    transport.open("vcan0");
+    const char *channel = std::getenv(ChannelEnvironmentVariableName);
+    channel = channel ? channel : DefaultChannel;
```
sfoster1, line 89: Feels a little gross to be changing a `const char*` even though it's legal; maybe use a temporary?

can/simulator/socket_can.cpp
```diff
@@ -45,6 +45,7 @@ auto SocketCanTransport::open(const char *address) -> bool {
     if (bind(s, (struct sockaddr *)&addr, sizeof(addr)) == -1) {
         return false;
     }
+    std::cout << "Connected to " << address << std::endl;
```
sfoster1, line 48: [nit] Can we just pick one of printf or iostream?

sfoster1 (approved):
Looks good overall, couple nits

can/simulator/socket_can.cpp
amitlissack: OOps.

PR #94, opened 2021-09-22
chore(flash): add pipettes-flash and gantry-flash targets.

amitlissack:
Run without debugger.

### DIFF
--- a/gantry/firmware/CMakeLists.txt
+++ b/gantry/firmware/CMakeLists.txt
@@ -119,3 +119,11 @@ add_custom_target(gantry-debug
         COMMAND gantry
         USES_TERMINAL
         )
+
+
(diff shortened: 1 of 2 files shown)


Laura-Danielle (approved):

sfoster1 (approved):
I think you'll want to make the command `program $<TARGET_FILE:pipettes>;reset;exit`, which will make the device run immediately; otherwise I think you'll have to manually reset after flash.

amitlissack:
> I think you'll want to make the command `program $<TARGET_FILE:pipettes>;reset;exit`, which will make the device run immediately; otherwise I think you'll have to manually reset after flash.

Very nice. I was wondering why I had to manually reset.

PR #97, opened 2021-09-24
refactor(can): remove SetSpeedRequest. change MoveRequest.

amitlissack:
Add add distance and speed to MoveRequest.


sfoster1 (approved):
Nice!

PR #98, opened 2021-09-27
feat(motion-control): add linear system config

ahiuchingau:
This PR adds linear motion system configuration to the Motor Control module. 

The linear motion configuration consists of the following:
1. The **linear mechanical configuration** for the two drive systems
Our linear axes are using either a belt-driven system or a leadscrew-driven system. In order to calculate the `mm_per_revolution` of the system, we need the correct measurements and/or counts of the system's components. For the belts, we need to know the belt pitch, pulley tooth counts. As for the leadscrew, we need to know the lead screw pitch.

2. The steps per revolution of the stepper motor 
3. The microstepping we decide to use

Using the above information, we can then calculate the `steps_per_mm` for our motion system, which will then be used when we need to command the motor to move a specific distance in mm.

Note: The config values used in the subsystems are merely placeholders. Please check the correctness of the equations added as those should not need to be changed once this PR is merged.


include/motor-control/core/linear_motion_system.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <concepts>
+
+namespace lms {
+
+template <typename MC>
+concept MotorMechanicalConfig = requires(MC mc) {
+    {mc.mm_per_rev};
+};
+
+struct BeltConfig {
```
sfoster1, line 12: If we change this struct so that it does not have a user-defined constructor, then it will be a candidate for [aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization) including designated-initialization, which means you could initialize it like
```c++
auto config = BeltConfig{.belt_pitch = 2, .pulley_tooth_count = 4};
```
It's also possible that it would be a [LiteralType](https://en.cppreference.com/w/cpp/named_req/LiteralType) which would allow instances of it to be `constexpr`:
```c++
constexpr auto config = BeltConfig{.belt_pitch=2, .pulley_tooth_count=2};
```
The only thing that might prevent this is the way `mm_per_rev` works, and changing that to a constexpr function as either
- a constexpr member function: `constexpr float mm_per_rev() const { return belt_pitch * pully_count; }`
- a constexpr static member function: `static constexpr float BeltConfig::mm_per_rev(const BeltConfig& bc) const { return bc.belt_pitch * bc.pulley_tooth_count; }`
- a constexpr free function: `constexpr float mm_per_rev(const BeltConfig& bc) { return bc.belt_pitch * bc.pulley_tooth_count; }` (with one version for each of the belt/leadscrew configs)

I think it would be really nice, if ultimately not really difference making, for these to be constexpr-safe.

sfoster1 (changes requested):
Really like the way this lets us split up configuration! I think there's 2.5 more things we should do:
- Make these config structs aggregate types and initialize them with designated-initialization, which will be a bit more clear than the comments
- Add test suites whose primary job is to run some typical configurations through these and then run some calculations _with the same data type_ to move from millimeters to steps and check the accuracy of the translation. The thing I'm worried about here is that with small enough numbers we could get into one of the regimes of ieee754s that have unintuitive accuracy limits, and having some tests that would make us know that would be great
- Make them constexpr, mostly as an intellectual exercise since whether these are compile time or runtime initialized they're still getting used like <5 times ever

include/motor-control/core/linear_motion_system.hpp
ahiuchingau: Totally agree with you!

include/motor-control/core/linear_motion_system.hpp
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#include <concepts>
+
+namespace lms {
+
+template <typename MC>
+concept MotorMechanicalConfig = requires(MC mc) {
```
amitlissack, line 8: I don't get the value of this concept. It's not documented. It doesn't require mm_per_rev to be any particular type.

include/motor-control/core/linear_motion_system.hpp
```diff
@@ -0,0 +1,49 @@
...
+    float get_steps_per_mm() {
+        return (steps_per_rev * microstep) / (mech_config.mm_per_rev);
+    }
+
+    MEConfig& mech_config;
+    float steps_per_rev;
+    float microstep;
+};
+
+template <class Config>
```
amitlissack, line 44: Same as above. What is the value of this concept? What is `get_steps_per_mm` need to return?

amitlissack (commented):
Readability wise, using concepts to define class interfaces is very very poor compared to inheritance. If we are going to continue down this path we should annotate concepts with as much typing and description as possible.

sfoster1 (approved):
Nice!

Laura-Danielle (approved):

PR #99, opened 2021-09-28
refactor(can): use Inheritance for can interface.

amitlissack:
Replace CanBusWriter concept with CanBusWriter abstract base class.


Laura-Danielle (approved):
so much cleaner 😭

sfoster1 (approved):
Nice!

PR #100, opened 2021-09-28
refactor(can): make motor message handler execute actual motor control functions

Laura-Danielle:
## Overview
With this PR, you should be able to send CAN messages to execute actual motor control. As noted with the TODO message, there is definitely some refactors we can do around templating the motor control stuff.

@amitlissack whenever you make it to the office next, I would appreciate us trying to test this together.

### DIFF
--- a/include/can/core/messages.hpp
+++ b/include/can/core/messages.hpp
@@ -113,22 +113,18 @@ struct GetStatusResponse : BaseMessage<MessageId::get_status_response> {
 };
 
 struct MoveRequest : BaseMessage<MessageId::move_request> {
-    uint32_t distance;
-    uint32_t speed;
+    uint32_t steps;
 
     template <bit_utils::ByteIterator Input, typename Limit>
     static auto parse(Input body, Limit limit) -> MoveRequest {
-        uint32_t distance = 0;
-        uint32_t speed = 0;
-        body = bit_utils::bytes_to_int(body, limit, distance);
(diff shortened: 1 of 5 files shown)


sfoster1 (approved):
Nice

include/can/core/messages.hpp
```diff
@@ -113,22 +113,18 @@ struct GetStatusResponse : BaseMessage<MessageId::get_status_response> {
 };
 
 struct MoveRequest : BaseMessage<MessageId::move_request> {
-    uint32_t distance;
-    uint32_t speed;
+    uint32_t steps;
```
amitlissack, line 116: Hold on here. I thought we wanted the can message to contain speed and distance?

amitlissack (changes requested):
I don't get why the can move request was changed.

include/can/core/messages.hpp
Laura-Danielle: I changed it for the demo. If we want to keep it as we expect it to be in the future, that's fine but we'll need to use UART for the demo then.

amitlissack: Understood.

You'll have to fix some unit tests I believe.

amitlissack (approved):

include/can/core/messages.hpp
Laura-Danielle: yep I'm in the midst of doing that now.

ahiuchingau (commented):
cool!

PR #101, opened 2021-10-04
fix(can): race condition

amitlissack:
Fixes a race condition between for accessing the `RxHeader`. 

I'm not sure if using a lock is better than allocating the `RxHeader` on the stack.


sfoster1 (approved):
Nice! I think having a lock here is the right thing to do because it's more explicit.

PR #102, opened 2021-10-08
feat(motion-control): Use fixed point math to pulse the motor

Laura-Danielle:
## Overview
This is the first milestone for what the eventual production level code will look like for motion control. We are now converting mm -> steps to pulse the motor.

## Review Requests
1. Definitely check the math
2. Let me know if there are other tests that should be formatted etc.

## Notes
- Removing UART stuff immediately after this bc I'm tired of having to change a million files.

### DIFF
--- a/include/motor-control/core/motion_controller.hpp
+++ b/include/motor-control/core/motion_controller.hpp
@@ -77,7 +77,6 @@ class MotionController {
     uint32_t acc = 0x0;
     uint32_t speed = 0x0;  // mm/s
     uint32_t dist = 0x0;
-    uint32_t inc_multiplier = 0x0;
     bool direction = true;  // direction true: forward, false: backward
     uint32_t steps_per_mm;
     lms::LinearMotionSystemConfig<MEConfig> linear_motion_sys_config;


include/motor-control/core/motion_controller.hpp
```diff
@@ -36,32 +35,35 @@ class MotionController {
...
-    void set_direction(bool d) {
-        if (d) {
-            set_pin(hardware_config.direction);
-        } else {
-            reset_pin(hardware_config.direction);
-        }
-        direction = d;
+    void set_speed(uint32_t s) {
+        speed = s;
+        //        inc_multiplier = speed * steps_per_mm / (2 * clk_frequency);
```
sfoster1, line 56: remove?

include/motor-control/core/motion_controller.hpp
```diff
@@ -36,32 +35,35 @@ class MotionController {
...
+
     void setup() {
-        start_motor_handler(queue);
+        start_motor_handler(&queue);
         timer_interrupt_start();
-        steps_per_mm = linear_motion_sys_config.get_steps_per_mm();
+        // convert steps to mm to fixed point here
+        steps_per_mm =
+            static_cast<int32_t>(linear_motion_sys_config.get_steps_per_mm());
+        //        speed = clk_frequency / 2 / steps_per_mm;
```
sfoster1, line 49: remove?

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -37,45 +37,85 @@ namespace motor_handler {
  * GenericQueue -> A FreeRTOS queue of any shape.
  */
 
+constexpr const int clk_frequency = 85000000 / (5001 * 2);
```
sfoster1, line 40: I wonder if this could become part of the motor config? that's created in the right place (in the task root) and has similar hardware config stuff

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -37,45 +37,85 @@ namespace motor_handler {
...
+    MotorInterruptHandler&& operator=(MotorInterruptHandler&&) = delete;
+    MotorInterruptHandler(MotorInterruptHandler&) = delete;
+    MotorInterruptHandler(MotorInterruptHandler&&) = delete;
 
     void set_message_queue(GenericQueue* g_queue) { queue = g_queue; }
 
     bool has_messages() { return queue->has_message_isr(); }
 
-    bool can_step() { return (has_active_move && inc < buffered_move.steps); }
+    bool can_step() { return (step_count < buffered_move.target_position); }
```
sfoster1, line 59: this probably wants to be a `!=` to support bidirectional motion

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -37,45 +37,85 @@ namespace motor_handler {
...
+        // (TODO: lc) We should check the direction (and set respectively)
+        // the direction pin for the motor once a move is being pulled off the
+        // queue stack. We'll probably want to think about moving the hardware
+        // pin configurations out of motion controller.
     }
 
     void finish_current_move() {
+        // (TODO: lc) We need to handle overflow cases on the position
+        // tracker.
+        position_tracker += step_count;
```
sfoster1, line 78: isn't the position tracker the same thing as the step count?

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -37,45 +37,85 @@ namespace motor_handler {
...
+            finish_current_move();
+            return false;
+        }
+        return false;
+    }
 
     void reset() {
         queue->reset();
-        inc = 0x0;
+        step_count = 0x0;
```
sfoster1, line 99: we shouldn't reset the step count. by keeping the step count and the position tracker the same thing, we should automatically account for leftover sub-step positioning.

pipettes/tests/test_motor_pulse.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+#include <iostream>
+
+#include "catch2/catch.hpp"
+#include "motor-control/core/motor_interrupt_handler.hpp"
+#include "pipettes/tests/mock_message_queue.hpp"
+
+using namespace motor_handler;
+
+TEST_CASE("move a full rotation") {
```
sfoster1, line 11: These tests aren't really testing the right layers. With this kind of step math, we really need to test the _numeric_ edge cases, not the motion-system edge cases, because it's so limited to numerics; and, whenever we start doing direction and enable in the same subsystem, those lines too. So for instance, we should have tests that exercise
- Behavior starting from `step_count=0` and going positive and negative;
- Behavior starting above 1.0 and going through 0 negative going
- Behavior starting in (1.0, 0.0) somewhere and moving through 0.0 negative going
- The same with below -1.0 and up through 0; and between 0 and -1.0 and up through 0
- Maybe all the way from > 1.0 to < -1.0 and the reverse

We can keep the overflow cases, but maybe transition to just detecting and raising an error; if we have the step counter and position tracker be the same, it can't get that big anyway.

pipettes/tests/test_motor_pulse.cpp
```diff
@@ -0,0 +1,136 @@
...
+            handler.pulse();
+            if (!handler.has_active_move) {
+                running = false;
+            }
+        }
+        REQUIRE(handler.get_current_position() == temp_steps);
+    };
+}
+
+TEST_CASE("move less than a full step") {
```
sfoster1, line 49: This one's good

pipettes/tests/test_motor_pulse.cpp
```diff
@@ -0,0 +1,136 @@
...
+            handler.pulse();
+            if (!handler.has_active_move && !handler.has_messages()) {
+                running = false;
+            }
+        }
+        REQUIRE(handler.get_current_position() == temp_steps + temp_steps);
+    }
+}
+
+TEST_CASE("move multiple full rotations in a row") {
```
sfoster1, line 88: we should be careful how we name/talk about stuff in this layer; things like rotations or microstepping really aren't relevant to this controller. there are only steps

include/motor-control/core/motor_messages.hpp
```diff
@@ -2,8 +2,19 @@
...
 #include <cstdint>
 
+typedef int32_t sq0_31;  // 0: signed bit,  1-31: fractional bits
+typedef int64_t
+    sq32_31;  // 0: signed bit, 1-32: integer bits, 33-64: fractional bits
+
 namespace motor_messages {
+
+struct CanMove {
+    int32_t target_position;  // in mm
```
sfoster1, line 12: we might want to reframe this; integer mm are almost certainly not the right unit here. either fixed- or floating-point mm, or integer microns are probably better.

include/motor-control/core/motion_controller.hpp
```diff
@@ -36,32 +35,35 @@ class MotionController {
...
-    }
 
-    void move(const Move& msg) {
-        set_pin(hardware_config.enable);
-        set_pin(hardware_config.direction);
+    void move(const CanMove& can_msg) {
+        // TODO: set direction in
+        //        motor interrupt handler instead
+        int64_t converted_steps =
+            static_cast<int64_t>(can_msg.target_position * steps_per_mm) << 31;
```
amitlissack, line 65: What is significance of 31?

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -37,45 +37,85 @@ namespace motor_handler {
...
+    MotorInterruptHandler(MotorInterruptHandler&&) = delete;
 
     void set_message_queue(GenericQueue* g_queue) { queue = g_queue; }
 
     bool has_messages() { return queue->has_message_isr(); }
 
-    bool can_step() { return (has_active_move && inc < buffered_move.steps); }
+    bool can_step() { return (step_count < buffered_move.target_position); }
+
+    bool tick() {
```
amitlissack, line 61: Please comment every line in a complex function.

Also add doc strings

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -37,45 +37,85 @@ namespace motor_handler {
...
+        // tracker.
+        position_tracker += step_count;
         has_active_move = false;
         buffered_move = Move{};
-        inc = 0x0;
+        step_count = 0x0;
     }
 
-    void increment_counter() { inc++; }
+    bool pulse() {
```
amitlissack, line 84: What does this function do? It's very complex and it's name is not descriptive. 

What does the return value mean?

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -37,45 +37,85 @@ namespace motor_handler {
...
+    void set_current_position(sq32_31 pos_tracker) {
+        position_tracker = pos_tracker;
+    }
+
+    Move* get_buffered_move() { return &buffered_move; }
+
   private:
-    uint32_t inc = 0x0;
+    sq0_31 steps_per_tick = 0x20000000;  // 0.5 steps per tick
+    sq32_31 tick_flag = 0x80000000;
```
amitlissack, line 114: Why this number? Please make a constant.

pipettes/tests/test_motor_pulse.cpp
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+        REQUIRE(handler.get_current_position() == temp_steps);
+    }
+
+    GIVEN("two fractional moves") {
+        // equivalent of 28.23 steps
+        sq32_31 temp_steps = 2823LL << 29;
+        Move msg1 = Move{.target_position = temp_steps};
+        queue.try_write(msg1);
```
amitlissack, line 74: Does a message queue have to be a part of these tests? 

Isn't there a method that gets a `Move` message that can do the computation?

pipettes/tests/test_motor_pulse.cpp
```diff
@@ -0,0 +1,136 @@
...
+        queue.try_write(msg1);
+        REQUIRE(handler.get_current_position() == 0);
+        bool running = true;
+        while (running) {
+            handler.pulse();
+            if (!handler.has_active_move && !handler.has_messages()) {
```
amitlissack, line 80: It's unfortunate that it's possible to make this test get into an infinite loop rather than fail.

pipettes/tests/test_motor_pulse.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+TEST_CASE("move multiple full rotations in a row") {
+    MotorInterruptHandler<mock_message_queue::MockMessageQueue> handler{};
+    mock_message_queue::MockMessageQueue<Move> queue;
+    handler.set_message_queue(&queue);
+
+    GIVEN("Movement in incrememnts of 200 steps") {
+        sq32_31 temp_steps = 0xC8LL << 31;
+        Move msg1 = Move{.target_position = temp_steps};
+        queue.try_write(msg1);
```
amitlissack, line 96: Same comment as above. Can we rewrite this test without  message queue and loop?

pipettes/tests/test_motor_pulse.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+}
+
+TEST_CASE("overflow position") {
+    MotorInterruptHandler<mock_message_queue::MockMessageQueue> handler{};
+    mock_message_queue::MockMessageQueue<Move> queue;
+    handler.set_message_queue(&queue);
+
+    GIVEN("Move past the largest possible value for position") {
+        sq32_31 max_position = 0x7FFFFFF76B48C000;
```
amitlissack, line 117: Why this number? Is it a constant?

include/motor-control/core/motion_controller.hpp
```diff
@@ -36,38 +35,37 @@ class MotionController {
...
-        // pass total steps to IRS
-    }
 
-    void move(const Move& msg) {
-        set_pin(hardware_config.enable);
-        set_pin(hardware_config.direction);
+    void move(const CanMove& can_msg) {
+        // TODO: set direction in
+        //        motor interrupt handler instead
+        int64_t converted_steps =
```
sfoster1, line 61: We should make this a `uint64_t`

include/can/core/messages.hpp
```diff
@@ -113,18 +113,18 @@ struct GetStatusResponse : BaseMessage<MessageId::get_status_response> {
...
 struct MoveRequest : BaseMessage<MessageId::move_request> {
-    uint32_t steps;
+    int32_t target_position;
 
     template <bit_utils::ByteIterator Input, typename Limit>
     static auto parse(Input body, Limit limit) -> MoveRequest {
-        uint32_t steps = 0;
-        body = bit_utils::bytes_to_int(body, limit, steps);
-        return MoveRequest{{}, steps};
+        int32_t target_position = 0;
```
sfoster1, line 120: if we're going to keep positions such that one axis extent is always 0 and the other is positive, let's keep absolute positions as unsigned types like `uint32_t`

include/motor-control/core/motion_controller.hpp
```diff
@@ -79,8 +77,9 @@ class MotionController {
     uint32_t acc = 0x0;
     uint32_t speed = 0x0;  // mm/s
     uint32_t dist = 0x0;
+    uint32_t inc_multiplier = 0x0;
     bool direction = true;  // direction true: forward, false: backward
-    float steps_per_mm;
+    int32_t steps_per_mm;
```
sfoster1, line 82: can be unsigned

include/motor-control/core/motion_controller.hpp
```diff
@@ -79,8 +77,9 @@ class MotionController {
     uint32_t acc = 0x0;
     uint32_t speed = 0x0;  // mm/s
     uint32_t dist = 0x0;
+    uint32_t inc_multiplier = 0x0;
```
sfoster1, line 80: unneeded?

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -37,45 +36,135 @@ namespace motor_handler {
...
+    bool overflow(sq32_31 current, sq32_31 future) {
+        /*
+         * Check whether the position has overflowed. Return true if this has
+         * happened, and false otherwise.
+         *
+         * (TODO lc): If an overflow has happened, we should probably notify the
+         * server somehow that the position needed to be capped.
+         * Note: could not get the built-ins to work as expected.
+         */
+        return bool((current ^ future) & overflow_flag);
```
sfoster1, line 148: so this is us implicitly limiting this to a uint63? honestly I think that's fine and perhaps better but we should be up front about it

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -37,45 +36,135 @@ namespace motor_handler {
...
+    Move* get_buffered_move() { return &buffered_move; }
+
+    void set_buffered_move(Move new_move) { buffered_move = new_move; }
+
   private:
-    uint32_t inc = 0x0;
+    const sq0_31 steps_per_tick = 0x40000000;  // 0.5 steps per tick
+    const sq32_31 tick_flag = 0x80000000;
+    const uint64_t overflow_flag = 0x8000000000000000;
+    sq32_31 position_tracker = 0x0;
```
sfoster1, line 165: this is a `q31_31` (`uint64_t`) now right? losing a bit to check for overflow, and let's make it unsigned

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -37,45 +36,135 @@ namespace motor_handler {
...
+    const uint32_t steps_per_sec = clk_frequency * steps_per_tick;
```
sfoster1, line 166: unneeded?

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -37,45 +36,135 @@ namespace motor_handler {
...
+            return false;
+        }
+        return bool((old_position ^ position_tracker) & tick_flag);
+    }
 
     void update_move() {
-        finish_current_move();
-        get_move(&buffered_move);
+        has_active_move = queue->try_read_isr(&buffered_move);
+        buffered_move.velocity += steps_per_tick;
```
sfoster1, line 86: what's this about?

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -37,45 +36,135 @@ namespace motor_handler {
...
+        position_tracker = pos_tracker;
+    }
+
+    Move* get_buffered_move() { return &buffered_move; }
+
+    void set_buffered_move(Move new_move) { buffered_move = new_move; }
+
   private:
-    uint32_t inc = 0x0;
+    const sq0_31 steps_per_tick = 0x40000000;  // 0.5 steps per tick
```
sfoster1, line 162: we get this from the move now don't we?

include/motor-control/core/motor_messages.hpp
sfoster1: and let's make it unsigned

PR #104, opened 2021-10-12
Ret 169 za bring up

sotaoverride:
ZA stage (head) bring up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .idea
 
 CMakeFiles
+CMakeCache.txt
 
 cmake_install.cmake
 
--- a/README.md
+++ b/README.md
@@ -1,51 +1,81 @@
 # ot3_firmware
 
-This repository holds the firmware for the OT-3 and all of its peripheral systems. **Note** that the cmake directory is actually a subtree of cmake-utils. Do not make changes to the `cmake-utils` directory in the ot3_firmware repository. Instead, navigate to the [CMake Utils Repository](https://github.com/Opentrons/cmake-utils). 
+This repository holds the firmware for the OT-3 and all of its peripheral systems. **Note** that the cmake directory is
+actually a subtree of cmake-utils. Do not make changes to the `cmake-utils` directory in the ot3_firmware repository.
+Instead, navigate to the [CMake Utils Repository](https://github.com/Opentrons/cmake-utils).
 
 ## Structure of directories
+
 Aside from the common directory, each repository should contain a `firmware`, `include`, `src`, `tests` folder.
+
 1. `firmware` should hold code that will be controlling the peripheral system.
-2. `include` should hold the majority of the header files you will need. This folder should include another subset of directories labeled `firmware`, `src` and `tests` for import cleaness.
-3. `src` should include any libraries, 3D party or otherwise, that are required for `firmware` to control the peripheral system.
+2. `include` should hold the majority of the header files you will need. This folder should include another subset of
+   directories labeled `firmware`, `src` and `tests` for import cleaness.
+3. `src` should include any libraries, 3D party or otherwise, that are required for `firmware` to control the peripheral
+   system.
 4. `tests` should include tests for the firmware folder.
 
 ## Setup
+
 To setup this directory to run on a nucleo board, you should run:
+
 1. `cmake --preset=cross .`
 2. `cmake --build ./build-cross --target <TARGET>`
 
 To setup this directory to run tests, you should run:
-1. `cmake --preset=host .`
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -9,3 +9,4 @@ build-cross/
 stm32-tools/
 build-host/
 cmake-build-debug/
+node_modules
```
sfoster1, line 12: why do we need this?

head/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+# head source tree
+
+# add_subdirectory(src)
```
sfoster1, line 3: delete?

head/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+
+if (${CMAKE_CROSSCOMPILING})
```
sfoster1, line 5: don't seem to be adding `core` anywhere

sfoster1 (changes requested):
We should
- Get rid of the uart stuff
- Clean up some commented out stuff in the cmake
- Add CI in github actions

.gitignore
sfoster1: oh that makes sense i guess

head/core/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+# This CMakeLists.txt handles compiling all the parts of the head
+# module that are portable between host and cross compilation as a static
+# library. It is included in both host and cross configurations.
+
+add_library(head-core STATIC
+  ${CMAKE_CURRENT_SOURCE_DIR}/
```
sfoster1, line 6: you can't really do this, you need to specify files. and as you've probably noticed from the failing builds, it needs a file to also determine the language

amitlissack (approved):
Looks good! 

Tested with CAN analyzer and FW happily responded to device info request.

head/core/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+add_library(head-core STATIC
+        head.cpp
+        )
+
+set_target_properties(head-core
+  PROPERTIES CXX_STANDARD 20
+             CXX_STANDARD_REQUIRED TRUE)
+
+target_include_directories(head-core
+  PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../include)
```
Laura-Danielle, line 14: This is also the wrong path for the include directory -- if you copied from gantry that project hasn't been updated in awhile. The path specified here would point to an include directory inside of the `head` subproject.

```
target_include_directories(head
        PUBLIC ${CMAKE_SOURCE_DIR}/include)
```

head/tests/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+# this CMakeLists.txt file is only used when host-compiling to build tests
+find_package(Catch2 REQUIRED)
+include(CTest)
+include(Catch)
+include(AddBuildAndTestTarget)
+
+add_executable(
+        head
+        test_main.cpp
+        test_linear_motion_config.cpp
```
Laura-Danielle, line 10: I don't know if we need to run this test file multiple times.

include/head/core/communication.hpp
```diff
@@ -0,0 +1,132 @@
+#pragma once
+
+#include <array>
+#include <span>
+
+#include "common/core/bit_utils.hpp"
+#include "common/core/io.hpp"
```
Laura-Danielle, line 7: This is leftover from a uart task which we don't have anymore. We should delete this file.

include/head/core/head_messages.hpp
```diff
@@ -0,0 +1,47 @@
+#pragma once
+
+#include <variant>
+
+#include "motor-control/core/motor_messages.hpp"
```
Laura-Danielle, line 5: Same with this file. It should be deleted.

head/core/CMakeLists.txt
sotaoverride: So there is an include directory at root level, with sub directories for can, pipettes etc in it. Should we get rid of that and each subproject would have an include of their own nested inside of it? I figured the idea was for all subprojects to use a common include..

head/core/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+# This CMakeLists.txt handles compiling all the parts of the head
+# module that are portable between host and cross compilation as a static
+# library. It is included in both host and cross configurations.
+
+add_library(head-core STATIC
+        head.cpp
```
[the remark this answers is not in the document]
sotaoverride: yes, it's on the same level as the CMakeLists.txt , so I think thats how this just works.

head/core/CMakeLists.txt
sotaoverride: thanks, finally makes sense to me. [0230e63](https://github.com/Opentrons/ot3-firmware/pull/104/commits/0230e63cf179970f6c151b24b1296bbe0661a901)

PR #110, opened 2021-10-14
refactor(ot3-firmware/common): remove uart from ot3-firmware/common

sotaoverride:
UART -> CAN


Laura-Danielle (changes requested):
For now, we wanted to leave the uart HAL stuff in place in case we wanted to re-implement some other feature using uart.

The files that were in [this pr](https://github.com/Opentrons/ot3-firmware/pull/104) had to do a basic communication protocol that Amit set-up before CAN was implemented.

If you want to put a note about that somewhere, that is fine with me.

PR #112, opened 2021-10-15
feat(motor-control): add separate motor enable and disable messages

Laura-Danielle:
## Overview
We should be able to enable and disable motors via CAN messaging. I also added in a default velocity until we incorporate speed into the CAN message for moves.


include/can/core/message_handlers/motor.hpp
```diff
@@ -48,6 +48,14 @@ class MotorHandler {
         motor.motion_controller.move(mv_msg);
     }
 
+    void visit(EnableMotorRequest &m) {
```
sotaoverride, line 51: in the python can_comms script is this the setup_request message?

Laura-Danielle: No, I have to add new messages, which I did [here](https://github.com/Opentrons/opentrons/pull/8531).

PR #115, opened 2021-10-20
feat(can): move group implementation

amitlissack:


include/motor-control/core/motion_group.hpp
```diff
@@ -1,2 +1,98 @@
...
+class MoveGroup {
+  public:
+    using MoveTypes =
+        std::variant<std::monostate, can_messages::AddLinearMoveRequest>;
+
+    MoveGroup() {}
+
+    /**
+     *
+     * @param move
```
sfoster1, line 22: we should probably either get rid of these comments or fill them oiut the rest of the way

include/motor-control/core/motion_group.hpp
```diff
@@ -1,2 +1,98 @@
...
+        storage[move.seq_id] = move;
+        return true;
+    }
+
+    /**
+     *
+     * @return
+     */
+    auto size() const -> std::size_t {
+        int count = 0;
```
sfoster1, line 38: [nit] might as well make this a `size_t`

include/motor-control/core/motion_group.hpp
```diff
@@ -1,2 +1,98 @@
...
+
+    /**
+     *
+     * @param seq_id
+     * @return
+     */
+    const MoveTypes& get_move(uint8_t seq_id) const { return storage[seq_id]; }
+
+    auto get_duration() const -> uint32_t {
+        auto dash_fold = [](uint32_t accum, const MoveTypes& b) -> uint32_t {
```
sfoster1, line 75: [nit] Let's change the name of this lambda, it's called that because in the example it's implementing the equivalent of the python `'-'.join(strs)`

sfoster1 (approved):
Some nits but overall looks solid

PR #126, opened 2021-10-28
refactor(motor-control): update move request

ahiuchingau:
We're now passing in `duration` (number of ticks) instead of `target position` (number of steps) in the move request. Velocity and acceleration in the move request are now also getting passed down to the motor interrupt handler.


sfoster1 (approved):
Nice work!

PR #127, opened 2021-11-01
feat(motor-control): change motor direction

ahiuchingau:
This PR allows the motor interrupt handler to change the motor direction by evaluating the buffered move's velocity. If it changes direction from the previous step, the direction pin would be toggled before the next motor step. 

We should (1) test with an actual motor, (2) make sure the direction pin is set at least 20 nanoseconds before the motor takes a step.


sfoster1 (changes requested):
we should have some tests

also
-  i think it's probably a better idea to set the direction specifically rather than toggling (e.g. call `set_direction_pin(bool(velocity & DIRECTION_BIT))`, no state, easy to test
- just do it every time we load a move, easier to test

gantry/firmware/step_motor.cpp
```diff
@@ -11,6 +11,11 @@ static auto handler_class = motor_handler::MotorInterruptHandler<
 
 void step_motor() {
     if (handler_class.pulse()) {
+        if (handler_class.set_direction_pin(get_direction_bit())) {
```
sfoster1, line 14: wait hold on. we can just make this fully dependent on the velocity. we don't have to read the gpio's current status at all. all we have to do is `if (current_move.velocity > 0 ) { turn_on_direction_pin(); } else { turn_off_direction_pin(); }`. it doesn't matter if we set it every tick, it's just writing a bit. lets us keep it super super simple

sfoster1 (changes requested):
one more tiny change and we can simplify a bunch of the direction setting I think

sfoster1 (changes requested):
One more last thing - we should have a test for the new `set_direction_pin()` call based on the sign of the velocity.

Laura-Danielle (approved):

PR #130, opened 2021-11-09
refactor(pipettes): port all GPIO pins to match the pipette main board

Laura-Danielle:
## Overview

We want to start using the pipette main board for firmware development since it has the new MCU
embedded in it already.

closes RET-232


sfoster1 (approved):
Lgtm!

PR #132, opened 2021-11-09
feat(can): handle motor status request

ahiuchingau:
This allows the status of the currently executed move to be read via can.

### DIFF
--- a/head/firmware/step_motor.cpp
+++ b/head/firmware/step_motor.cpp
@@ -28,3 +28,12 @@ void start_motor_handler(
 }
 
 void reset_motor_handler() { handler_class.reset(); }
+
+MoveStatus get_move_status() {
+    auto buffered_move = handler_class.get_buffered_move();
+    MoveStatus msg = {.group_id = buffered_move.group_id,
+                      .seq_id = buffered_move.seq_id,
+                      .remaining_duration = handler_class.get_remaining_ticks(),
+                      .current_position = handler_class.get_current_position()};
+    return msg;
+}
--- a/include/can/core/messages.hpp
+++ b/include/can/core/messages.hpp
@@ -5,14 +5,13 @@
 
 #include "common/core/bit_utils.hpp"
 #include "ids.hpp"
-#include "motor-control/core/motor_messages.hpp"
 #include "parse.hpp"
 
 namespace can_messages {
 
 using namespace can_ids;
 
-using ticks = uint32_t;
+using can_ticks = uint32_t;
 using um_per_tick = int32_t;
 using um_per_tick_sq = int32_t;
 
@@ -128,13 +127,13 @@ struct GetStatusResponse : BaseMessage<MessageId::get_status_response> {
 };
 
 struct MoveRequest : BaseMessage<MessageId::move_request> {
-    ticks duration;
+    can_ticks duration;
     um_per_tick velocity;
     um_per_tick_sq acceleration;
 
     template <bit_utils::ByteIterator Input, typename Limit>
     static auto parse(Input body, Limit limit) -> MoveRequest {
-        ticks duration = 0;
+        can_ticks duration = 0;
         um_per_tick velocity = 0;
         um_per_tick_sq acceleration = 0;
         body = bit_utils::bytes_to_int(body, limit, duration);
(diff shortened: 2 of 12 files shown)


head/firmware/step_motor.cpp
```diff
@@ -27,4 +27,13 @@ void start_motor_handler(
     handler_class.set_message_queue(queue, completed_queue);
 }
 
-void reset_motor_handler() { handler_class.reset(); }
\ No newline at end of file
+void reset_motor_handler() { handler_class.reset(); }
+
+MoveStatus get_move_status() {
+    auto buffered_move = handler_class.get_buffered_move();
```
sfoster1, line 33: This is a data race with `buffered_move.velocity` - it can be altered during the copy at the return of `get_buffered_move`. Nothing is querying velocity, so it shouldn't affect anything, but it's kind of a bad habit. Let's at least note it down so nobody at some point tries to add velocity querying and gets weird results because of it.

head/firmware/step_motor.cpp
```diff
@@ -27,4 +27,13 @@ void start_motor_handler(
...
+    MoveStatus msg = {.group_id = buffered_move.group_id,
+                      .seq_id = buffered_move.seq_id,
+                      .remaining_duration = handler_class.get_remaining_ticks(),
```
sfoster1, line 36: These are _also_ data races, and they _are_ in fact being used. Every so often, you're going to get the motor interrupt firing during the execution of this call, and the data fetched by `get_remaining_ticks` and/or `get_current_position` is going to get corrupted. This access needs to be protected somehow, and protected in a way that cannot significantly delay execution of the interrupt service routine.

include/can/core/messages.hpp
```diff
@@ -128,13 +127,13 @@ struct GetStatusResponse : BaseMessage<MessageId::get_status_response> {
...
 struct MoveRequest : BaseMessage<MessageId::move_request> {
-    ticks duration;
+    can_ticks duration;
     um_per_tick velocity;
     um_per_tick_sq acceleration;
 
     template <bit_utils::ByteIterator Input, typename Limit>
     static auto parse(Input body, Limit limit) -> MoveRequest {
-        ticks duration = 0;
+        can_ticks duration = 0;
```
amitlissack, line 136: Why the `can_ticks` name?

include/can/core/messages.hpp
```diff
@@ -215,18 +214,66 @@ struct ReadFromEEPromResponse : BaseMessage<MessageId::read_eeprom_response> {
...
+    uint8_t padding;
+
+    template <bit_utils::ByteIterator Input, typename Limit>
+    static auto parse(Input body, Limit limit) -> GetMoveStatusResponse {
+        uint8_t group_id = 0;
+        uint8_t seq_id = 0;
+        uint64_t remaining_duration = 0;
+        uint64_t current_position = 0;
+        uint8_t node_id = 0;
+        uint8_t padding =
```
amitlissack, line 235: Do you strictly need the padding on the CPP side?

amitlissack (commented):
Some questions

sfoster1 (changes requested):
The access of the ticks and current position counter is a data race. The data retrieved by those methods will eventually be corrupted when the motor interrupt happens to occur during one of the calls. The accesses need to be protected, in one of a couple ways:
- The worst: have a critical section that masks off the timer interrupt in `get_move_status()`. This will always ensure the interrupt doesn't occur. Or have a mutex protection or something. This is bad because it can delay execution of the motor interrupt, and we absolutely do not ever want that to happen if it can possibly be avoided - the timing is very important
- The easiest, but lowest-fidelity: have some caching layer somewhere and have this can message actually ask something like "what was the last position you recorded", i.e. before the interrupt started executing the current move. This will be the same thing when nothing is moving, and when you are moving, does the exact current step really matter?
- The hardest: have another queue or something that the interrupt can use to periodically send status updates back to whatever task owns this functionality - the can task i guess.

One of these needs to happen though.

include/can/core/messages.hpp
ahiuchingau: Because firmware is receiving the duration in 32 bits but sending out 64 bits remaining duration, both are in ticks.

include/can/core/messages.hpp
ahiuchingau: I thought it might be easier to debug if the padding is on the CPP side as well, but we can totally omit it.

PR #133, opened 2021-11-09
Ff board head pin setup RET-170

sotaoverride:
Z/A ff board bring up
 - - separating out SPI to manage Z/A related pin configurations etc


common/firmware/head-ff/spi/spi.c
```diff
@@ -0,0 +1,243 @@
+#include "common/firmware/spi.h"
```
sfoster1, line 1: Can we just have this in the `head` directory? i don't think we need to retain support for the `AxisController` development platform boards.

common/firmware/head-ff/spi/spi.c
```diff
@@ -0,0 +1,243 @@
...
+SPI_HandleTypeDef handle, handle_SPI2, handle_SPI3;
+
+/**
+ * @brief SPI MSP Initialization
+ * This function configures SPI for the Z/A axis motors
+ * @param hspi: SPI handle pointer
+ * @retval None
+ */
+void HAL_SPI_MspInit(SPI_HandleTypeDef* hspi) {
+    // PB14 -> MISO
```
sfoster1, line 15: got some redundant comments here

common/firmware/head-ff/spi/spi.c
```diff
@@ -0,0 +1,243 @@
...
+
+void SPI2_init() {
+    handle_SPI2 = MX_SPI2_Init();
+}
+
+void SPI3_init() {
+    handle_SPI3 = MX_SPI3_Init();
+}
+
+void Set_CS_Pin() {
```
sfoster1, line 224: why do we set both here? i think we probably want these interfaces fully separate

common/firmware/head-ff/spi/spi.c
```diff
@@ -0,0 +1,243 @@
...
+    HAL_GPIO_WritePin(GPIOA, GPIO_PIN_4, GPIO_PIN_RESET);
+}
+
+void hal_transmit_receive(uint8_t* transmit, uint8_t* receive, uint16_t buff_size, uint32_t timeout) {
+    HAL_SPI_TransmitReceive(&handle, transmit, receive, buff_size, timeout);
+}
+void hal_transmit_receive_SPI2(uint8_t* transmit, uint8_t* receive, uint16_t buff_size, uint32_t timeout) {
+    HAL_SPI_TransmitReceive(&handle_SPI2, transmit, receive, buff_size, timeout);
+}
+void hal_transmit_receive_SPOI3(uint8_t* transmit, uint8_t* receive, uint16_t buff_size, uint32_t timeout) {
```
sfoster1, line 241:
```suggestion
void hal_transmit_receive_SPI3(uint8_t* transmit, uint8_t* receive, uint16_t buff_size, uint32_t timeout) {
```

common/firmware/spi/spi.c
```diff
@@ -81,6 +80,38 @@ void HAL_SPI_MspDeInit(SPI_HandleTypeDef* hspi) {
     }
 }
 
+
```
sfoster1, line 83: This seems redundant - is it compiled?

sfoster1 (changes requested):
I think the handling of SPI here isn't quite what we want. Unfortunately we might need a bit of a refactor here to make sure that the shared spi stuff is much more limited since different systems have different SPI configurations.

PR #134, opened 2021-11-10
feat(sim): gantry simulator. simulates motor interrupt handler.

amitlissack:
This is a hacked together simulator that creates csv data (pulse, position, direction) from a move group configured in code.

Very much the beginnings of this kind of work, but it can be useful even in this state.


Laura-Danielle (approved):
Good first step!

sfoster1 (approved):
Nice

ahiuchingau (approved):
wow wow wow

PR #135, opened 2021-11-10
refactor(can): update ack message

ahiuchingau:
Adds current position to the ack message.


sfoster1 (approved):
Like it, love it, can't get enough of it

PR #136, opened 2021-11-10
feat(can): add clear all move groups request.

amitlissack:
Add `ClearAllMoveGroupsRequest` to empty all move groups. 

Should consider just removing the message that clears them one at a time. Not clear what the use case is for that.


sfoster1 (approved):
Nice. Let's remove the one at a time one

PR #138, opened 2021-11-11
chore(can): socket can reusability

amitlissack:
Extracting the socketcan code to its own include directory.  The classes are now templates.


can/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ else()
     if("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Linux")
         # Simulator requires linux only kernel interfaces
         add_subdirectory(simulator)
```
Laura-Danielle, line 7: Would it mess things up to make the simulator the static library here rather than creating a separate directory?

amitlissack: I decided to not make a shared library. The reason is that the socket can code depended on FreeRTOS which is a pain using shared libraries due to the freertosconfig.h requirement.

I removed the dependency and made the socket can stuff templates.

can/simulator/main.cpp
```diff
@@ -5,9 +5,10 @@
 #include "can/core/freertos_can_dispatch.hpp"
 #include "can/core/message_writer.hpp"
 #include "can/core/messages.hpp"
-#include "can/simulator/sim_canbus.hpp"
-#include "can/simulator/socket_can.hpp"
+#include "can/simlib//sim_canbus.hpp"
```
sfoster1, line 8:
```suggestion
#include "can/simlib/sim_canbus.hpp"
```

include/common/core/freertos_synchronization.hpp
```diff
@@ -61,4 +61,19 @@ class FreeRTOSMutexFromISR {
...
+  public:
+    FreeRTOSCriticalSection() {}
+    FreeRTOSCriticalSection(const FreeRTOSCriticalSection &) = delete;
+    FreeRTOSCriticalSection(const FreeRTOSCriticalSection &&) = delete;
+    FreeRTOSCriticalSection &operator=(const FreeRTOSCriticalSection &) =
+        delete;
+    FreeRTOSCriticalSection &&operator=(const FreeRTOSCriticalSection &&) =
+        delete;
+
+    void acquire() { taskENTER_CRITICAL(); }
```
Laura-Danielle, line 74: Are we purposefully trying to disable interrupts as well here?

Laura-Danielle (approved):
Looks good, just had one question about the freertos lock.

include/common/core/freertos_synchronization.hpp
amitlissack: In this case, yes. The socketcan transport classes uses a blocking read and needs to not have it be interrupted by FreeRTOS.

I don't know of a use case for this class on the firmware.

PR #140, opened 2021-11-11
feat(gantry): add limit switch & LED indicator configuration

ahiuchingau:
This PR bringup the limit switch and LED indicator for the gantry X & Y axis boards.


gantry/firmware/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ set(GANTRY_FW_NON_LINTABLE_SRCS
         ${CMAKE_CURRENT_SOURCE_DIR}/system_stm32g4xx.c
         ${CMAKE_CURRENT_SOURCE_DIR}/stm32g4xx_it.c
         ${CMAKE_CURRENT_SOURCE_DIR}/clocking.c
+        ${CMAKE_CURRENT_SOURCE_DIR}/utility_config.c
```
Laura-Danielle, line 27: Can we call this `utility_gpios` instead? Config is a bit confusing.

Laura-Danielle (changes requested):
Looks good to me aside from the naming stuff.

gantry/firmware/main.cpp
```diff
@@ -10,10 +10,12 @@
 
 #include "common/firmware/can_task.hpp"
 #include "common/firmware/clocking.h"
+#include "common/firmware/utility_gpio.h"
 
 auto main() -> int {
     HardwareInit();
     RCC_Peripheral_Clock_Select();
+    utility_init();
```
Laura-Danielle, line 18: can we also rename this to `utility_gpio_init()`

Laura-Danielle (approved):
one last comment, looks good otherwise

PR #142, opened 2021-11-12
refactor(gantry): add motor driver clock pin config

ahiuchingau:
The clock pins are different for gantry X and Y and will be configured based on node ID.


gantry/firmware/spi.c
```diff
@@ -120,4 +122,23 @@ void Reset_CS_Pin() { HAL_GPIO_WritePin(GPIOB, GPIO_PIN_12, GPIO_PIN_RESET); }
 void hal_transmit_receive(uint8_t* transmit, uint8_t* receive,
                           uint16_t buff_size, uint32_t timeout) {
     HAL_SPI_TransmitReceive(&handle, transmit, receive, buff_size, timeout);
+}
+
+void Gantry_Driver_CLK_init(uint8_t gantry_axis) {
+    switch (gantry_axis) {
+        case 0x30:
```
sfoster1, line 129: Rather than doing this, and having the exact value of the node id be what matters, can we have a symbolic value that we check? like define `GANTRY_X` or `GANTRY_Y` or something and use that to define this and the node id at the same time? Because now, if we ever change the node id value, this will silently break at runtime.

Laura-Danielle (approved):
lgtm if it works.

sfoster1 (changes requested):
This will silently break at runtime if we ever change the node id. We should make this code look at a more limited value, like some new define or config value or something, not the literal value of the canbus address.

gantry/firmware/can_task.cpp
```diff
@@ -38,6 +40,13 @@ extern FDCAN_HandleTypeDef fdcan1;
 static auto can_bus_1 = HalCanBus(&fdcan1);
 static auto message_writer_1 = MessageWriter(can_bus_1);
 
+static std::map<GantryAxisType, NodeId> AxisToNodeIdMap{
```
sfoster1, line 43: this is probably a bad idea since it allocates on the heap, sadly enough. it can probably be this instead though:

```
static constexpr NodeId node_from_axis(GantryAxisType which) {
   switch (which) {
      case GantryAxisType::gantry_x: return NodeId::gantry_x;
      case GantryAxisType::gantry_y: return NodeId::gantry_y;
   }
}
```

the compiler will yell at you if you leave off a case because it's an enum, and it probably gets fully optimized out, and then you can call it inline below.

gantry/firmware/spi.c
```diff
@@ -120,4 +123,23 @@ void Reset_CS_Pin() { HAL_GPIO_WritePin(GPIOB, GPIO_PIN_12, GPIO_PIN_RESET); }
 void hal_transmit_receive(uint8_t* transmit, uint8_t* receive,
                           uint16_t buff_size, uint32_t timeout) {
     HAL_SPI_TransmitReceive(&handle, transmit, receive, buff_size, timeout);
+}
+
+void Gantry_Driver_CLK_init(enum GantryAxisType gantry_axis) {
```
sfoster1, line 128: [nit] can it be `gantry_driver_CLK_init` instead

sfoster1 (changes requested):
sorry, tiny little extra change

sfoster1 (approved):
Nice!

sfoster1 (approved):
Let's uh really coordinate this, #143 , and #145 because I think it's gonna be really confusing having a couple different `spi_comms.cpp` floating around the tree here

PR #143, opened 2021-11-13
feat(pipettes): add a function to handle the motor driver clock pin in pipettes

Laura-Danielle:
## Overview
For now, we should set the clock pin of the motor driver to low in the pipettes subproject. Dependent on #133 and #142 being merged.


sfoster1 (approved):
Good temporary merge but we should have a bit of a refactor fest after merging #145

PR #145, opened 2021-11-15
Spi comms refactor

sotaoverride:
## Spi comms refactor
Trying break up the SPI refactor, this PR aims to make: 
- One spi_comms and spi.c/h for all projects. 
- spi_comms to get hal SPI handle from project specific can_task.cpp.


common/firmware/spi/spi.c
```diff
@@ -117,12 +117,20 @@ void SPI2_init() {
     handle = MX_SPI2_Init();
 }
 
+void get_SPI_handle(void* tmp){
+    SPI_HandleTypeDef handle = MX_SPI2_Init();
```
amitlissack, line 121: Can this function return a `SPI_HandleTypeDef *` that is globally allocated in this file?

common/firmware/spi/spi_comms.cpp
```diff
@@ -25,12 +26,24 @@ using namespace spi;
  * Public Functions
  */
 
-Spi::Spi() { SPI2_init(); }
+Spi::Spi() {
+    // get SPI handle from spi.c
+    void* tmp;
```
amitlissack, line 31: Same as above. I don't get why we would use `void*`.

amitlissack (commented):
I believe that this is more complicated than it needs to be.

common/firmware/spi/spi.c
```diff
@@ -117,12 +117,20 @@ void SPI2_init() {
...
+}
+
 void Set_CS_Pin() { HAL_GPIO_WritePin(GPIOB, GPIO_PIN_12, GPIO_PIN_SET); }
 
 void Reset_CS_Pin() {
     HAL_GPIO_WritePin(GPIOB, GPIO_PIN_12, GPIO_PIN_RESET);
 }
 
-void hal_transmit_receive(uint8_t* transmit, uint8_t* receive, uint16_t buff_size, uint32_t timeout) {
+void hal_transmit_receive(uint8_t* transmit, uint8_t* receive,
```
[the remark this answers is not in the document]
sotaoverride: think I cant use types from BSP (SPI_HandleTypeDef) in spi.h (where the signatures for these functions are at hence the void*)

common/firmware/spi/spi.c
```diff
@@ -89,7 +89,7 @@ void HAL_SPI_MspDeInit(SPI_HandleTypeDef* hspi) {
 SPI_HandleTypeDef MX_SPI2_Init() {
     /* SPI2 parameter configuration*/
     __HAL_RCC_SPI2_CLK_ENABLE();
-    SPI_HandleTypeDef hspi2 = {
+    static SPI_HandleTypeDef hspi2 = {
```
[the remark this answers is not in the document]
sotaoverride: [2e47656](https://github.com/Opentrons/ot3-firmware/pull/145/commits/2e476562682799348290143c475adbb72ead3dc9)

common/firmware/spi/spi.c
sotaoverride: [2e47656](https://github.com/Opentrons/ot3-firmware/pull/145/commits/2e476562682799348290143c475adbb72ead3dc9) makes spi.c and spi_comms.cpp less complicated.

common/firmware/spi/spi_comms.cpp
sotaoverride: [2e47656](https://github.com/Opentrons/ot3-firmware/pull/145/commits/2e476562682799348290143c475adbb72ead3dc9) makes spi.c and spi_comms.cpp less complicated.

common/firmware/spi/spi.c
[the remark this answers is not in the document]
amitlissack: I don't understand. Why can't you? We use HAL CAN types in header files.

gantry/firmware/spi.c
```diff
@@ -111,13 +111,21 @@ SPI_HandleTypeDef MX_SPI2_Init() {
     return hspi2;
 }
 
+void get_SPI_handle(void* tmp){
```
sfoster1, line 114: Hold on. I don't think we need this file anymore. At this point it duplicates what happens in `common/firmware/spi/spi.c`, and we don't need the duplication to capture the right handle anymore because it's now passed in to `hal_transmit_receive`. I think this file can go away.

common/firmware/spi/spi_comms.cpp
```diff
@@ -25,12 +26,16 @@ using namespace spi;
...
 void Spi::transmit_receive(const BufferType& transmit, BufferType& receive) {
-    Reset_CS_Pin();
+    Reset_CS_Pin(SPI_intf.GPIO_handle, SPI_intf.pin);
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
     hal_transmit_receive(const_cast<uint8_t*>(transmit.data()), receive.data(),
-                         BUFFER_SIZE, TIMEOUT);
-    Set_CS_Pin();
+                         BUFFER_SIZE, TIMEOUT,
+                         (SPI_HandleTypeDef*)this->SPI_intf.SPI_handle);
+    /*
```
sfoster1, line 37: can we remove this kind of comment? it just restates what happens in the line of code.

include/common/firmware/spi_comms.hpp
```diff
@@ -2,15 +2,34 @@
 #include <cstdint>
 #include <span>
 
+#include "platform_specific_hal_conf.h"
+
+
 namespace spi {
+struct SPI_interface {
+    /*
+      Incase there are multiple handles
```
sfoster1, line 11: This isn't true anymore, right. The idea is each `Spi` object will have precisely one hardware handle. So we don't need this comment. We can also make the type of `SPI_handle` the actual handle typedef, since we're including the hal_conf.h file above.

gantry/firmware/spi.c
sfoster1: And then add the same code to gantry `can_task` as was added in the pipette and head subprojects

sfoster1 (changes requested):
Almost there I think - just gotta do the same thing in the `gantry` executable as well as the other two.

How are we doing on compiler warnings? The warnings from including hal headers in c++ files look like this:

```
/Users/sethfoster/dev/ot3-firmware/stm32-tools/stm32g4xx-bsp/Drivers/CMSIS/Device/ST/STM32G4xx/Include/stm32g4xx.h:165:38
 warning: compound assignment with 'volatile'-qualified left operand is deprecated [-Wvolatile]
  165 | #define SET_BIT(REG, BIT)     ((REG) |= (BIT))
```
because that's not legal in cpp. What we should probably do is that includes of those files should be wrapped in
```
#pragma gcc diagnostic push
#pragma gcc diagnostic ignored "-Wvolatile"
#pragma gcc diagnostic pop
```

While we could do this in `platform_specific_hal_conf.h` internally, inside the `#ifdef __c++` or whatever block, I think that's bad because this warning is good. We should in fact have to have it in every c++ header that includes hal to make it apparent what's happening.

common/firmware/spi/spi.c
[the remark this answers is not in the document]
sotaoverride: using SPI_HandleTypeDef* now

sfoster1 (changes requested):
One final change but I think this is almost done.

In this PR, the spi init functions are called at static-constructor time. For instance, lines 241 and 242 of `head/firmware/can_task.cpp`. This happens [before main is called](https://en.cppreference.com/w/cpp/language/initialization). That may not be a good thing, since other hardware initialization is carried out either in `main`'s call chain or shortly before `main` based on the startup assembler, so it's possible that for instance clocking details of the spi hardware aren't set up properly, since they rely on core clock configuration that becomes invalid sometime in `main`'s call chain, _after_ the nonlocal constructors are called that depend on them.

The fix to this is to call `MX_SPI2_Init()` and friends from inside an actual funciton that's eventually called by `main`, such as the task setup functions, in the same way that `MX_FDCAN1_Init` is called inside `head/firmware/can_task.cpp:task_entry`. That in turn means that we probably want to
- store the hardware handle in the `Spi` class as a pointer
- create the handles at file scope _without_ actually calling into the hal initialize, so either create the structs directly at the file level or have a separate funciton that just creates and returns the struct without calling `HAL_SPI_Init`, then pass them with `&` to `Spi`
- finally, in the task entry points, make sure to call `HAL_SPI_Init` somewhere before handing control over to the task itself

gantry/firmware/can_task.cpp
```diff
@@ -43,7 +44,106 @@ static freertos_message_queue::FreeRTOSMessageQueue<Move> motor_queue(
...
+             .BaudRatePrescaler = SPI_BAUDRATEPRESCALER_32,
+             .FirstBit = SPI_FIRSTBIT_MSB,
+             .TIMode = SPI_TIMODE_DISABLE,
+             .CRCCalculation = SPI_CRCCALCULATION_DISABLE,
+             .CRCPolynomial = 7,
+             .CRCLength = SPI_CRC_LENGTH_DATASIZE,
+             .NSSPMode = SPI_NSS_PULSE_DISABLE}
+
+};
+SPI_HandleTypeDef MX_SPI2_Init() {
```
sfoster1, line 114: this doesn't appear to be called, we should delete

pipettes/firmware/can_task.cpp
```diff
@@ -47,7 +48,100 @@ static freertos_message_queue::FreeRTOSMessageQueue<Move> motor_queue(
...
+             .BaudRatePrescaler = SPI_BAUDRATEPRESCALER_32,
+             .FirstBit = SPI_FIRSTBIT_MSB,
+             .TIMode = SPI_TIMODE_DISABLE,
+             .CRCCalculation = SPI_CRCCALCULATION_DISABLE,
+             .CRCPolynomial = 7,
+             .CRCLength = SPI_CRC_LENGTH_DATASIZE,
+             .NSSPMode = SPI_NSS_PULSE_DISABLE}
+
+};
+SPI_HandleTypeDef MX_SPI2_Init() {
```
sfoster1, line 112: doesn't appear to be called, should delete

sfoster1 (changes requested):
Just delete that last unused code and I think we're good!

sfoster1 (approved):
Now tested and (after pushed fixes)
- functions on gantry
- functions on head; can drive both motors (only one at a time though)
looks good!

PR #146, opened 2021-11-15
feat(motion-controller): set and get motion constraints

ahiuchingau:
This PR does the following:
1. allows the ability to set and retrieve min&max velocity and acceleration for each subsystem via CAN
2. sets the initial values for min&max velocity and acceleration for each subsystem (values are arbitrary and should be modified in the future)
3. remove GetSpeedRequest&Response

Note: The firmware will only store, update and send over the motion constraints via CAN and not use them to evaluate or reject any moves. That should be handled on the Python side instead. 

closes RET-41 and closes RET-78

Corresponding changes on the Python side: https://github.com/Opentrons/opentrons/pull/8772

### DIFF
--- a/include/can/core/message_handlers/motor.hpp
+++ b/include/can/core/message_handlers/motor.hpp
@@ -20,8 +20,8 @@ class MotorHandler {
                      DisableMotorRequest, GetMotionConstraintsRequest,
                      SetMotionConstraints>;
 
-    MotorHandler(MessageWriter &message_writer, Motor &motor)
-        : message_writer{message_writer}, motor{motor} {}
+    MotorHandler(MessageWriter &message_writer, Motor &motor, NodeId node_id)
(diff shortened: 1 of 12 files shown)


include/can/core/message_handlers/motor.hpp
```diff
@@ -54,6 +57,21 @@ class MotorHandler {
         motor.motion_controller.disable_motor();
     }
 
+    void visit(GetMotionConstraintsRequest &m) {
+        auto constraints = motor.motion_controller.get_motion_constraints();
+        GetMotionConstraintsResponse response_msg{
```
amitlissack, line 62: See my note in the companion python PR about NodeId being required.

amitlissack (changes requested):

amitlissack (approved):

PR #147, opened 2021-11-16
refactor(pipettes): use the STM32L562RET as the MCU for the pipette form factor boards

Laura-Danielle:
## Overview

The nucleo boards given were the L552 whereas the pipette form factor board uses the L562. This PR
changes the naming of the start-up files as well as utilizes the startup and flash files of the L562
rather than the L552.


common/firmware/STM32L562QEIx/STM32L562QEIx_FLASH.ld
```diff
@@ -1,13 +1,11 @@
...
+
 **  File        : LinkerScript.ld
 **
-**  Author		: Auto-generated by STM32CubeIDE
-**
-**  Abstract    : Linker script for STM32L5x2xE Device from STM32L5 series
-**                      512Kbytes ROM
-**                      192Kbytes RAM
+**  Abstract    : Linker script for STM32L562QEIx Device with
+**                511KByte FLASH, 256KByte RAM
```
sfoster1, line 8: 511?

common/firmware/STM32L562QEIx/STM32L562QEIx_FLASH.ld
```diff
@@ -37,33 +33,34 @@
...
-/* Sections */
+/* Define output sections */
 SECTIONS
 {
-  /* The startup code into "ROM" Rom type memory */
+  /* The startup code goes first into FLASH */
   .isr_vector :
   {
-    . = ALIGN(8);
+    . = ALIGN(4);
```
sfoster1, line 55: worrying, although i think it'll be ok because this should be at 0

common/firmware/STM32L562QEIx/STM32L562QEIx_FLASH.ld
```diff
@@ -37,33 +33,34 @@
...
-  } >ROM
+    . = ALIGN(4);
+  } >FLASH
 
-  /* The program code and other data into "ROM" Rom type memory */
+  /* The program code and other data goes into FLASH */
   .text :
   {
-    . = ALIGN(8);
+    . = ALIGN(4);
```
sfoster1, line 63: actually do you know why this characteristically aligns on 4-byte boundaries instead of 8?

sfoster1 (approved):
Looks good to me if it compiles and runs although I'm a bit 🤔 about the alignment

common/firmware/STM32L562QEIx/STM32L562QEIx_FLASH.ld
Laura-Danielle: I got this from the ST's github, but I'll put a note here in case things still don't work after clock adjustments.

common/firmware/STM32L562QEIx/STM32L562QEIx_FLASH.ld
Laura-Danielle: Typo on their part. I changed it.

sfoster1 (approved):
do it

PR #150, opened 2021-11-17
refactor(can): add Node ID to all CAN responses

ahiuchingau:
As a followup on https://github.com/Opentrons/opentrons/pull/8772, this PR adds a NodeId to all of the CAN responses from firmware.

### DIFF
--- a/include/can/core/message_writer.hpp
+++ b/include/can/core/message_writer.hpp
@@ -6,13 +6,16 @@
 #include "can_bus.hpp"
 #include "common/core/freertos_synchronization.hpp"
 #include "common/core/synchronization.hpp"
+#include "ids.hpp"
 #include "message_core.hpp"
 
 namespace can_message_writer {
 
 class MessageWriter {
   public:
-    explicit MessageWriter(can_bus::CanBusWriter& writer) : writer{writer} {}
+    explicit MessageWriter(can_bus::CanBusWriter& writer,
+                           can_ids::NodeId node_id)
+        : writer{writer}, node_id(node_id) {}
 
     /**
      * Write a message to the can bus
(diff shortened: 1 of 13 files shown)


include/can/core/messages.hpp
```diff
@@ -48,13 +48,26 @@ struct Empty : BaseMessage<MId> {
...
 };
 
+template <MessageId MId>
+struct Response : BaseMessage<MId> {
+    NodeId node_id;
+
+    void set_node_id(NodeId id) { node_id = id; }
+
+    template <bit_utils::ByteIterator Output, typename Limit>
+    auto serialize_node_id(Output body, Limit limit) const {
```
amitlissack, line 58: I figured you made this function in order to return an iterator rather than only a size.  But I think this class should still implement a `serialize` method. 

Just for completeness sake.

sfoster1 (approved):
Looks good as long as we can accommodate one microprocessor having multiple ids somehow (head probably wants one for z and one for a, for instance)

include/can/core/message_writer.hpp
```diff
@@ -21,14 +24,15 @@ class MessageWriter {
      * @param node The node id
      * @param message The message to send
      */
-    template <message_core::Serializable Serializable>
-    void write(can_ids::NodeId node, const Serializable& message) {
+    template <message_core::CanResponseMessage ResponseMessage>
```
amitlissack, line 27: Sort of a nitpick.... but if all response messages have the node_id then I feel that it should be the first item in the buffer.

amitlissack (commented):
Really nice job. 

Just a few little things.  Not 100% necessary to make changes.

amitlissack (approved):
Very nice solution.

PR #152, opened 2021-11-18
refactor(can): remove NodeID from CAN message handlers

ahiuchingau:
Now that we have the CAN message writer to handle adding the Node ID to the responses, we no longer need to pass the Node ID to the CAN message handlers.


sfoster1 (approved):
noice

Laura-Danielle (approved):
🆒

PR #153, opened 2021-11-18
refactor(pipettes): modify sys clock configurations

Laura-Danielle:
## Overview

The sys clock is slightly different for the L5 as compared to the G4 and thus configurations need to
change. This, in turn, means configurations for both CAN and the motor timer interrupt need to be modified as well.

I will merge both #147 and this PR in once they have been verified on hardware.


pipettes/firmware/system_stm32l5xx.c
```diff
@@ -353,26 +353,64 @@ void SystemCoreClockUpdate(void)
...
+    RCC_ClkInitStruct.SYSCLKSource = RCC_SYSCLKSOURCE_PLLCLK;
+    RCC_ClkInitStruct.AHBCLKDivider = RCC_SYSCLK_DIV1;
+    RCC_ClkInitStruct.APB1CLKDivider = RCC_HCLK_DIV1;
+    RCC_ClkInitStruct.APB2CLKDivider = RCC_HCLK_DIV1;
+
+    if (HAL_RCC_ClockConfig(&RCC_ClkInitStruct, FLASH_LATENCY_5) != HAL_OK)
+    {
+        Error_Handler();
+    }
+//    RCC_ClkInitTypeDef RCC_ClkInitStruct = {0};
```
sfoster1, line 394: Cast this commented code from mine sight

sfoster1 (approved):

PR #154, opened 2021-11-18
feat(gantry): Add motor driver clock pin based on gantry axis

ahiuchingau:
Same changes as in this older [PR](https://github.com/Opentrons/ot3-firmware/pull/142). It was closed because there were too many new changes to rebase on main.


Laura-Danielle (approved):
lgtm

PR #156, opened 2021-11-18
feat(pipettes): add a function to handle the motor driver clock pin in pipettes

Laura-Danielle:
## Overview

For now, we should set the clock pin of the motor driver to low in the pipettes subproject.


sfoster1 (approved):
noice

PR #157, opened 2021-11-19
refactor(motion-control): allow different driver configurations for subsystems

ahiuchingau:
Driver setup configurations are now in each subsystem's `can_task.cpp`, probably should do a followup refactor PR to pull all subsystem config out of the task file.

### DIFF
--- a/include/motor-control/core/motor_driver_config.hpp
+++ b/include/motor-control/core/motor_driver_config.hpp
@@ -1,7 +1,5 @@
 #pragma once
 
-#include <map>
-
 namespace motor_driver_config {
 
 enum class DriverRegisters : uint8_t {
--- a/include/motor-control/core/motor_driver_config.hpp
+++ b/include/motor-control/core/motor_driver_config.hpp
@@ -23,6 +23,22 @@ enum class DriverRegisters : uint8_t {
     ENCM_CTRL = 0x72
 };
 
-typedef std::map<DriverRegisters, uint32_t> RegisterConfig;
+struct RegisterConfig {
+    uint32_t gconf = 0;
+    uint32_t gstat = 0;
+    uint32_t ioin = 0;
+    uint32_t ihold_irun = 0;
+    uint32_t tpowerdown = 0;
+    uint32_t tpwnthrs = 0;
+    uint32_t tcoolthrs = 0;
+    uint32_t chopconf = 0;
+    uint32_t thigh = 0;
+    uint32_t coolconf = 0;
+    uint32_t xdirect = 0;
+    uint32_t vdcmin = 0;
+    uint32_t dcctrl = 0;
+    uint32_t pwmconf = 0;
+    uint32_t encm_ctrl = 0;
+};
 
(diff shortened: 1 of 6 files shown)


include/motor-control/core/motor_driver_config.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <map>
```
sfoster1, line 3: Sadly this uses `malloc`/`new` under the hood and we do want to avoid that. it's also true that what we're looking for doesn't necessarily fit the concept of a map, a general-purpose associative store; we have a finite number of default-valued options. IMO that's a better fit for a aggregate-initializable struct type anyway, e.g.

```
struct RegisterConfig {
    uint32_t gconf = 0;
    uint32_t gstat = ...
    ....
    uint32_t encm_ctrl = 0;
};
```
which can then be initialized in the can tasks like
```
RegisterConfig{
   .gconf = 0x7002,
   .thigh = ....
   // omit what you don't want to configure and it will take the default value
}```

**Review by sfoster1 — CHANGES_REQUESTED** [review]
sorry but we should really avoid using `map` unfortunately. i think having a struct type as noted there would work but if you want to keep it dynamic, i think it would also be cool for that type to be
```
template<size_t Size> using RegistorConfig = std::array<std::pair<DriverRegisters, uint32_t>, Size>``` 
i think that's how you spell it, but you leave the `Size` as a template param. you then would have to have `MotorDriver` also parametrized either by the `RegisterConfig` type or by the size... eh I kind of like the big `struct` better.

Also just sayin while you're at it you could start calling `setup()` in `can_task::task_entry`

sfoster1:
> sorry but we should really avoid using `map` unfortunately. i think having a struct type as noted there would work but if you want to keep it dynamic, i think it would also be cool for that type to be
> 
> ```
> template<size_t Size> using RegistorConfig = std::array<std::pair<DriverRegisters, uint32_t>, Size>``` 
> i think that's how you spell it, but you leave the `Size` as a template param. you then would have to have `MotorDriver` also parametrized either by the `RegisterConfig` type or by the size... eh I kind of like the big `struct` better.
> 
> Also just sayin while you're at it you could start calling `setup()` in `can_task::task_entry`
> ```

Oh also if you wanted to go on a conceptual voyage through the wonderscape of the mind and also stick with `std::map` you could write a [custom allocator](https://en.cppreference.com/w/cpp/memory/allocator) for std::map as with [this](https://stackoverflow.com/questions/11373796/custom-memory-allocator-for-stl-map) that static allocates enough memory for an entry for each register and then gives the `std::map` blocks of that

include/motor-control/core/motor_driver_config.hpp
sfoster1: oh we should get rid of this since it's not used anymore

sfoster1 (approved):
Looks awesome, love to see it, let's remove that `#include <map>` and rock and roll. never have to send the canbus setup message again unless we change params!

PR #158, opened 2021-11-19
feat(motion-control): add CAN messages to read/write to motor driver registers

ahiuchingau:
This PR adds CAN messages to allow us to read and write to motor driver registers via SPI. Mainly for debug purposes.

I've created a pretty terribly useless concept `RegisterAddress` in case we want to touch any undefined motor register addresses via CAN. Should I be doing this? Let me know your thoughts.

### DIFF
--- a/include/can/core/message_handlers/motor.hpp
+++ b/include/can/core/message_handlers/motor.hpp
@@ -4,12 +4,14 @@
 #include "can/core/message_writer.hpp"
 #include "can/core/messages.hpp"
 #include "common/core/message_queue.hpp"
+#include "motor-control/core/motor_driver_config.hpp"
 #include "motor-control/core/motor_messages.hpp"
 
 namespace motor_message_handler {
 
 using namespace can_message_writer;
 using namespace can_messages;
+using namespace motor_driver_config;
 
 template <class Motor>
 class MotorHandler {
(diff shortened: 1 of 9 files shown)


include/can/core/message_handlers/motor.hpp
```diff
@@ -72,6 +73,20 @@ class MotorHandler {
         motor.motion_controller.set_motion_constraints(m);
     }
 
+    void visit(WriteMotorDriverRegister &m) {
+        motor.driver.write(m.reg_address, m.data);
```
sfoster1, line 77: Let's check that this is a valid address first

include/can/core/message_handlers/motor.hpp
```diff
@@ -72,6 +73,20 @@ class MotorHandler {
...
+    }
+
+    void visit(ReadMotorDriverRegister &m) {
+        uint32_t data;
+        motor.driver.read(m.reg_address, data);
+        ReadMotorDriverRegisterResponse response_msg{
```
sfoster1, line 83: check the address

sfoster1 (approved):
Looks good but we should really consider this a debug element

sfoster1 (approved):
Perfect

PR #159, opened 2021-11-22
refactor(can): remove HAL references from CAN cpp files.

amitlissack:
- HAL is not included in C++ CAN files. A C layer is introduced between C++ and HAL.
- Use callback for new can messages rather than a message buffer owned by CAN.

(i need to do a little more clean up before merging)


gantry/firmware/can_task.cpp
```diff
@@ -21,11 +21,11 @@
 #pragma GCC diagnostic push
 // NOLINTNEXTLINE(clang-diagnostic-unknown-warning-option)
 #pragma GCC diagnostic ignored "-Wvolatile"
-#include "can/firmware/hal_can_bus.hpp"
-#include "can/firmware/hal_can_message_buffer.hpp"
-#include "common/firmware/can.h"
+//#include "can/firmware/hal_can_bus.hpp"
```
sfoster1, line 24: can remove these outright now right

head/firmware/can_task.cpp
```diff
@@ -20,11 +19,11 @@
 #pragma GCC diagnostic push
 // NOLINTNEXTLINE(clang-diagnostic-unknown-warning-option)
 #pragma GCC diagnostic ignored "-Wvolatile"
-#include "can/firmware/hal_can_bus.hpp"
-#include "can/firmware/hal_can_message_buffer.hpp"
-#include "common/firmware/can.h"
+//#include "can/firmware/hal_can_bus.hpp"
```
sfoster1, line 22: can remove

pipettes/firmware/can_task.cpp
```diff
@@ -24,10 +24,10 @@
 #pragma GCC diagnostic push
 // NOLINTNEXTLINE(clang-diagnostic-unknown-warning-option)
 #pragma GCC diagnostic ignored "-Wvolatile"
-#include "can/firmware/hal_can_message_buffer.hpp"
-#include "common/firmware/can.h"
+//#include "can/firmware/hal_can_message_buffer.hpp"
```
sfoster1, line 27: can remove

sfoster1 (approved):
This is awesome! We can remove some of those commented includes but that's the only change we need I think

sfoster1 (approved):
lgtm!

PR #160, opened 2021-11-22
chore(gantry): separate gantry targets

sfoster1:
Having the gantry type defined as a combination of cmake variable and
automatically generated file is good, but the downside is that it
requires a cmake reconfigure in between switching gantry types, which is
kind of annoying when it's a frequent task to want to build and use both
gantry types in sequence. Another way to do it is just to have two side
by side targets. They share all the same code except for the gantry type
cpp files, and each target uses a different one.


amitlissack (approved):

ahiuchingau (approved):
Sweet!

PR #161, opened 2021-11-23
Head motor handlers

sotaoverride:
Add dispatch handlers for a and z motors. Make a and z motors move with the same move group and task. Follow up PR to add unique IDs to both motors, and dedicated tasks.

### DIFF
--- a/head/firmware/can_task.cpp
+++ b/head/firmware/can_task.cpp
@@ -72,7 +72,7 @@ spi::SPI_interface SPI_intf3 = {
 };
 static spi::Spi spi_comms3(SPI_intf3);
 
-struct motion_controller::HardwareConfig PinConfigurations {
+struct motion_controller::HardwareConfig pin_configurations_left {
     .direction = {.port = GPIOC,
                   .pin = GPIO_PIN_1,
                   .active_setting = GPIO_PIN_SET},
@@ -81,7 +81,7 @@ struct motion_controller::HardwareConfig PinConfigurations {
         .port = GPIOC, .pin = GPIO_PIN_4, .active_setting = GPIO_PIN_SET},
 };
 
-struct motion_controller::HardwareConfig PinConfigurations2 {
+struct motion_controller::HardwareConfig pin_configurations_right {
     .direction = {.port = GPIOC,
                   .pin = GPIO_PIN_7,
                   .active_setting = GPIO_PIN_SET},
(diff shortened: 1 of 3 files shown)


head/firmware/can_task.cpp
```diff
@@ -81,6 +81,15 @@ struct motion_controller::HardwareConfig PinConfigurations {
         .port = GPIOC, .pin = GPIO_PIN_4, .active_setting = GPIO_PIN_SET},
 };
 
+struct motion_controller::HardwareConfig PinConfigurations2 {
```
sfoster1, line 84: can we give these better names? `pin_configurations_left` and `pin_configurations_right` maybe?

head/firmware/can_task.cpp
```diff
@@ -109,8 +118,26 @@ static motor_class::Motor motor{
     motor_queue,
     complete_queue};
 
+/*z motor*/
+static motor_class::Motor motor2{
```
sfoster1, line 122: let's just name this `motor_left` or something, and so on

head/firmware/motor_hardware.c
```diff
@@ -4,9 +4,8 @@ void HAL_SPI_MspInit(SPI_HandleTypeDef* hspi) {
     GPIO_InitTypeDef GPIO_InitStruct = {0};
     if (hspi->Instance == SPI2) {
         /* Peripheral clock enable */
-        __HAL_RCC_SPI2_CLK_ENABLE();
```
sfoster1, line 7: why remove these? where did they go instead?

sfoster1 (changes requested):
Let's fix up some names and figure out these clocks and then this looks good!

head/firmware/motor_hardware.c
sotaoverride: [here](https://github.com/Opentrons/ot3-firmware/blob/main/head/firmware/motor_hardware.c)
```
HAL_StatusTypeDef initialize_spi(SPI_HandleTypeDef* hspi) {
    if (hspi->Instance == SPI2) {
        __HAL_RCC_SPI2_CLK_ENABLE();
    } else if (hspi->Instance == SPI3) {
        __HAL_RCC_SPI3_CLK_ENABLE();
    } else {
        return HAL_ERROR;
    }
    return HAL_SPI_Init(hspi);
}
```

head/firmware/can_task.cpp
```diff
@@ -135,19 +135,19 @@ static motor_class::Motor motor2{
     complete_queue};
 
 /** The parsed message handler */
-static auto can_motor_handler = MotorHandler{message_writer_1, motor};
-static auto can_motor_handler2 = MotorHandler{message_writer_1, motor2};
+static auto can_motor_handler = MotorHandler{message_writer_1, motor_right};
```
amitlissack, line 138: Can we use `right` and `left` instead of `2`?

amitlissack (approved):

sfoster1 (approved):

PR #165, opened 2021-11-24
feat(simulation): build out gantry simulator

amitlissack:
This PR fleshes out the gantry simulator applications.

The main effort was to move `can_task.cpp` out of `firmware` and into `core`. To do this I added `interfaces.hpp` that provides access to abstract versions of hardware interfaces implemented differently in fw and simulation. These are CAN, SPI, and Motor hardware interface.

To enable simulation on a non-Linux machine, I added a CAN transport over plain old sockets. A companion PR for `opentrons-hardware` is in the works.

**This is ready for review, but cannot be merged until I can test on hardware.**


include/common/core/logging.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
```
amitlissack, line 2: Why this approach over BOOST? Because it's simple.

Also, I didn't see how to conditionally compile BOOST logging.

gantry/simulator/CMakeLists.txt
```diff
@@ -1,18 +1,67 @@
+# Load free for the posix port
+find_package(FreeRTOS)
+
 # Set up the include directories
 include_directories(
         .
+        ../firmware
```
sfoster1, line 7: Why do we need this? Ideally we can keep all of firmware completely separate from all of simulator. We should consider refactoring until we don't need this.

gantry/simulator/CMakeLists.txt
```diff
@@ -1,18 +1,67 @@
...
+list(APPEND FREERTOS_SOURCES "${FreeRTOS_SOURCE_DIR}/FreeRTOS/Source/portable/MemMang/heap_3.c")
+list(APPEND FREERTOS_SOURCES "${FreeRTOS_SOURCE_DIR}/FreeRTOS/Source/portable/ThirdParty/GCC/Posix/utils/wait_for_event.c")
+list(APPEND FREERTOS_SOURCES "${FreeRTOS_SOURCE_DIR}/FreeRTOS/Source/portable/ThirdParty/GCC/Posix/port.c")
+
+
+# Create the freertos lib
+add_library(freertos-gantry STATIC ${FREERTOS_SOURCES})
+
+if("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Linux")
+    add_compile_definitions(USE_SOCKETCAN)
```
sfoster1, line 31: We shouldn't use project-wide cmake commands. We should instead use target-specific commands (the equivalent here would be [target_compile_definitions](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html) on a target. If we need to add it to multiple targets, then we should.

gantry/simulator/CMakeLists.txt
```diff
@@ -1,18 +1,67 @@
...
+# Add the posix port files
+list(APPEND FREERTOS_SOURCES "${FreeRTOS_SOURCE_DIR}/FreeRTOS/Source/portable/MemMang/heap_3.c")
+list(APPEND FREERTOS_SOURCES "${FreeRTOS_SOURCE_DIR}/FreeRTOS/Source/portable/ThirdParty/GCC/Posix/utils/wait_for_event.c")
+list(APPEND FREERTOS_SOURCES "${FreeRTOS_SOURCE_DIR}/FreeRTOS/Source/portable/ThirdParty/GCC/Posix/port.c")
+
+
+# Create the freertos lib
+add_library(freertos-gantry STATIC ${FREERTOS_SOURCES})
+
+if("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Linux")
```
sfoster1, line 30: we probably should use [generator expressions](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#conditional-expressions) for this

gantry/simulator/CMakeLists.txt
```diff
@@ -1,18 +1,67 @@
...
+    add_compile_definitions(USE_SOCKETCAN)
+endif()
+
+add_compile_definitions(ENABLE_LOGGING)
+
+set(GANTRY_SIMULATOR_SRC
+        ${CMAKE_CURRENT_SOURCE_DIR}/main.cpp
+        ${CMAKE_CURRENT_SOURCE_DIR}/interfaces.cpp
+        ../firmware/freertos_idle_timer_task.cpp
```
sfoster1, line 39: same with the include - IMO we really shouldn't be mixing firmware and simulator code even at the expense of duplication.

include/can/simlib/socket_transport.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+    if ((s = socket(AF_INET, SOCK_STREAM, 0)) == -1) {
+        return false;
+    }
+
+    addr.sin_family = AF_INET;
+    addr.sin_port = htons(port);
+    ::inet_aton(ip, &addr.sin_addr);
+
+    if (connect(s, (struct sockaddr *)&addr, sizeof(addr)) == -1) {
```
sfoster1, line 50: direct bsd socket api... like settling into a warm bath...

include/can/simlib/socket_transport.hpp
```diff
@@ -0,0 +1,100 @@
...
+        return false;
```
sfoster1, line 51: could `throw` here

include/can/simlib/socket_transport.hpp
sfoster1: but also maybe boost sockets? don't really have a preference but it might be easier, allow `unix://` vs `tcp://` protocol specification, allow namelookup, etc

include/common/core/logging.hpp
sfoster1: you'd kind of do both, i think. in the `#ifdef ENABLE_LOGGING` section your log macro would forward to boost, otherwise blackhole it like you do today.

include/can/simlib/socket_transport.hpp
```diff
@@ -0,0 +1,100 @@
...
+    //  process message. Need to find a better solution.
+    vTaskDelay(1);
+
+    // Critical section block
+    auto lock = synchronization::Lock(critical_section);
+
+    ::read(handle, &arb_id, sizeof(arb_id));
+    ::read(handle, &buff_len, sizeof(buff_len));
+    arb_id = ntohl(arb_id);
+    buff_len = ntohl(buff_len);
```
sfoster1, line 91: for some definition of can max message size,
```suggestion
    buff_len = std::min(ntohl(buff_len), CAN_MAX_MESSAGE_SIZE);
```
? sorry, trusting network input for message size even for internal apis is making me twitch a bit

sfoster1 (changes requested):
Some inline comments but structurally, I disagree with the implementation of the hardware-interface concept, because
- Linker errors are harder to read than compiler errors
- It bundles together everything in a way that doesn't promote separation of concerns

Another approach could be inverting control and making the can task in core instantiated-by or owned-by objects, functions, or files in firmware and simulator.

gantry/simulator/CMakeLists.txt
amitlissack: For the freertos config

gantry/simulator/CMakeLists.txt
amitlissack: The amount of duplication in this codebase makes me pukey. But ok.

include/can/simlib/socket_transport.hpp
amitlissack: I would like to avoid boost until the benefits truly outweigh the pain

include/common/core/logging.hpp
amitlissack: I don't see the big benefit of boost yet.

amitlissack:
> Some inline comments but structurally, I disagree with the implementation of the hardware-interface concept, because
> 
> * Linker errors are harder to read than compiler errors
> * It bundles together everything in a way that doesn't promote separation of concerns
> 
> Another approach could be inverting control and making the can task in core instantiated-by or owned-by objects, functions, or files in firmware and simulator.

You're right. I really hate how this code is written. I'll spend more time on it.

sfoster1 (approved):
I think this is good for now, we should merge it so the simulators work. In the future, we should split some of these tasks; make dispatchers only know of destination queues; and explicitly separate the two axes of configuration:
1. between executables (pipettes vs head vs gantry)
2. between compilation targets (simulation vs firmware vs test)

PR #166, opened 2021-11-24
refactor(spi): simulated spi

amitlissack:
- Convert SPI from a concept to an abstract base class.
- Moved SPI driver related code out of `MotorDriver`
- Created a simulated version of SPI
- Unit tests.


sfoster1 (approved):
Looks good, I'm just always a little worried with this kind of thing that it locks us in to a particular way of storing this spi object since inheritance only works through reference or pointer - we now _must_ separate the allocation and storage of the spi object from the owning object, which is also the only owning object. Sticking with templates gives us more flexibility. But it is somewhat less wordy this way.

amitlissack:
> Looks good, I'm just always a little worried with this kind of thing that it locks us in to a particular way of storing this spi object since inheritance only works through reference or pointer - we now _must_ separate the allocation and storage of the spi object from the owning object, which is also the only owning object. Sticking with templates gives us more flexibility. But it is somewhat less wordy this way.

We are limiting this kind of inheritance only for objects that are created at initialization and live for the lifetime of the application. Most importantly, there's one version for sim and one for the the real stuff.

sfoster1 (approved):
Love it!

ahiuchingau (approved):
Abstract base class rules

PR #167, opened 2021-11-29
refactor(*): parametrize motor hardware iface

sfoster1:
Refactor the way the motor control system (what emits step and direction
pulses) interacts with the hardware layer.

The current implementation had no single source of truth for what pins
did what and how to interact with them, leading to a lot of duplicated
code and a tough time following the way motor commands went from coming
in on the canbus to going out on gpio lines.

Instead of using static c functions to implement setting pin state, we
can make them one implementation of a c++ pure virtual base class, which
we can give in place of the pin configurations to the motion controller
and also provide to the motor interrupt handler. This makes the full
loop from getting a motor command to calling some function that sets a
pin live in the core.

This commit creates and implements this interface for firmware and
reshuffles a bunch of code to integrate it into the three end
executables and rehome some of the implementations in a new common
subdirectory. It also removes some templatization from things like the
motor interrupt handler, which were only templatized over their queue
implementation, which is actually invariant since we're using POSIX
freertos in test and simulator.

There's a lot of churn in this pr, but most of it is from
- making `motor-control/(firmware,tests)` and `include/motor-control/(core,firmware,tests)` and moving things into them
- moving all the stuff in the individual executable `tim7.c`s into their `motor_hardware.c`s


include/motor-control/core/motion_controller.hpp
```diff
@@ -71,15 +51,15 @@ class MotionController {
         queue.try_write(msg);
     }
 
-    void enable_motor() { set_pin(hardware_config.enable); }
+    void stop() { hardware.stop_timer_interrupt(); }
```
amitlissack, line 54: When we talked in pairing session you suggested that this is not the right way to stop. Should we leave a TODO or ticket this?

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -184,27 +198,27 @@ class MotorInterruptHandler {
...
+    // test interface
     [[nodiscard]] auto get_current_position() const -> q31_31 {
         return position_tracker;
     }
-
     void set_current_position(q31_31 pos_tracker) {
         position_tracker = pos_tracker;
     }
-
+    bool has_active_move = false;
```
amitlissack, line 208: Is this supposed to be here?

motor-control/tests/test_motor_pulse.cpp
```diff
@@ -1,13 +1,23 @@
...
+#include "motor-control/tests/mock_motor_hardware.hpp"
 
 using namespace motor_handler;
 
 #define TO_RADIX 31
+
 static constexpr sq0_31 default_velocity =
     0x1 << (TO_RADIX - 1);  // half a step per tick
 
+#define BUILD_INTERRUPT_HANDLER_Q(name, queuename, cqueuename) \
```
amitlissack, line 13: Are my eyes deceiving me? Is that a macro?? Nice to see that the oldies still have their place in our code base.

amitlissack (commented):
This is a really nice improvement. 

Let's test on hardware before merging.

include/motor-control/core/motor_interrupt_handler.hpp
sfoster1: Yeah that was that testing-only member. the diff on this file is because I manually copied it back to the header

motor-control/tests/test_motor_pulse.cpp
sfoster1: A classic of the genre, really. Probably should have just made it an object but I also kind of missed doing macros and it's just one file

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -40,40 +39,85 @@ using namespace motor_messages;
...
+         *
+         * Logic:
+         * 1. If there is not currently an active move, we should check if there
+         * are any available on the queue.
+         * 2. If there is an active move, and stepping is possible, then we
+         * should increment the step counter and return true.
+         * 3. Finally, if there is an active move, but you can no longer step
+         * then active move should be set to false.
+         *
+         * This function is called from a timer interrupt. See `step_motor.cpp`.
```
Laura-Danielle, line 100:
```suggestion
         * This function is called from a timer interrupt. See `motor_hardware.cpp`.
```

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -92,8 +136,19 @@ class MotorInterruptHandler {
...
+        /*
+         * A motor should only try to take a step when the current position
+         * does not equal the target position.
+         */
+        return tick_count < buffered_move.duration;
+    }
+
     void update_move() {
-        has_active_move = queue->try_read_isr(&buffered_move);
+        has_active_move = queue.try_read_isr(&buffered_move);
```
Laura-Danielle, line 151: Why did we decide to not make the queue a pointer?

motor-control/tests/test_motor_pulse.cpp
Laura-Danielle: I was under the impression we didn't like macros

Laura-Danielle (approved):
Huge improvement. Thanks for looking into this.

include/motor-control/core/motor.hpp
```diff
@@ -2,42 +2,41 @@
...
-    using GenericQueue = PendingQueueImpl<Move>;
-    using CompletedQueue = CompletedQueueImpl<Ack>;
+    using GenericQueue = FreeRTOSMessageQueue<Move>;
+    using CompletedQueue = FreeRTOSMessageQueue<Ack>;
     Motor(SpiDriver& spi, lms::LinearMotionSystemConfig<MEConfig> lms_config,
-          motion_controller::HardwareConfig& config,
+          motor_hardware::MotorHardwareIface& hardware_iface,
           MotionConstraints constraints, RegisterConfig driver_config,
           GenericQueue& queue, CompletedQueue& completed_queue)
         : pending_move_queue(queue),
```
ahiuchingau, line 31: We can probably not make `pending_move_queue` and `completed_move_queue` members of the Motor struct

ahiuchingau (approved):

motor-control/tests/test_motor_pulse.cpp
sfoster1: I can change it. We don't like macros but the big reason we don't like them is their maintainability - and one that's scoped to precisely one test file isn't really that bad IMO.

include/motor-control/core/motor_interrupt_handler.hpp
sfoster1: Now that the queues are provided to the object's constructor, they will always be present and never be null. That makes references (the c++ language feature) the better way to represent them - when you're choosing a reference (the general programming language concept) type in c++ and you're not affected by allocation details, the choice of whether to use a c++ ref or a pointer basically boils down to "do I have to handle this not being present? pointer. will it always be present? reference."

include/motor-control/core/motor.hpp
sfoster1: Actually those queue refs are used by the move group executor to get a way to talk to the motor interrupt handler - the `Motor` class is a big bag of references that link the motor system together.

motor-control/tests/test_motor_pulse.cpp
amitlissack: Sorry for starting this. I wasn't really criticizing. Just thought it was funny to see a macro.

motor-control/firmware/motor_control_hardware.c
```diff
@@ -0,0 +1,24 @@
...
+#include "platform_specific_hal_conf.h"
+
+void motor_hardware_set_pin(void* port, uint16_t pin, uint8_t active_setting) {
+    HAL_GPIO_WritePin(port, pin, active_setting);
+}
+
+void motor_hardware_reset_pin(void* port, uint16_t pin, uint8_t active_setting) {
+    HAL_GPIO_WritePin(
+        port, pin,
+        (active_setting == GPIO_PIN_SET) ? GPIO_PIN_RESET : GPIO_PIN_RESET);
```
amitlissack, line 19: What is the purpose of this statement? Is this supposed to support active low pins? This looks wrong.

sfoster1: yeah the idea is to support active-low pins. This is "turning the pin off" (i.e. set to low if active high, high if active low)

sfoster1: could probably be a bit easier to read

amitlissack (approved):

PR #168, opened 2021-11-30
Node dispatcher

sotaoverride:
dispatcher for subsystems with multiple motors (head).


amitlissack:
We chatted about change suggestions in person.

include/can/core/dispatch.hpp
```diff
@@ -81,6 +81,46 @@ requires HandlesMessages<HandlerType, MessageTypes...> &&
...
+                limit);
+            handler.handle(result);
+        }
+    }
+
+  private:
+    HandlerType& handler;
+    Parser<MessageTypes...> parser;
+    uint16_t node;
+    auto check_motor(uint32_t arbitration_id) {
```
sfoster1, line 115: a bit of a nitpick, but can we a) call this `check_node` or something since it could be used for something other than a motor; make the method `const`; and annotate the return type, even though it's not technically necessary
```suggestion
    auto check_node(uint32_t arbitration_id) -> bool const {
```

include/can/core/ids.hpp
```diff
@@ -27,7 +27,9 @@ enum class NodeId : uint8_t {
     pipette = 0x20,
     gantry_x = 0x30,
     gantry_y = 0x40,
-    head = 0x50,
+    head_left = 0x50,
```
sfoster1, line 30: also a nitpick but if we swap these we can keep it (somewhat) backwards compatible. And I think it might be better to use the low nibble here - we're reserving all this space in the node id, might as well use it (I think we are, anyway - is it valid to have that low nibble be anything other than 0?)
```suggestion
    head = 0x50,
    head_right = 0x51,
    head_left = 0x52
```

head/firmware/can_task.cpp
```diff
@@ -171,39 +171,40 @@ static auto can_move_group_executor_handler2 =
 static auto device_info_handler =
     can_device_info::DeviceInfoHandler(message_writer_1, 0);
 static auto device_info_dispatch_target =
-    DispatchParseTarget<decltype(device_info_handler),
-                        can_messages::DeviceInfoRequest>{device_info_handler};
+    DispatchParseTargetNode<decltype(device_info_handler),
```
sfoster1, line 174: we shouldn't use the `0x50` literal here - let's use the enum values

head/firmware/can_task.cpp
```diff
@@ -171,39 +171,40 @@ static auto can_move_group_executor_handler2 =
...
+                            can_messages::DeviceInfoRequest>{
+        device_info_handler, 0x50};
 
-static auto motor_dispatch_target = DispatchParseTarget<
+static auto motor_dispatch_target = DispatchParseTargetNode<
```
sfoster1, line 178: can we rename these to `motor_dispatch_target_left` and `_right` or something

can/tests/test_dispatch.cpp
```diff
@@ -164,4 +164,43 @@ SCENARIO("DispatchParseTarget") {
...
+    GIVEN(
+        "A DispatchParseTargetNode that accepts hearbeat request and response "
+        "types for head's left motor") {
+        auto l = Handler{};
+        auto buff = BufferType{};
+        auto node = 0x60;
+        auto subject = DispatchParseTargetNode<Handler, HeartbeatRequest,
+                                               HeartbeatResponse>(l, node);
+
+        WHEN("Given a HeartbeatRequest") {
```
sfoster1, line 194: let's also add a test that if the node id part is different but the heartbeat message id part is correct, this parser does _not_ call the handler

include/can/core/dispatch.hpp
```diff
@@ -81,6 +81,46 @@ requires HandlesMessages<HandlerType, MessageTypes...> &&
...
+/**
+ * A CanMessageBufferListener for specific nodes that parses messages and
+ * notifies a Handler
+ * @tparam HandlerType A HandlesMessages type
+ * @tparam MessageTypes The Message types this handles
+ */
+template <typename HandlerType, CanMessage... MessageTypes>
+requires HandlesMessages<HandlerType, MessageTypes...> &&
+    (!std::movable<HandlerType> &&
+     !std::copyable<HandlerType>)class DispatchParseTargetNode {
```
sfoster1, line 93: [nit, feel free to ignore this] maybe `NodeDispatchParseTarget`?

sfoster1 (changes requested):
Looks good but some small but important style bits here

include/can/core/dispatch.hpp
```diff
@@ -136,4 +136,35 @@ class Dispatcher {
...
 
+/**
+ * A CanMessageBufferListener that will dispatch messages to other
+ * CanMessageBufferListeners
+ * @tparam Listener CanMessageBufferListener objects
+ */
+template <CanMessageBufferListener... Listener>
+class NodeDispatcher {
+  public:
+    using ArbitrationIdTest = bool (*)(uint32_t identifier, uint16_t node_id);
```
amitlissack, line 147: There's no need to have this method take a `node_id`. 

Ideally, this should be:
`using ArbitrationIdTest = std::function<bool (uint32_t identifier)>;`

What that means is that the argument passed as `test` can be either:
- A callable object (implementing operator() )
```
struct CheckForNodeId {
  NodeId node_id;
  bool operator()(uint32_t arbitration_id) {
    // omitting the check.....
    return node_id_is_in_arbitration_id;
  }
};

/// Constructed like...
CheckForNodeId{.node_id=NodeId::head_left};
```

- A plain old function.
```
bool check_for_node_id(uint32_t) {
  // omitting the check.....
  return node_id_is_in_arbitration_id;
}
```

include/can/core/dispatch.hpp
```diff
@@ -136,4 +136,35 @@ class Dispatcher {
...
+    explicit NodeDispatcher(ArbitrationIdTest test, uint16_t node_id,
+                            Listener&... listener)
```
amitlissack, line 149: `node_id` should not be passed into this constructor or be a member variable.

include/can/core/dispatch.hpp
```diff
@@ -136,4 +136,35 @@ class Dispatcher {
     std::tuple<Listener&...> registered;
 };
 
+/**
+ * A CanMessageBufferListener that will dispatch messages to other
+ * CanMessageBufferListeners
+ * @tparam Listener CanMessageBufferListener objects
+ */
+template <CanMessageBufferListener... Listener>
+class NodeDispatcher {
```
amitlissack, line 145: This is a misnomer. The point of this class is to be a `Dispatcher` that can apply a filter based on the arbitration id. The filter is just a function that returns true if the message should be passed to the `CanMessageBufferListener` or false if it should be ignored.

We discussed extending the existing `Dispatcher` class to accept the `ArbitrationIdTest` argument.  

In my opinion you either have a class that accepts a `ArbitrationIdTest` and a single `CanMessageBufferListener` OR modify `Dispatcher` to accept the `ArbitrationIdTest`. A copy and paste of the `Dispatcher` with an additional argument is overly complicated.

amitlissack (commented):
A step in the right direction, but I still think it's too complex and not flexible enough.

include/can/core/dispatch.hpp
sotaoverride: changing test for ArbitrationIdTest accepting Dispatcher

head/firmware/can_task.cpp
```diff
@@ -174,42 +175,70 @@ static auto device_info_dispatch_target =
...
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-union-access)
+        return (_node_id == static_cast<uint16_t>(node_id));
+    }
+};
+
+CheckForNodeId CheckForNodeId_left{.node_id = NodeId::head_left};
+
+CheckForNodeId CheckForNodeId_right{.node_id = NodeId::head_right};
+
+static auto test(uint32_t arbitration_id) { return true; }
```
sfoster1, line 225: Maybe this should be called something like `all_pass`?

sfoster1 (changes requested):
This makes sense to me but we definitely need more tests, and we need to make sure that all the predicates in the end executables aren't named things like `test`. Maybe the simple ones could just be lambdas passed in directly, like `static auto main_dispatcher = Dispatcher([](auto _) -> bool { return true; }, dispatcher_right_motor, dispatcher_left_motor);`

In general we should make sure the names here all make sense - it's usually a bad sign if you have a comment right before a variable is a two word phrase that could just be the name of the variable.

head/firmware/can_task.cpp
```diff
@@ -153,63 +155,108 @@ static motor_class::Motor motor_left{
...
 
-static auto motion_group_executor_dispatch_target2 =
-    DispatchParseTarget<decltype(can_move_group_executor_handler2),
+static auto motion_group_executor_dispatch_target_left =
+    DispatchParseTarget<decltype(can_move_group_executor_handler_left),
                         can_messages::ExecuteMoveGroupRequest>{
-        can_move_group_executor_handler2};
+        can_move_group_executor_handler_left};
+
+struct CheckForNodeId {
```
sfoster1, line 226: Can we drop a comment here that messages to `head 0x50` are really "head-both"? might come up in the future if we have a "control plane" concept or something

sfoster1 (approved):
I really like this! I don't think it's brute forcing, I think it's explicit

head/firmware/can_task.cpp
```diff
@@ -153,63 +155,115 @@ static motor_class::Motor motor_left{
...
+        auto _node_id = static_cast<uint16_t>(arb.parts.node_id);
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-union-access)
+        auto tmp = static_cast<uint16_t>(node_id);
+        return ((_node_id == tmp) ||
+                (_node_id == static_cast<uint16_t>(NodeId::broadcast)) ||
+                (_node_id == static_cast<uint16_t>(NodeId::head)));
+    }
+};
+
+CheckForNodeId CheckForNodeId_left{.node_id = NodeId::head_left};
```
amitlissack, line 246: Can we use lower camel case for var names?

gantry/firmware/can_task.cpp
```diff
@@ -160,9 +160,11 @@ static auto motion_group_executor_dispatch_target =
                         can_messages::ExecuteMoveGroupRequest>{
         can_move_group_executor_handler};
 
+static auto test(uint32_t arbitration_id) { return true; }
```
amitlissack, line 163: Can we use a lambda in the `dispatcher` ?

pipettes/firmware/can_task.cpp
```diff
@@ -144,10 +144,13 @@ static auto device_info_dispatch_target =
     DispatchParseTarget<DeviceInfoHandler, can_messages::DeviceInfoRequest>{
         device_info_handler};
 
+static auto test(uint32_t arbitration_id) { return true; }
```
amitlissack, line 147: A lambda again, please?

can/tests/test_dispatch.cpp
```diff
@@ -35,7 +36,13 @@ SCENARIO("Dispatcher") {
         auto l2 = Listener{};
         auto buff = BufferType{1};
         uint32_t arb_id = 1234;
-        auto subject = Dispatcher(l1, l2);
+        struct CheckForNodeId {
+            NodeId node_id;
+            auto operator()(uint32_t arbitration_id) const { return true; }
```
amitlissack, line 41: Since this just returns true, can we use a lambda in the `Dispatcher` declaration that simply returns true?

can/tests/test_dispatch.cpp
```diff
@@ -49,6 +56,78 @@ SCENARIO("Dispatcher") {
...
+                REQUIRE(l1.iter == buff.begin());
+                REQUIRE(l1.limit == buff.end());
+                REQUIRE(l2.id == static_cast<uint32_t>(arbitration_id.id));
+                REQUIRE(l2.iter == buff.begin());
+                REQUIRE(l2.limit == buff.end());
+            }
+        }
+    }
+
+    GIVEN("a dispatcher with two listeners for head_right ") {
```
amitlissack, line 96: Perhaps instead of a test for `head_right` we have a test that proves that the listener is not called when the arbitration id test method returns false?

amitlissack (changes requested):

amitlissack (approved):
Very nice.

PR #172, opened 2021-12-01
RET-205: add python bindings

sfoster1:
We face this problem of coordinating repos, where the python side of
things needs to know many constants and message data structures that
need to match c++. It's really annoying to keep these synchronized, and
really annoying to do cross-repo work. The solution is to move the
bindings to the same place as the firmware, and (eventually) make a
single source of truth for the messages by having the python generate
the c++ (since that code is written already).

This adds a new directory python/ containing a new python package
opentrons_ot3_firmware, which contains the section of opentrons_hardware
that just contains static object definitions for messages and also some
constants, along with its tests.

The python `constants.py` and associated scripts is used to generate the canbus
ids header.


python/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
+set(SHELL_FOR_PYTHON "/bin/sh" CACHE STRING "shell to use for calling python (supports pyenv)")
```
amitlissack, line 1: Can we space lines out and add some one liner comments about what is going on?

amitlissack (approved):
I love this!

Laura-Danielle (commented):
So I understand this properly, the new way of adding CAN messages would be:

1. Add the message ids/definitions in `python` of this repo
2. Have that code generate the C++ code
3. Also synchronize it with the monorepo, or are we eventually adding all CAN Bus handling stuff into this repo?

sfoster1:
> So I understand this properly, the new way of adding CAN messages would be:
> 
>     1. Add the message ids/definitions in `python` of this repo

Yup

>     2. Have that code generate the C++ code

Yeah, but you don't have to do anything special for this - cmake should notice that the python was edited and regenerate everything the next time you build.
 
>     3. Also synchronize it with the monorepo, or are we eventually adding all CAN Bus handling stuff into this repo?

The next step is to have the monorepo use this package, yeah, so in theory we'll be able to update this and the monorepo will pick it up. Exactly how that works I haven't decided yet.

PR #173, opened 2021-12-02
refactor(CAN): arbitration id is a class instead of a union

amitlissack:
When in Rome, you don't use Unions and bit fields.


include/can/core/arbitration_id.hpp
```diff
@@ -5,33 +5,79 @@
...
+                                                << message_id_shift;
+
+  private:
+    /**
+     * Set bits in the id.
+     * @param val The value.
+     * @param mask The mask to apply.
+     * @param shift The amount to left shift masked value
+     */
+    void set(uint32_t val, int mask, int shift) {
```
sotaoverride, line 66: can we use uint32_t for all these params?

sotaoverride (approved):
Looks good. Can we just use uint32_t for mask and shift, or am I missing something?

include/can/core/arbitration_id.hpp
amitlissack: Good point

can/tests/test_arbitratration_id.cpp
```diff
@@ -0,0 +1,55 @@
+#include "can/core/arbitration_id.hpp"
```
sfoster1, line 1: filename should be `test_arbitration_id.cpp`

can/tests/CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ add_executable(
         test_can_bus.cpp
         test_can_message_buffer.cpp
         test_dispatch.cpp
+        test_arbitratration_id.cpp
```
sfoster1, line 15:
```suggestion
        test_arbitration_id.cpp
```

include/can/core/arbitration_id.hpp
```diff
@@ -5,33 +5,79 @@
...
 
-/**
- * Node id mask.
- */
-constexpr auto node_id_mask = ArbitrationId{
-    .parts{.function_code = 0, .node_id = 0xFF, .message_id = 0, .padding = 0}};
+    [[nodiscard]] auto function_code() const -> uint8_t {
+        return get(function_code_bit_mask, function_code_shift);
+    }
+    [[nodiscard]] auto node_id() const -> uint8_t {
```
sfoster1, line 30: we should either have this return an enum member or keep it this way and have a second one that returns an enum member and handles incorrect values

include/can/core/arbitration_id.hpp
```diff
@@ -5,33 +5,79 @@
...
+        return get(node_id_bit_mask, node_id_shift);
+    }
+    [[nodiscard]] auto message_id() const -> uint16_t {
```
sfoster1, line 33: same deal w/r/t enum members

include/can/core/arbitration_id.hpp
```diff
@@ -5,33 +5,79 @@
...
+    }
+
+    void function_code(uint8_t v) {
+        set(v, function_code_mask, function_code_shift);
+    }
+    void node_id(uint8_t v) { set(v, node_id_mask, node_id_shift); }
+    void message_id(uint16_t v) { set(v, message_id_mask, message_id_shift); }
+
+    // Bits 0-6
+    static constexpr auto function_code_mask = 0x7F;
```
sfoster1, line 44: do these need to be public?

include/can/core/arbitration_id.hpp
```diff
@@ -5,33 +5,79 @@
...
+        return get(function_code_bit_mask, function_code_shift);
+    }
+    [[nodiscard]] auto node_id() const -> uint8_t {
+        return get(node_id_bit_mask, node_id_shift);
+    }
+    [[nodiscard]] auto message_id() const -> uint16_t {
+        return get(message_id_bit_mask, message_id_shift);
+    }
+
+    void function_code(uint8_t v) {
```
sfoster1, line 37: These IMO should take enum members rather than `uint8`s. It reinforces that these are _not_ general ints but specific values, and it means we don't have to clutter up call sites with `static_cast`s.

sfoster1 (changes requested):
So much better! Let's use more enums though

include/can/core/arbitration_id.hpp
amitlissack: Yeah. We use them for setting up the hardware can filters.

sfoster1 (approved):
Nice! At some point we should probably get some input validation but this is good for now!

PR #180, opened 2021-12-09
feat(common): enable software timers in FreeRTOS

Laura-Danielle:
Quick PR to add the ability to use software timers in FreeRTOS.


include/common/core/freertos_timer.hpp
```diff
@@ -0,0 +1,39 @@
...
+// FreeRTOS tick rates less than 1000 Hz.
+template <TickType_t timer_period = pdMS_TO_TICKS(2)>
+class FreeRTOSTimer {
+  public:
+    /*
+     * A software timer class. The priority of software timers in FreeRTOS is
+     * currently set to 6. Any tasks utilizing this timer should have either the
+     * same priority or higher priority than 6 for execution.
+     */
+    FreeRTOSTimer(const char* name, TimerCallbackFunction_t callback)
```
sfoster1, line 18: you should c++ify this imo. take a `std::function` instead of a `TimerCallbackFunction_t`, make an `extern "c" void callback(void* me) { me->do_callback(); }` that you pass to `xTimerCreateStatic`. That way callers can drop in a lambda, a wrapper around a member function pointer, whatever they want.

sfoster1 (approved):
This will work but we should C++ify the interface

sfoster1 (approved):
This looks good to me but keep in mind that the template won't get compiled until instantiated which doesn't happen in this pr - could you add a test or other toy instantiation just to make sure it compiled?

Laura-Danielle:
Rebasing is hurting my soul. I'm going to open a new PR.

PR #181, opened 2021-12-09
fix(yaml): bad yaml

amitlissack:


sfoster1 (approved):
oof. sorry. i think `python` is fine but maybe `python3` is better. if this is to fix the emulation tests, that won't help because the docker images that this builds in don't have python at all - needs to be fixed on the action side

PR #182, opened 2021-12-10
feat(can): simulate hardware filtering

amitlissack:
`SimCANBus` needs to implement `add_filter` to support the HAL can message filtering.


sfoster1 (approved):
lgtm!

PR #183, opened 2021-12-10
refactor(can): a little clean up

amitlissack:
move device_info handler to the proper directory.
use DispatchTarget type aliases for cleanerness.


sfoster1 (approved):
noice

PR #184, opened 2021-12-10
refactor(motor-control): convert velocity and acceleration from can message to use steps per tick

Laura-Danielle:
## Overview

Steps per mm is a float. We should convert it to fixed point and multiply velocity/acceleration (which is coming in from python as `mm/tick` and `mm/tick^2`) to convert them to `steps/tick` and `steps/tick^2` respectively.

## Changes
- Shift the current position in the motor interrupt handler such that it's in whole units
- Switch the direction of the Z/A board to have it move down when given positive direction
- Added the conversion to steps per mm
- Added tests to check the steps per mm functions

## Future
- We should make the fixed point numbers structs so we can simply pass in numerical values and not worry about shifting values
- We should have more comprehensive fixed point math functions.


include/motor-control/core/motion_controller.hpp
```diff
@@ -43,12 +44,7 @@ class MotionController {
     ~MotionController() = default;
 
     void move(const can_messages::AddLinearMoveRequest& can_msg) {
-        Move msg{.duration = can_msg.duration,
-                 .velocity = can_msg.velocity,
-                 .acceleration = can_msg.acceleration,
-                 .group_id = can_msg.group_id,
-                 .seq_id = can_msg.seq_id};
-        queue.try_write(msg);
+        convert_move_to_steps(can_msg) queue.try_write(msg);
```
amitlissack: I don't understand this statement.

include/motor-control/core/motion_controller.hpp
```diff
@@ -79,7 +75,21 @@ class MotionController {
     MotionConstraints motion_constraints;
     GenericQueue& queue;
     CompletedQueue& completed_queue;
-    uint32_t steps_per_mm{0};
+    sq0_31 steps_per_mm{0};
+
+    Move convert_move_to_steps(
+        const can_messages::AddLinearMoveRequest& can_msg) {
```
amitlissack, line 82: I don't understand what this type is.  It looks an anonymous function? Is that a thing in c++?

amitlissack: I'm an idiot

amitlissack (approved):
Looks good to me.

Nice tests.

PR #185, opened 2021-12-13
fix(pipettes): clean-up pipette GPIO configurations

Laura-Danielle:
## Overview
- make sure the correct GPIO clocks are getting called
- make sure SPI is using the correct GPIO pins when sending configuration settings over to the driver
- cleaned up a few other things.


sfoster1 (approved):
nice

PR #186, opened 2021-12-13
feat(common): add FreeRTOS software timer support

Laura-Danielle:


pipettes/firmware/main.cpp
```diff
@@ -9,12 +9,19 @@
...
 
+static void callback() {}
+
 auto main() -> int {
     HardwareInit();
     RCC_Peripheral_Clock_Select();
     MX_ICACHE_Init();
 
+    // Dummy timer code, should be removed once SW timers are implemented
+    auto timer = freertos_timer::FreeRTOSTimer("test timer", callback);
```
sfoster1, line 23: do we want to commit this

sfoster1 (approved):
Looks good but do we want to commit that software one? I think it's fine if you do but the comment does say not to

pipettes/firmware/main.cpp
Laura-Danielle: Nah was leaving it there till someone reviewed it. I was planning to take it out before merging.

PR #187, opened 2021-12-15
feat(can): CAN Writer Task

amitlissack:
**This is a PR into a feature branch for multi task refactor.**

First step in refactoring our FW to use multiple tasks with queues being the interface between the tasks.

- Adds a task that writes CAN messages using a CanBus reference. 
- Each subsystem implements a start_writer and start_reader method to start the CAN tasks.
- `MessageWriter` is instantiated with a message queue rather than a CanBus reference.

Not tested on HW yet. Only tested with Gantry simulator.


include/can/core/freertos_can_dispatch.hpp
```diff
@@ -77,7 +77,8 @@ struct FreeRTOSCanDispatcherTarget {
           buffer_target{message_buffer},
           parse_target{handler},
           poller{message_buffer, parse_target},
-          task{poller} {}
+          task_control{},
```
Laura-Danielle, line 80: nice refactor

include/can/core/freertos_sender_task.hpp
```diff
@@ -0,0 +1,65 @@
...
+    can_messages::ResponseMessageType message;
+};
+
+/**
+ * Entry point for a CAN sender class.
+ * @tparam QueueImpl
+ */
+template <template <class> class QueueImpl>
+requires MessageQueue<QueueImpl<TaskMessage>, TaskMessage>
+class MessageSenderTask {
```
Laura-Danielle, line 25: so any task that needs to send messages back to the can bus would build a message sender task?

include/gantry/core/can_task.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include "can/core/freertos_sender_task.hpp"
+#include "can/core/message_handlers/device_info.hpp"
+#include "can/core/message_handlers/motor.hpp"
+#include "can/core/message_handlers/move_group.hpp"
+#include "can/core/message_handlers/move_group_executor.hpp"
+#include "common/core/freertos_message_queue.hpp"
+#include "common/core/freertos_task.hpp"
+#include "motor-control/core/motor.hpp"
+
+namespace can_bus {
```
Laura-Danielle, line 13: once we refactor pipettes and head to have interfaces & the tasks living in core, then we should be able to have this header file shared, right?

Laura-Danielle (approved):
Looks great. Just had a few questions.

gantry/core/can_task.cpp
```diff
@@ -47,23 +52,37 @@ static auto motion_group_executor_dispatch_target =
...
 
-    poller();
+    auto poller = can_task::CanReaderTaskEntry{canbus, reader_message_buffer};
+    return freertos_task::FreeRTOSTask<can_task::reader_task_stack_depth,
+                                       decltype(poller)>{
+        poller, reader_task_control, 5, "can task"};
 }
 
-auto static task = freertos_task::FreeRTOSTask<512, 5>("can task", task_entry);
+auto can_task::start_writer(can_bus::CanBus& canbus)
```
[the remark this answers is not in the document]
amitlissack: I'm sorry, but I don't understand. Which message writer code do you mean?

head/firmware/can_task.cpp
```diff
@@ -280,34 +285,42 @@ extern "C" void motor_callback_glue() {
     motor_interrupt_right.run_interrupt();
 }
 
-[[noreturn]] void task_entry() {
-    can_bus_1.set_incoming_message_callback(nullptr, callback);
+can_task::CanReaderTaskEntry::CanReaderTaskEntry(can_bus::CanBus& bus)
```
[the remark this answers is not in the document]
amitlissack: Good point. I'll change it

include/can/core/freertos_sender_task.hpp
amitlissack: No. There'll be one in each subsystem. It will be shared among all tasks.

include/gantry/core/can_task.hpp
amitlissack: Possibly.

amitlissack: I hope so, but I get the feeling that the interfaces will evolve.

include/can/core/freertos_sender_task.hpp
Laura-Danielle: I see

gantry/core/can_task.cpp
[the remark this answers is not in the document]
Laura-Danielle: Sorry I should've been more clear. Since we have separate reader and writer tasks now, I was thinking it might make things a little more readable if we keep individual tasks in separate files. So can task writer and implementation could be in a file called `can_writer_task.cpp` and the reader task could be called `can_reader_task.cpp` etc.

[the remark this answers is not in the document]
amitlissack: Oh. Doesn't matter to me either way. If you prefer two separate files I'll do it.

include/can/core/freertos_sender_task.hpp
sfoster1: In that case we probably want to delete the copy constructor (might have to change the builder functions to return with `std::move`)

include/common/core/freertos_task.hpp
```diff
@@ -8,31 +8,86 @@
...
+     * @param control The task control block.
+     * @param priority The tasks priority to use
+     * @param task_name Task name.
+     */
+    FreeRTOSTask(EntryPoint& entry, FreeRTOSTaskControl<StackDepth>& control,
+                 UBaseType_t priority, const char* task_name)
+        : entry{entry}, control{control} {
+        start(priority, task_name);
+    }
+    FreeRTOSTask(FreeRTOSTask&) noexcept = default;
```
sfoster1, line 60: do we really want this to be copyable? i'm not sure if it will autogenerate this or not, but if we want the semantics of this class to be an RAII-style task where the constructor also starts the task, it really doesn't seem like it should be copyable, or if it is the copy indicates creating a new underlying task.

sfoster1 (changes requested):
I think we probably want to remove some copy ctors here and take a look at the stack overwriting. IMO there are two right ways to handle the stack thing:

1. task starting functions are provisioned with pointers to data to initialize, c "constructor function" style
2. the creator functions are called at static-constructor time by having them owned by a `static` owner object in main.cpp, e.g. `struct Tasks { CanWriterTask writer_task; ... }` and, unfortunately, we stop doing the RAII idiom for the tasks and have separate `.start` methods that are called in `main`
3. a fusion of 1 and 2 where you define the task aggregator object as above, but in main.cpp it's allocated as like `auto tasks = std::variant<std::monostate, Tasks>{}` or something and gets move-initialized in the body of `main`. IDK I kind of hate this

A really nice refactor though!

sfoster1:
I'm also getting a little lost in the dataflow and the tasks; would you mind getting a diagram or something together for can message dataflow?

amitlissack:
> I'm also getting a little lost in the dataflow and the tasks; would you mind getting a diagram or something together for can message dataflow?

gantry/firmware/main.cpp
```diff
@@ -20,5 +23,9 @@ auto main() -> int {
     RCC_Peripheral_Clock_Select();
     utility_gpio_init();
 
+    interfaces::initialize();
+    auto can_writer = can_task::start_writer(interfaces::get_can_bus());
```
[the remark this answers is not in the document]
amitlissack: Gadzooks. Forgot about that.

include/can/core/freertos_sender_task.hpp
sfoster1: to be clear, by "deleting the copy constructor" I mean explicitly doing so by doing `MessageSenderTask(const MessageSenderTask&) = delete;` and `(...) operator=(...) = delete;`

sfoster1 (approved):
We can merge this, but I'd love to see those copy constructors deleted first.

PR #188, opened 2021-12-15
refactor(can-sim): can simulation transport create method

amitlissack:
Create a `create` method to build the socket transport for `SimCanBus`. 

- Can simulator app (FMA loopback) uses the new create method. Can use plain old sockets.
- Gantry simulator uses new create method.

### DIFF
--- a/can/simlib/transport.cpp
+++ b/can/simlib/transport.cpp
@@ -1,4 +1,5 @@
 #include "can/simlib/transport.hpp"
+
 #include "common/core/freertos_synchronization.hpp"
 #ifdef USE_SOCKETCAN
 #include "can/simlib/socketcan_transport.hpp"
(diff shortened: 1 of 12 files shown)


can/simlib/transport.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include "can/simlib/socket_transport.hpp"
+#endif
+
+/**
+ * Create simulating bus transport
+ * @return pointer to bus transport
+ */
+auto can_transport::create()
+    -> std::shared_ptr<can_transport::BusTransportBase> {
+#ifdef USE_SOCKETCAN
```
sfoster1, line 16: if we're building this in to the same function and using getenv anyway, should we just make this fully runtime-dynamic? or can SocketCanTransport not compile without linux?

include/can/simlib/sim_canbus.hpp
```diff
@@ -20,7 +21,9 @@ using namespace freertos_task;
  */
 class SimCANBus : public CanBus {
   public:
-    explicit SimCANBus(can_transport::BusTransportBase& transport)
+    using TranportType = std::shared_ptr<can_transport::BusTransportBase>;
```
sfoster1, line 24: typo

sfoster1 (approved):
Looks good but there's a quick typo (`s/tranport/transport`) and can we make the socketcan/can via socket determination fully runtime?

can/simlib/transport.cpp
amitlissack: I wish. We cannot compile `SocketCanTransport` on non linux machine.

PR #189, opened 2021-12-15
refactor(can): CAN arbitration id refactor

amitlissack:
Moving originating node id into arbitration id:

Haven't tested on hardware.


sfoster1 (approved):
This avoids the problems with CAN transmission collision when multiple nodes reply simultaneously to a broadcast! hooray!

PR #191, opened 2021-12-16
refactor(tasks): multiple tasks

amitlissack:
This is a PR into `can-sender-task` branch. Which itself is a PR into a feature branch.

This PR refactors ot3 subprojects to use multiple FreeRTOS tasks.

The added tasks:
- motor driver (messages that interface to the MotorDriver class).
- motion controller (messages that interface to the MotionController class).
- move group (messages that interact with the move groups)
- move status reporter (notifier of movement status messages....currently just the move completed message)
- eeprom (just pipettes)

Each task defines a message queue and task message type. Additionally it defines a `TaskClient` concept. This describes a class that can write messages to the queue.

Each subproject defines its own `AllTasks` struct which holds pointers to each task as well as a `QueueClient` struct which holds pointers to all the message queues. `QueueClient` implements all the `TaskClient` concepts required by each subsystem.

**This sort of builds ~(tests don't pass).~ **

I've tested with the Gantry simulator, but not hardware.

Next PRs will add simulators for Pipettes and Head.


include/common/core/freertos_message_queue_poller.hpp
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * A FreeRTOSTask entry point that polls a MessageQueue and notifies a Handler
+ * with new messages.
+ *
+ * @tparam Message Message type
+ * @tparam Handler Handler Type
+ */
+template <typename Message, typename Handler>
+requires MessageHandler<Handler, Message>
+class FreeRTOSMessageQueuePoller {
```
sfoster1, line 27: bit of a nit but this isn't really and doesn't really need to be a poller, it's fully reactive I think

sfoster1 (approved):
Big step forward! Let's merge this but put a pin in taking a very close look at specifically stuff that got moved from legacy can_task.cpp to other places when we merge the main feature branch to make sure that some bug fixes from after the split don't get lost.

PR #192, opened 2021-12-16
refactor(sim): remove an annoying build warning

amitlissack:
template-ize freertos timeout arg as it's a different type in FW and the posix port.


Laura-Danielle (approved):
Nice. Looks good to me!

PR #193, opened 2021-12-16
fix(gantry): change the microstepping to 16

Laura-Danielle:
Change to chopconf configuration actually didn't get saved properly after rebasing changes in the
gantry firmware.


sfoster1 (approved):
Looks good - ended up being a general bugfixing task

PR #195, opened 2021-12-22
feat(pipettes): initial pipette sensor board bringup

Laura-Danielle:
## Overview
Initial work to bring-up all of the sensors available on the pipette sensor board. Additional work is needed for bringing up a sensor task and sending/receiving messages through CAN Bus. Those tickets will be added to Jira shortly.

Closes RET-46 RET-47 and RET-48

### DIFF
--- a/pipettes/firmware/i2c.c
+++ b/pipettes/firmware/i2c.c
@@ -3,14 +3,53 @@
 #include "common/firmware/errors.h"
 
 static I2C_HandleTypeDef hi2c;
+static I2C_HandleTypeDef hi2c3;
+
+
+/**
+ * @brief Tip Sense GPIO Initialization Function
+ * @param None
+ * @retval None
+ */
+void tip_sense_gpio_init() {
+    /* GPIO Ports Clock Enable */
+    __HAL_RCC_GPIOA_CLK_ENABLE();
+
+    /*Configure GPIO pin : A10 */
+    GPIO_InitTypeDef GPIO_InitStruct = {0};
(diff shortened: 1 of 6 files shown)


pipettes/firmware/i2c.c
```diff
@@ -53,3 +112,65 @@ I2C_HandleTypeDef* MX_I2C_Init()
...
+
+void enable_eeprom() {
+    HAL_GPIO_WritePin(GPIOB, GPIO_PIN_8, GPIO_PIN_SET);
+}
+
+void disable_eeprom() {
+    HAL_GPIO_WritePin(GPIOB, GPIO_PIN_8, GPIO_PIN_RESET);
+}
+
+I2CHandlerStruct i2c_setup(PipetteType pipette_type) {
```
amitlissack, line 164: It is very un-C like to pass or return structs by value. 

I realize this function is only called once so this isn't a big deal, but for the purpose of writing in the C idiom I suggest you rewrite this function to accept a pointer to a `I2CHandlerStruct`.

Laura-Danielle: That makes sense! Generally, is there anything you would want to return by value aside from a primitive type?

amitlissack: In C, just primitives.

sfoster1: You're right but also that's a thing I've always hated about C, and it's kind of held over from older implementations and standards. I'm ok with either way.

sfoster1 (approved):

PR #196, opened 2021-12-23
feat(head): head simulator

amitlissack:
This PR builds upon #191 to add a HEAD subsystem simulator

- added cmake project for head
- can_task.cpp is moved to core
- simulated hardware is initialized in simulator/main.cpp


Laura-Danielle (approved):
lgtm

sfoster1 (approved):
Nice! We should take a close look at the eventual feature branch merge to make sure bugfixes don't get lost in merged files.

Hopefully a future refactor can get the motor driver configs and similar shared between firmware and sim.

PR #198, opened 2021-12-28
feat(pipettes): pipettes simulator

amitlissack:
PR builds on #196 to add pipettes simulator

- i2c now has abstract base class.
- can task is moved to core
- task set up is in core


common/simulation/eeprom.cpp
```diff
@@ -0,0 +1,21 @@
...
+    for (auto i = 0; i < size; i++) {
+        this->data.push_back(data[i]);
+    }
+    return true;
+}
+
+auto sim_i2c::SimEEProm::master_receive(uint16_t dev_address, uint8_t *data,
+                                        uint16_t size, uint32_t timeout)
+    -> bool {
+    for (uint32_t i = 0; i < size && i < this->data.size(); i++) {
```
sfoster1, line 17: can we `std::copy` this

common/simulation/eeprom.cpp
```diff
@@ -0,0 +1,21 @@
+
+#include "common/simulation/eeprom.hpp"
+
+auto sim_i2c::SimEEProm::master_transmit(uint16_t dev_address, uint8_t *data,
+                                         uint16_t size, uint32_t timeout)
+    -> bool {
+    this->data.clear();
```
sfoster1, line 7: doesn't matter that much but doing
```
this->data.clear();
this->data.reserve(size);
std::copy(...)
```
 is a little more idiomatic

sfoster1 (approved):
Looks good aside from some nits about using std::copy

PR #199, opened 2022-01-03
fix: Fix poetry errors

DerekMaggio:
# Overview

Fixed `OT-3 Emulator` test failing due to issue with opentrons-emulation

# Changelog

- Bumped version to 1.3
- Note that [v1.3](https://github.com/Opentrons/opentrons-emulation/releases/tag/v1.3) is a temporary fix and source code for it will be removed with v2.0 

# Review requests

None

# Risk assessment

Low as long as the Github Actions work, which they do


sfoster1 (approved):
:raised_hands:

PR #203, opened 2022-01-06
feat(HEAD): head task refactor to support presence sensing

amitlissack:
## Summary

The head FW has three node ids: HEAD, HEAD_LEFT, and HEAD_RIGHT.  This PR adds ability for HEAD FW to respond as node id HEAD rather than either HEAD_LEFT or HEAD_RIGHT.

The presence sensing driver should not be addressable as LEFT or RIGHT. 

## Changes
Create a new task and queue collection for head that is independent of motor.


sfoster1 (approved):
Nice!

PR #206, opened 2022-01-10
head: support adc readings from tool connections

sotaoverride:
Sets up ADC to enable presence sensing mechanism on head.  Also sets ReadPresenceSensingVoltageRequest/Response CAN messages requirements - presence sensing driver task, dispatcher, queue client, and handlers.


common/firmware/adc/adc.c
```diff
@@ -0,0 +1,155 @@
+#include "common/firmware/adc.h"
```
sfoster1, line 1: This C file is entirely specific to the hardware of the head project. It configures the ADC in a way that is only relevant to head, it maps pinmux in a way that is only relevant to head, and it defines helper functions like `read_voltage_[gripper|a_motor|z_motor]` that are not only specific to head in implementation but specific to head in semantics. This file and its header really need to either in the `head` subproject.

common/firmware/adc/adc.c
```diff
@@ -0,0 +1,155 @@
...
+    sConfig.SingleDiff = ADC_SINGLE_ENDED;
+    sConfig.OffsetNumber = ADC_OFFSET_NONE;
+    sConfig.Offset = 0;
+    if (HAL_ADC_ConfigChannel(handle, &sConfig) != HAL_OK) {
+        Error_Handler();
+    }
+}
+
+static uint32_t adc_read_voltage(ADC_HandleTypeDef* adc_handle, uint32_t adc_channel) {
+    uint32_t return_value = 4294967295;
```
sfoster1, line 136: [nit] this isn't k&r c, imo this line just muddles things. it's fine to do `return HAL_ADC_GetValue(adc_handle);` below.

gantry/firmware/interfaces.cpp
```diff
@@ -105,6 +105,7 @@ void interfaces::initialize() {
     if (initialize_spi(get_axis_type()) != HAL_OK) {
         Error_Handler();
     }
+    interfaces::get_motor().driver.setup();
```
sfoster1, line 108: can we put this in another pr

head/core/can_task.cpp
```diff
@@ -7,13 +7,15 @@
...
 #include "can/core/message_handlers/move_group.hpp"
+#include "can/core/message_handlers/presence_sensing.hpp"
 #include "can/core/messages.hpp"
 #include "common/core/freertos_message_queue.hpp"
 #include "common/core/freertos_task.hpp"
 #include "head/core/tasks.hpp"
 
 static auto& right_queues = head_tasks::get_right_queues();
 static auto& left_queues = head_tasks::get_left_queues();
+static auto& head_queues = head_tasks::get_queue_client();
```
sfoster1, line 18: [nit] kind of weird naming - `head_queues` feels to me like it _includes_ both left and right. maybe `common_queues` or `control_queues` or something?

head/core/tasks.cpp
```diff
@@ -48,16 +54,22 @@ void head_tasks::start_tasks(
...
         right_motion_controller,
-    motor_driver::MotorDriver& right_motor_driver) {
+    motor_driver::MotorDriver& right_motor_driver,
+    presence_sensing_driver::PresenceSensingDriver& presence_sensing_driver) {
     // Start the head tasks
     auto& can_writer = can_task::start_writer(can_bus);
     can_task::start_reader(can_bus);
 
+    auto& presence_sensing = presence_sensing_driver_task_builder.start(
+        5, presence_sensing_driver, head_queues);
```
sfoster1, line 64: is the `5` here the task priority? feels a bit too prioritized if so

common/firmware/adc/adc.c
```diff
@@ -0,0 +1,155 @@
...
+    ADC_set_chan(adc_channel, adc_handle);
+    HAL_ADC_Start(adc_handle);
+    HAL_ADC_PollForConversion(adc_handle, HAL_MAX_DELAY);
+    return_value = HAL_ADC_GetValue(adc_handle);
+
+    // HAL_ADC_Stop(&adc2);
+    return return_value;
+}
+
+uint32_t adc_read_voltage_z_motor() {
```
sfoster1, line 147: these don't return voltages, they return ADC readings. we should rename these functions to just like `adc_read_z_motor`

common/firmware/adc/adc_comms.cpp
```diff
@@ -0,0 +1,18 @@
...
+using namespace adc;
+/*
+ * Public Functions
+ */
+ADC::ADC(ADC_interface ADC_intf_instance1, ADC_interface ADC_intf_instance2)
+    : ADC_intf1(ADC_intf_instance1), ADC_intf2(ADC_intf_instance2) {}
+
+auto ADC::get_readings() -> adc::VoltageRead {
+    auto voltage_read = adc::VoltageRead{
+        .z_motor = static_cast<uint16_t>(adc_read_voltage_z_motor()),
```
sfoster1, line 14: this IMO is where we should convert the ADC readings into voltages - `adc_read_z_motor() * FULLSCALE_VOLTAGE / ADC_FULLSCALE_OUTPUT`. It could also happen in c I suppose

sfoster1 (changes requested):
Some inline requests - I think we should put some thought into the way the hardware bindings are structured and named, and where they live in the code - but there's a couple overarching things.

The first is that this code at no point actually reports the voltage present on the sense pins from the hardware interfaces. It only reports the raw ADC reading. Maybe there's a place for that interface on canbus (although I don't think so, since it's a pretty simple conversion that doesn't really have edge cases), but the main interface both to the task and to CAN should really report voltage after conversion, probably in a uint16 reporting millivolts since that fits the scale ok.

The second thing - can we add some testing for the ADC reader task itself? There's not much to test there because we're just reading an ADC, but the next step here is adding a predicate that filters on voltage levels to decide presence, and that will definitely need a test, so building the infrastructure for a mock ADC here and doing a quick test gets us a head start.

Also let's pop that motor-initialize line out into another PR.

gantry/firmware/interfaces.cpp
amitlissack: This looks like it is from another PR. Perhaps this branch should be rebased off main?

can/tests/test_dispatch.cpp
```diff
@@ -113,6 +113,40 @@ SCENARIO("Dispatcher") {
             }
         }
     }
+
+    GIVEN("a dispatcher with two listeners for head ") {
```
amitlissack, line 117: Is this really a new test?

head/firmware/main.cpp
```diff
@@ -152,7 +153,20 @@ extern "C" void motor_callback_glue() {
     motor_interrupt_left.run_interrupt();
     motor_interrupt_right.run_interrupt();
 }
+adc::ADC_interface ADC_intf2 = {
+
+    .ADC_handle = &adc2};
+
+// adc::ADC ADC_comms2(ADC_intf2);
```
amitlissack, line 160: Please remove comment

head/firmware/main.cpp
```diff
@@ -152,7 +153,20 @@ extern "C" void motor_callback_glue() {
...
+
+adc::ADC_interface ADC_intf1 = {
+
+    .ADC_handle = &adc1
+
+};
 
+adc::ADC ADC_comms(ADC_intf1, ADC_intf2);
+static presence_sensing_driver::PresenceSensingDriver psd(ADC_comms);
 auto main() -> int {
```
amitlissack, line 170: Can we space this out a bit please?

include/common/firmware/adc.h
```diff
@@ -0,0 +1,27 @@
...
+extern "C" {
+#endif  // __cplusplus
+
+extern ADC_HandleTypeDef adc1;
+extern ADC_HandleTypeDef adc2;
+
+void MX_ADC1_Init(ADC_HandleTypeDef* adc1);
+void MX_ADC2_Init(ADC_HandleTypeDef* adc2);
+void ADC_set_chan(uint32_t chan, ADC_HandleTypeDef* handle);
+//void adc_setup(ADC_HandleTypeDef adc1, ADC_HandleTypeDef adc2);
```
amitlissack, line 16: Delete commented out code.

common/firmware/adc/adc_comms.cpp
```diff
@@ -0,0 +1,18 @@
+#include "common/firmware/adc_comms.hpp"
```
sfoster1, line 1: The same thing with the adc c file goes for this. This C++ file is entirely bound up in the concerns of the `head`, and should be in `head/` not `common/`

head/firmware/main.cpp
```diff
@@ -152,6 +153,18 @@ extern "C" void motor_callback_glue() {
...
+    .ADC_handle = &adc2};
+
+adc::ADC_interface ADC_intf1 = {
+
+    .ADC_handle = &adc1
+
+};
+
+adc::ADC ADC_comms(ADC_intf1, ADC_intf2);
+static presence_sensing_driver::PresenceSensingDriver psd(ADC_comms);
```
sfoster1, line 167: nit: you use this style of initialization (`typename varname(ctor_args);`) here but the other style `auto varname = typename(ctor_args);` in `simulator/main.cpp`. can we use the same one (and ideally `auto`, but that's personal pref)

include/can/core/message_handlers/presence_sensing.hpp
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include "can/core/ids.hpp"
+//#include "can/core/message_writer.hpp"
```
sfoster1, line 4: remove commented line

head/firmware/adc/adc_comms.cpp
```diff
@@ -0,0 +1,24 @@
...
+/*
+ * Public Functions
+ */
+ADC::ADC(ADC_interface ADC_intf_instance1, ADC_interface ADC_intf_instance2)
+    : ADC_intf1(ADC_intf_instance1), ADC_intf2(ADC_intf_instance2) {}
+
+auto ADC::get_readings() -> adc::VoltageRead {
+    auto voltage_read = adc::VoltageRead{
+        .z_motor = static_cast<uint16_t>(
+            (static_cast<float>(adc_read_z_motor()) * FULLSCALE_VOLTAGE) /
```
sfoster1, line 15: This conversion is problematic. `adc_read_z_motor()` returns a `uint16_t` in [0, 4095] and `ADC_FULLSCALE_OUTPUT` is `4095`. This all makes sense. But `FULLSCALE_VOLTAGE` is `3.3f`, and the result is cast back into a `uint16_t`.

 That means that the end result of this conversion can only have the values `3`, `2`, `1`, and `0`, because this conversion is mapping [0, 4095] -> [0.0, 3.3] then making it an int again which will truncate.

One great way to fix this would just be to make `FULLSCALE_VOLTAGE` be the int `3300`. That makes the result of the conversion be an integer number of millivolts (and we should change the name of the struct or the struct members to reflect this).

presence-sensing/tests/test_adc.cpp
```diff
@@ -0,0 +1,16 @@
+#include "catch2/catch.hpp"
+#include "presence-sensing/tests/mock_adc.hpp"
+#include "presence-sensing/core/presence_sensing_driver.hpp"
+
+SCENARIO("get readings called on ADC"){
```
sfoster1, line 5: What does this test? It seems like it calls a mock instance directly and then checks the output of the mock, which just tests the mock and nothing else. That can be useful to make sure the mock is right but it doesn't really indicate anything about the shipping code.

sfoster1 (changes requested):
Couple notes on this:
- This voltage conversion won't have the results we want.
- We need some tests of the presence sensing driver itself.
- There's no need for empty `CMakeLists.txt` files, they don't really do anything. If that means that a directory is empty, fine, let's remove it.
- Need to add any new subdirectories of `head` and the top level `presence-sensing` subdirectory to https://github.com/Opentrons/ot3-firmware/blob/main/.github/workflows/head.yaml to trigger the github actions that cause CI builds
- Should `presence-sensing` really be a top level directory? It's only used by head. If it should be, then we'll need new github actions for building it and its test, triggered on changes to those files - see https://github.com/Opentrons/ot3-firmware/blob/main/.github/workflows/motorcontrol.yaml for an example

head/firmware/adc/adc_comms.cpp
sotaoverride: Changed `FULLSCALE_VOLTAGE` to int and calling the. struct `MillivoltsReadings` now

head/core/tasks.cpp
sotaoverride: 5 feels reasonable, given
`#define configMAX_PRIORITIES (7)
`
head/firmware/FreeRTOSConfig.h

head/firmware/adc/adc_comms.cpp
sfoster1: Nice!

head/presence-sensing/tests/test_adc.cpp
```diff
@@ -0,0 +1,13 @@
+#include "catch2/catch.hpp"
+#include "presence-sensing/core/presence_sensing_driver.hpp"
+#include "presence-sensing/tests/mock_adc.hpp"
+
+SCENARIO("get readings called on ADC") {
```
sfoster1, line 5: This test is just directly calling a mock. The only thing that it tests is that mock. It doesn't test anything about the actual code that will run on hardware. We should get rid of it.

sfoster1 (changes requested):
This is getting a lot closer. However,

- Having `head/presence-sensing` as a subdirectory doesn't really follow the schema we're looking for, and it's different from all the other applications. We could make it `head/core/presence-sensing` and `head/tests/presence-sensing`, for instance; that would be a new concept since none of the other applications have subdirectories under those, but it would follow the right directory schema. What I would personally say instead is to get rid of those subdirectories and just have the files in `head/firmware`, `head/core`, etc.
- Whatever you do with the stuff in `head`, the stuff in `includes/` needs to follow. We can't have `include/presence-sensing` when the `presence-sensing` subdirectory isn't top-level. Those include files should be in (to match what the current state is) `include/head/presence-sensing`, or if you want to flatten it, `include/head/core` or something like that.
- Let's do the adc->millivolts conversion in testable code (like in the presence sensing driver) and have the ADC hardware interface just return a uint16_t raw output, and test it.

sfoster1 (approved):
Nice! This looks good. Big step forward!

PR #207, opened 2022-01-10
feat(common): add a version constant to each subproject

Laura-Danielle:
## Overview
Quick change to add a version constant to the relevant subprojects. I kept getting confused by the
number passed into the `DeviceInfoHandler`.

## Review Requests

Let me know if there is a preferred way to do this.

### DIFF
--- a/head/core/can_task.cpp
+++ b/head/core/can_task.cpp
@@ -11,6 +11,7 @@
 #include "common/core/freertos_message_queue.hpp"
 #include "common/core/freertos_task.hpp"
 #include "head/core/tasks.hpp"
+#include "head/core/constants.hpp"
 
 static auto& right_queues = head_tasks::get_right_queues();
 static auto& left_queues = head_tasks::get_left_queues();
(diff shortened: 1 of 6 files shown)


include/gantry/core/constants.hpp
```diff
@@ -0,0 +1,3 @@
+#pragma once
```
sotaoverride, line 1: arent we using python scripts for generating such files with constants now?

head/core/can_task.cpp
```diff
@@ -53,12 +54,12 @@ static auto can_move_group_handler_left =
 
 /** Handler of device info requests. */
 static auto device_info_handler_right =
-    device_info_handler::DeviceInfoHandler(right_queues, 0);
+    device_info_handler::DeviceInfoHandler(right_queues, VERSION);
```
sotaoverride, line 57: could prolly just use one handler to take care of both the heads.

include/gantry/core/constants.hpp
amitlissack: We are using python scripts to generate constants that are shared between C++ and Python.

amitlissack (approved):

sfoster1 (changes requested):
We don't need to go all the way towards tying it into git, but IMO this file should be

1. in `common` somewhere
2. generated by cmake (sorry!)

Laura-Danielle:
@amitlissack, I wonder if it makes sense for us to share a version between python and the firmware though. I could imagine a world where the python side is say v1.2, head/gantry are v1.0 and pipettes is v1.1.

However, if you think in general it's easier to keep those versions tracked on the python side I can do that.

head/core/can_task.cpp
Laura-Danielle: I think we can address this in a follow-up PR.

PR #208, opened 2022-01-11
fix(spi): motor driver bug fix

amitlissack:


include/motor-control/core/motor_driver.hpp
```diff
@@ -29,12 +29,18 @@ class MotorDriver {
     }
 
     auto read(DriverRegisters::Addresses motor_reg, uint32_t command_data)
```
sfoster1, line 31: If you make this `[[nodiscard]]` it should prevent problems like this in the future

sfoster1 (approved):
Looks good!

Laura-Danielle (approved):

include/motor-control/core/motor_driver.hpp
```diff
@@ -29,12 +29,18 @@ class MotorDriver {
...
         auto txBuffer = spi::TMC2130Spi::BufferType{};
         spi::TMC2130Spi::build_command(txBuffer, spi::TMC2130Spi::Mode::READ,
                                        motor_reg, command_data);
         spi_comms.transmit_receive(txBuffer, rxBuffer);
-        return txBuffer;
+
+        // Extract data bytes after the address.
+        uint32_t response = 0;
+        const auto* iter = rxBuffer.cbegin();                      // NOLINT
+        iter = bit_utils::bytes_to_int(iter + 1, rxBuffer.cend(),  // NOLINT
```
fsinapi, line 41: Just a heads up that this isn't going to return the value of the register that you just requested a read from, this is going to be the return value from whatever the previous transaction was.

amitlissack: You don't say. That's interesting. So this never even worked a little.

PR #209, opened 2022-01-11
fix(spi): read requires two transmissions.

amitlissack:
As @fsinapi pointed out a read command to the TMC2130 will return the data in the register specified in the previous command.

This PR simulates that behavior in the simulator and fixes the driver.

### DIFF
--- a/common/simulation/spi.cpp
+++ b/common/simulation/spi.cpp
@@ -19,6 +19,10 @@ void sim_spi::SimTMC2130Spi::transmit_receive(
     constexpr uint8_t write_mask =
         static_cast<uint8_t>(spi::TMC2130Spi::Mode::WRITE);
 
+    auto out_iter = receive.begin();
+    // Write status byte into buffer.
+    out_iter = bit_utils::int_to_bytes(status, out_iter, receive.end());
+
     uint8_t reg = control & ~write_mask;
 
     if (control & write_mask) {
@@ -26,12 +30,9 @@ void sim_spi::SimTMC2130Spi::transmit_receive(
         register_map[reg] = data;
     } else {
         // A read command will return the data from the previous read command.
-        auto out_iter = receive.begin();
-        out_iter =
-            bit_utils::int_to_bytes(read_register, out_iter, receive.end());
         out_iter = bit_utils::int_to_bytes(register_map[read_register],
                                            out_iter, receive.end());
-        // For the next read.
-        read_register = reg;
     }
+    // The register is cached for the next read operation.
+    read_register = reg;
 }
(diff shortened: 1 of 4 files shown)


sfoster1 (approved):

common/simulation/spi.cpp
```diff
@@ -22,11 +22,16 @@ void sim_spi::SimTMC2130Spi::transmit_receive(
...
         register_map[reg] = data;
+    } else {
+        // A read command will return the data from the previous read command.
+        auto out_iter = receive.begin();
+        out_iter =
+            bit_utils::int_to_bytes(read_register, out_iter, receive.end());
+        out_iter = bit_utils::int_to_bytes(register_map[read_register],
+                                           out_iter, receive.end());
+        // For the next read.
+        read_register = reg;
```
Laura-Danielle, line 35: Can you explain what the purpose of modifying the read register is here?

Laura-Danielle (approved):
I had one question, but otherwise it looks good.

common/simulation/spi.cpp
```diff
@@ -22,11 +22,16 @@ void sim_spi::SimTMC2130Spi::transmit_receive(
     uint8_t reg = control & ~write_mask;
 
     if (control & write_mask) {
+        // This is a write command.
         register_map[reg] = data;
+    } else {
+        // A read command will return the data from the previous read command.
+        auto out_iter = receive.begin();
+        out_iter =
```
fsinapi, line 30: It looks like this is returning the address of the register as the first byte, according to the datasheet the first byte of the packet is actually a status byte with some flags. I don't think it'll affect any of the tests though.

common/simulation/spi.cpp
amitlissack: We use `read_register` for the read on line 32 rather than `reg`. `read_register` caches the current request's register for the next read request.

common/simulation/spi.cpp
```diff
@@ -22,11 +22,16 @@ void sim_spi::SimTMC2130Spi::transmit_receive(
...
+            bit_utils::int_to_bytes(read_register, out_iter, receive.end());
+        out_iter = bit_utils::int_to_bytes(register_map[read_register],
```
fsinapi, line 32: _Technically_ this should be returned even if the write_mask is enabled. This diagram from the datasheet does a good job of summarizing the behavior.
image

fsinapi (approved):
Couple of nitpicks but looks good! Should work correctly on hardware

common/simulation/spi.cpp
amitlissack: Awesome. Thanks.

PR #210, opened 2022-01-12
fix(head): Device Info's response is from HEAD. Not HEAD_L or HEAD_R

amitlissack:
device info handler is instantiated once for HEAD and not one for L and one for R.

This PR targets the `presense-sensing` branch as that branch has prereqs for this fix.


sotaoverride (approved):
looks good to me, and thanks!

PR #211, opened 2022-01-14
chore(ot3): add targets to build hex and bin files.

amitlissack:


sotaoverride (approved):
the new targets for bin files make sense.

sfoster1 (approved):
Looks good!

PR #212, opened 2022-01-19
feat(python): add fw update message definitions.

amitlissack:
This adds python message definitions to support FW update over CANFD.

There's a preliminary doc on the FW update plan. 

These messages assume that each data message sent will require an acknowledgement from the FW. That is not the ideal as it might be too slow. I'd rather have some flow control from the FW side based on the CAN FIFO level.

But it gets us to where we can download data to FW.

### DIFF
--- a/python/opentrons_ot3_firmware/messages/payloads.py
+++ b/python/opentrons_ot3_firmware/messages/payloads.py
@@ -149,3 +149,54 @@ class ReadPresenceSensingVoltageResponsePayload(utils.BinarySerializable):
     z_motor: utils.UInt16Field
     a_motor: utils.UInt16Field
     gripper: utils.UInt16Field
+
+
+@dataclass
+class FirmwareUpdateWithAddress(utils.BinarySerializable):
+    """A FW update payload with an address."""
+
+    address: utils.UInt32Field
+
+
+class FirmwareUpdateDataField(utils.BinaryFieldBase[bytes]):
+    """The data field of FirmwareUpdateData."""
+    NUM_BYTES = 56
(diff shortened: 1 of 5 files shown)


sotaoverride (approved):

python/opentrons_ot3_firmware/messages/payloads.py
```diff
@@ -149,3 +149,54 @@ class ReadPresenceSensingVoltageResponsePayload(utils.BinarySerializable):
...
+    FORMAT = f"{NUM_BYTES}s"
+
+
+@dataclass
+class FirmwareUpdateData(FirmwareUpdateWithAddress):
+    """A FW update data payload."""
+
+    num_bytes: utils.UInt8Field
+    reserved: utils.UInt8Field
+    data: FirmwareUpdateDataField
```
amitlissack, line 173: Data is a fixed 56 bytes. `num_bytes` indicates how many of the 56 are to be used on the off chance that not all of them are used.

sfoster1 (approved):
I think we need a mechanism for the remote to signal that it is now in firmware update mode. This could be a different device ID in a get-device-info response or something, or we could have the initiate use it: 
- when the application gets an initiate, it responds with some ack code that's like "ok rebooting"
- central keeps sending initiate until
- when the bootloader is up and gets an initiate, it responds with some ack code that's like "ok, ready"

I think everything else here is good for the stream-to-flash scheme; we won't have a cancel because in-progress updates are destructive anyway, we don't need a reboot confirm because in-progress updates are destructive, it's all good.

PR #213, opened 2022-01-19
chore(bootloader): bootloader subproject.

amitlissack:
Total copy and paste job from other subprojects.

Only modifications was to disable CAN and TIM7 interrupts so the FW would build.

What this PR doesn't address is how the bootloader knows which FW it's installed into. If we can safely get around building `bootloader-gantry-x`, `bootloader-head` (etc) binaries that would be nice.


bootloader/firmware/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+        ${CMAKE_CURRENT_SOURCE_DIR}/clocking.c
+        ${CMAKE_CURRENT_SOURCE_DIR}/can.c
+        ${COMMON_EXECUTABLE_DIR}/errors/errors.c)
+
+add_executable(bootloader
+        ${BOOTLOADER_FW_LINTABLE_SRCS}
+        ${BOOTLOADER_FW_NON_LINTABLE_SRCS})
+
+target_link_libraries(bootloader
+        PUBLIC STM32G491RETx
```
amitlissack, line 25: I believe that this is incorrect. `STM32G491RETx` is the package for our application subsystems. 

The bootloader needs its own set of files, right?

CMakePresets.json
```diff
@@ -118,6 +118,14 @@
       "configurePreset": "host-gcc10",
       "jobs": 4,
       "targets": ["build-and-test"]
+    },
+    {
+      "name": "bootloader",
```
sfoster1, line 123: this is unfortunately going to need an entry for the version that goes on the different pipette micro

bootloader/firmware/CMakeLists.txt
sfoster1: Yeah. It will need a different linker script and maybe a different startup assembly

bootloader/firmware/stm32g4xx_hal_conf.h
```diff
@@ -0,0 +1,383 @@
...
+#endif
+
+/* Exported types ------------------------------------------------------------*/
+/* Exported constants --------------------------------------------------------*/
+
+/* ########################## Module Selection ############################## */
+/**
+  * @brief This is the list of modules to be used in the HAL driver
+  */
+#define HAL_MODULE_ENABLED
```
sfoster1, line 37: honestly we probably want to turn as many of these off as we can - we're going to care a lot more about binary size. doesn't have to be in this pr though

sfoster1 (changes requested):
Your intuition is correct that we'll need a different system library (the one with the linker script) for the bootloader; however we end up arranging the software, the bootloader will live in a different place than the application firmware, and the linker script at least will be very different. Although I suspect there's some deeply obscure linker flag that will let us change that or something... to me it would be a lot clearer if we just had a different script.

In terms of making different bootloader binaries per system - IMO the only thing that would let us categorically avoid this is if each board had a resistor strap set on some GPIOs that was different between boards, like some set of 3 gpios with pull ups or pull downs that give us 8 possibilities to determine at runtime which board it is. Otherwise, we may not need a different _codebase_ per board, but we'll need different _targets_ that create different executables with some cmake-set flag different to set the canbus addresses.

Finally, there's one last extremely annoying wrinkle: the pipette micro is different. So regardless of how we handle setting the canbus address, we'll at least need a second executable for the pipette micro that uses a second variant of the pipette linker script.

We don't need to do all the above in this PR by any means, but I think at least the pipette targets should exist here.

I also wonder if maybe it's a better idea to, rather than having a specific cmake configure preset for the bootloader, just sort of piggyback on the cmake configure presets for the application firmwares and add the targets there. This might be really annoying to do in cmake, but eventually we're presumably going to have some targets that smush together the bootloader and application binary firmwares into one thing that a debugger can flash, so they'll need to live in the same configuration space somewhere.

amitlissack:
> Your intuition is correct that we'll need a different system library (the one with the linker script) for the bootloader; however we end up arranging the software, the bootloader will live in a different place than the application firmware, and the linker script at least will be very different. Although I suspect there's some deeply obscure linker flag that will let us change that or something... to me it would be a lot clearer if we just had a different script.
> 
> In terms of making different bootloader binaries per system - IMO the only thing that would let us categorically avoid this is if each board had a resistor strap set on some GPIOs that was different between boards, like some set of 3 gpios with pull ups or pull downs that give us 8 possibilities to determine at runtime which board it is. Otherwise, we may not need a different _codebase_ per board, but we'll need different _targets_ that create different executables with some cmake-set flag different to set the canbus addresses.
> 
> Finally, there's one last extremely annoying wrinkle: the pipette micro is different. So regardless of how we handle setting the canbus address, we'll at least need a second executable for the pipette micro that uses a second variant of the pipette linker script.
> 
> We don't need to do all the above in this PR by any means, but I think at least the pipette targets should exist here.
> 
> I also wonder if maybe it's a better idea to, rather than having a specific cmake configure preset for the bootloader, just sort of piggyback on the cmake configure presets for the application firmwares and add the targets there. This might be really annoying to do in cmake, but eventually we're presumably going to have some targets that smush together the bootloader and application binary firmwares into one thing that a debugger can flash, so they'll need to live in the same configuration space somewhere.

Great feedback.

My goal with this project is to get all the non-bootloader stuff done and testable first (CAN bringup, CAN talk, FLASH write, simulation, integration tests......). The binary placement(s), entering bootloader, starting app can happen once we're confident that this all works.

bootloader/tests/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+        PUBLIC
+        -Wall
+        -Werror
+        -Weffc++
+        -Wreorder
+        -Wsign-promo
+        -Wextra-semi
+        -Wctor-dtor-privacy
+        -fno-rtti)
+target_link_libraries(bootloader Catch2::Catch2)
```
sotaoverride, line 29: should this maybe link common-simulation as well for any future tests?

amitlissack:
> > In terms of making different bootloader binaries per system - IMO the only thing that would let us categorically avoid this is if each board had a resistor strap set on some GPIOs that was different between boards, like some set of 3 gpios with pull ups or pull downs that give us 8 possibilities to determine at runtime which board it is. Otherwise, we may not need a different _codebase_ per board, but we'll need different _targets_ that create different executables with some cmake-set flag different to set the canbus addresses.
> 
> As of now the controller boards themselves don't have ID straps. I'd recommend a single codebase with multiple targets to determine CAN addresses and IO configuration, since any scheme for board ID straps inevitably runs out of unique values for different versions and can potentially have assembly errors that cause the incorrect code to execute. Programming the boards with separate binaries in production is no trouble.
> 
> > Finally, there's one last extremely annoying wrinkle: the pipette micro is different. So regardless of how we handle setting the canbus address, we'll at least need a second executable for the pipette micro that uses a second variant of the pipette linker script.
> 
> The plan is still to _eventually_ change the pipettes back to the same microcontroller used on the other boards, but it's very unclear when that will happen due to the chip shortage. It would be best to support both microcontrollers if you want a pipette bootloader during development.

I wasn't optimistic about not having to build separate targets. You and Seth have convinced me that it's fine.

bootloader/tests/CMakeLists.txt
amitlissack: Good point. Maybe if I'll wait until we're sure we need that library.

bootloader/firmware/stm32g4xx_hal_conf.h
amitlissack: Absolutely. I have no idea if there are cascading requirements from needing FDCAN.

sfoster1 (changes requested):
I'm pretty sure we won't get correct output from building the -pipettes targets while we're in the `cross` configuration, or the -gantry targets while we're in the `cross-pipettes` configuration, because we'll be using the wrong cmake toolchain and setting the wrong architecture flags. I think we need to only add the platform directory that makes sense for the current cmake configurration.

PR #216, opened 2022-01-24
feat(bootloader): can bringup

amitlissack:
Bring up of the CAN connection. 

*Not tested on HW yet due to lockdown*, but no one is relying on this yet so it can be debugged when I can get back to the office safely.


sfoster1 (changes requested):
Ugh I should have seen this in the last PR, sorry, but IMO we want to reserve a node id bit or some other bit somewhere for whether a node is in bootloader state or application state, that's in the get device info response somewhere. That way it's easy to tell from the get device info response whether this is a bootloader or not while keeping the device info generic, and I think it's just good practice to consider those separate planes of control. 

We could make this part of the node ID; we could also change our arbitration ID slicing to take one of the bits that used to be the group code or something and make it the bootloader bit.

Other than that, does look good.

amitlissack:
> Ugh I should have seen this in the last PR, sorry, but IMO we want to reserve a node id bit or some other bit somewhere for whether a node is in bootloader state or application state, that's in the get device info response somewhere. That way it's easy to tell from the get device info response whether this is a bootloader or not while keeping the device info generic, and I think it's just good practice to consider those separate planes of control.
> 
> We could make this part of the node ID; we could also change our arbitration ID slicing to take one of the bits that used to be the group code or something and make it the bootloader bit.
> 
> Other than that, does look good.

Sounds wise.

python/opentrons_ot3_firmware/constants.py
```diff
@@ -13,8 +14,13 @@ class NodeId(int, Enum):
     head = 0x50
     head_l = 0x51
     head_r = 0x52
+    pipette_bootloader = pipette | 0xf
+    gantry_x_bootloader = gantry_x | 0xf
+    gantry_y_bootloader = gantry_y | 0xf
+    head_bootloader = head | 0xf
 
 
+@unique
```
sfoster1, line 23: didn't know about this, that's cool

sfoster1 (approved):
Nice! Looks good

PR #217, opened 2022-01-25
feat(bootloader): bootloader simulator.

amitlissack:
This is a barebones bootloader simulator. 

It creates a simulating can bus and routes messages to and from `message_handler` (which doesn't do anything yet).

The purpose of this application is to exercise the bootloaders business logic in integration tests.


Laura-Danielle (approved):
Looks good to me

PR #218, opened 2022-01-26
feat(python): FirmwareUpdateData payload helper

amitlissack:
Add FirmwareUpdateData.create method that computes checksum and makes a FirmwareUpdateData instance.

### DIFF
--- a/python/opentrons_ot3_firmware/constants.py
+++ b/python/opentrons_ot3_firmware/constants.py
@@ -14,10 +14,10 @@ class NodeId(int, Enum):
     head = 0x50
     head_l = 0x51
     head_r = 0x52
-    pipette_bootloader = pipette | 0xf
-    gantry_x_bootloader = gantry_x | 0xf
-    gantry_y_bootloader = gantry_y | 0xf
-    head_bootloader = head | 0xf
+    pipette_bootloader = pipette | 0xF
+    gantry_x_bootloader = gantry_x | 0xF
+    gantry_y_bootloader = gantry_y | 0xF
+    head_bootloader = head | 0xF
 
 
 @unique
--- a/python/opentrons_ot3_firmware/messages/payloads.py
+++ b/python/opentrons_ot3_firmware/messages/payloads.py
@@ -1,4 +1,7 @@
 """Payloads of can bus messages."""
+# TODO (amit, 2022-01-26): Figure out why using annotations import ruins
+#  dataclass fields interpretation.
+#  from __future__ import annotations
 from dataclasses import dataclass
 
 from .. import utils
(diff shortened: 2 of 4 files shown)


python/opentrons_ot3_firmware/constants.py
```diff
@@ -14,10 +14,10 @@ class NodeId(int, Enum):
...
     head_l = 0x51
     head_r = 0x52
-    pipette_bootloader = pipette | 0xf
-    gantry_x_bootloader = gantry_x | 0xf
-    gantry_y_bootloader = gantry_y | 0xf
-    head_bootloader = head | 0xf
+    pipette_bootloader = pipette | 0xF
+    gantry_x_bootloader = gantry_x | 0xF
+    gantry_y_bootloader = gantry_y | 0xF
+    head_bootloader = head | 0xF
```
DerekMaggio, line 20: Does this change anything by capitalizing it?

python/opentrons_ot3_firmware/messages/payloads.py
```diff
@@ -183,6 +186,21 @@ def __post_init__(self) -> None:
...
+        """Create a firmware update data payload."""
+        checksum = 0
+        obj = FirmwareUpdateData(
+            address=utils.UInt32Field(address),
+            num_bytes=utils.UInt8Field(len(data)),
+            reserved=utils.UInt8Field(0),
+            data=FirmwareUpdateDataField(data),
+            checksum=utils.UInt16Field(checksum),
+        )
+        checksum = (1 + ~sum(obj.serialize())) & 0xFFFF
```
DerekMaggio, line 200: Why do you bitwise and this?

DerekMaggio (commented):
Looks fine to me. But I don't understand enough what you are doing with the checksum to approve it

python/opentrons_ot3_firmware/constants.py
sfoster1: nope

python/opentrons_ot3_firmware/messages/payloads.py
sfoster1: it's a little clearer if you're used to reading these, if i don't see it i reflexively go "wait what happens on overflow"

sfoster1 (approved):
Looks good to me!

python/opentrons_ot3_firmware/messages/payloads.py
DerekMaggio: Also would `(0x0001  + ~sum(obj.serialize()))` be equivalent?

python/opentrons_ot3_firmware/constants.py
amitlissack: Nah. Black just reformatted it.

python/opentrons_ot3_firmware/messages/payloads.py
amitlissack: It's a common way to perform a checksum. Add the bytes, NOT them and add 1 (make it into a negative number).

`& 0xFFFF` is to restrict the result to 16 bits.

PR #221, opened 2022-01-28
feat(python): add CAN messages to send limit switch status

caila-marashaj:
adding message definition to get limit switch status via CAN


python/opentrons_ot3_firmware/messages/payloads.py
```diff
@@ -203,3 +203,10 @@ class FirmwareUpdateCompleteAcknowledge(utils.BinarySerializable):
     """A response to the FirmwareUpdateComplete message."""
 
     error_code: utils.UInt16Field
+
+
+@dataclass
+class GetLimitSwitchResponse(utils.BinarySerializable):
+    """A response to the Limit Swith Status request payload"""
+
```
DerekMaggio, line 229: Do we have a definition of the switch status?
I am assuming an Enum that maps to the int?

python/opentrons_ot3_firmware/scripts/generate_header.py
```diff
@@ -84,7 +84,7 @@ def generate_c(output: io.StringIO) -> None:
 
 def write_enum_c(e: Type[Enum], output: io.StringIO) -> None:
     """Generate constants from enumeration."""
-    output.write(f"/** {e.__doc__} */\n")
+    output.write(f"/** {e.__doc__} */\n"
```
DerekMaggio, line 87: Missing a parentheses no?

DerekMaggio (changes requested):
Everything looks good. Except maybe you a missing a paren. Left a comment about it pointing it out.

DerekMaggio (approved):

Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
sfoster1, line 1: This shouldn't be here - happens if you run `pipenv` in the root of the repo accidentally. We should remove this and add `./Pipfile` to `.gitignore`

sfoster1 (approved):
Let's get this pipfile out of there but it looks good to me otherwise. I think we may end up wanting to allow read limit switch response to contain values for multiple switches but then again we already have the patterns of nodes with multiple axes under control having different addresses for each.

Pipfile
amitlissack, line 1: I don't think this file should be included.

amitlissack (approved):
Please just remove that Pipfile. 

Otherwise, congrats on first PR!

python/opentrons_ot3_firmware/messages/message_definitions.py
```diff
@@ -276,11 +276,10 @@ class FirmwareUpdateCompleteAcknowledge:  # noqa: D101
 class ReadLimitSwitchRequest:  # noqa: D101
     payload: payloads.EmptyPayload
     payload_type: Type[BinarySerializable] = payloads.EmptyPayload
-    message_id: MessageId.limit_sw_request
-
+    message_id: Literal[MessageId.limit_sw_request] = MessageId.limit_sw_request
```
amitlissack, line 279: Sorry I missed this in the initial review. 

The significance of the `Literal[MessageId.limit_sw_request]` is to tell our type checker that `message_id` MUST be the value `MessageId.limit_sw_request`.

PR #222, opened 2022-01-31
head: push tool detection notifications

sotaoverride:
Notifications from presence sensing driver task when a change in attached tools is detected.


head/simulator/main.cpp
```diff
@@ -82,8 +82,13 @@ static motor_interrupt_driver::MotorInterruptDriver sim_interrupt_left(
 
 static auto adc_comms = adc::SimADC{};
 
+auto at = presence_sensing_driver::AttachedTool{
```
amitlissack, line 85: Can `AttachedTool` have a constructor that assigns everything to `UNDEFINED_TOOL` instead of duping this initialization?

include/head/core/presence_sensing_driver.hpp
```diff
@@ -26,8 +96,42 @@ class PresenceSensingDriver {
         return voltage_read;
     }
 
+    auto static get_tool(adc::MillivoltsReadings reading) -> AttachedTool {
+        auto at = AttachedTool{.z_motor = can_ids::ToolType::UNDEFINED_TOOL,
+                               .a_motor = can_ids::ToolType::UNDEFINED_TOOL,
+                               .gripper = can_ids::ToolType::UNDEFINED_TOOL};
+        for (auto& element : OT3ToolList) {
+            if ((reading.z_motor < element.detection_upper_bound) &&
```
amitlissack, line 104: If the `Tool` class had a method like `within_bounds` or something then we wouldn't have to duplicate this logic 3 times.

include/head/core/tasks/presence_sensing_driver_task.hpp
```diff
@@ -84,10 +101,13 @@ class PresenceSensingDriverTask {
         auto handler =
             PresenceSensingDriverMessageHandler{*driver, *can_client};
         TaskMessage message{};
+
         for (;;) {
             if (queue.try_read(&message, queue.max_delay)) {
                 handler.handle_message(message);
             }
+            handler.handle_message(
```
amitlissack, line 109: What are we trying to achieve with the push tool notification? Are we wanting to check if a tool is present periodically? As this is written, the check for tools will only happen when a can message asking for voltage arrives. 

When we chatted about this the other day I suggested a timer sending `PushToolsDetectedNotification` to this task.

head/tests/test_presence_sensing_driver.cpp
```diff
@@ -18,4 +23,41 @@ SCENARIO("get readings called on presence sensing driver") {
             }
         }
     }
+}
+
+SCENARIO("get_tool called on presence sensing driver") {
+    GIVEN("PresenceSensingDriver instance") {
+        auto raw_readings =
```
amitlissack, line 30: This test only tests one case. It doesn't test for the other tools or for there being no tools.

python/opentrons_ot3_firmware/constants.py
```diff
@@ -94,3 +96,15 @@ class ErrorCode(int, Enum):
     bad_checksum = 0x02
     invalid_byte_count = 0x03
     invalid_input = 0x04
+
+
+@unique
+class ToolType(int, Enum):
+    """Tool types detected on Head."""
+
+    UNDEFINED_TOOL = (0x00,)
```
amitlissack, line 105: This is not correct. This creates an enum of tuple values.

Also, notice that  the case is inconsistent with the other enums in this file.

amitlissack (changes requested):

python/opentrons_ot3_firmware/messages/payloads.py
```diff
@@ -154,6 +154,16 @@ class ReadPresenceSensingVoltageResponsePayload(utils.BinarySerializable):
     gripper: utils.UInt16Field
 
 
+@dataclass
+class ToolsDetectedNotificationPayload(utils.BinarySerializable):
+    """A tool detection notification for tools attached to z axis, a axis and gripper."""
+
+    # Tools are mapped to an enum
+    z_motor: utils.UInt8Field
```
sfoster1, line 162: These are `uint8` and the c++ side is `uint16` i believe. we should also ideally just make these `enum` elements with custom ser/des.

include/can/core/messages.hpp
```diff
@@ -336,13 +336,32 @@ struct ReadPresenceSensingVoltageResponse
         -> bool = default;
 };
 
+struct PushToolsDetectedNotification
+    : BaseMessage<MessageId::tools_detected_notification> {
+    uint16_t z_motor;
```
sfoster1, line 341: The python side of things appears to think these are `uint8`. We should fix that but also IMO these should just be `enum` elements, a new specialization of `int_to_bytes`/`bytes_to_int` can handle that just fine

include/head/core/tasks/presence_sensing_driver_task.hpp
sfoster1: Agree, and also it's not a good idea IMO to reuse messages like this. `can_messages::PushToolsDetectedNotification` is being used as 
a) something that the head sends via the canbus to the host to notify it of a change in the currently-attached tools
b) a head-internal message that's used as a trigger to actually do the check

That second one shouldn't be a canbus message. Also like Amit says we should use a timer or something for this.

python/opentrons_ot3_firmware/constants.py
```diff
@@ -94,3 +96,15 @@ class ErrorCode(int, Enum):
     bad_checksum = 0x02
     invalid_byte_count = 0x03
     invalid_input = 0x04
+
+
+@unique
+class ToolType(int, Enum):
```
sfoster1, line 102: How do we express that no tool is connected on a given carrier? We should have both the concept of "nothing is attached" and "something I don't know about is attached".

include/head/core/presence_sensing_driver.hpp
```diff
@@ -26,8 +96,42 @@ class PresenceSensingDriver {
         return voltage_read;
     }
 
+    auto static get_tool(adc::MillivoltsReadings reading) -> AttachedTool {
```
sfoster1, line 99: this method doesn't modify any state and can be `const` i believe. Although I will also say, this method really just appears to be a constructor of the `AttachedTool` class -  maybe that's just what it should be.

include/head/core/presence_sensing_driver.hpp
```diff
@@ -26,8 +96,42 @@ class PresenceSensingDriver {
...
+                (reading.gripper >= element.detection_lower_bound) &&
+                (element.tool_carrier == UNDEFINED_CARRIER)) {
+                at.gripper = element.tool_type;
+            }
+        }
+
+        return at;
+    }
+
+    auto getCurrentTools() -> AttachedTool { return this->current_tools; }
```
sfoster1, line 124: these are `camelCase` but `get_tool` and `get_readings` are `snake_case`. let's just do `snake_case` since that's what the rest of hte project uses.

sfoster1 (changes requested):
Inline changes.

Can we also add a message that the host sends to the head to query the current status? WIth push-only there's a race - if the head initializes before the host, the host would never get the initial message that talks about the attached tools at boot.

We also definitely need tests for more cases.

include/head/core/presence_sensing_driver.hpp
```diff
@@ -1,13 +1,83 @@
...
+
+#include "can/core/ids.hpp"
 #include "common/core/bit_utils.hpp"
 #include "head/core/adc.hpp"
 
 namespace presence_sensing_driver {
 
+using namespace can_ids;
+
+constexpr uint16_t PIPETTE_384_CHAN_Z_CARRIER_DETECTION_UPPER_BOUND = 1883;
```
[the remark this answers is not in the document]
sotaoverride: have added a header for dealing with the style of things above. have placed it under include/head/core for now thinking it'll be head specific

include/head/core/presence_sensing_driver.hpp
```diff
@@ -1,13 +1,83 @@
...
+constexpr uint16_t PIPETTE_SINGLE_A_CARRIER_DETECTION_LOWER_BOUND = 2362;
+
+constexpr uint16_t PIPETTE_MULTI_A_CARRIER_DETECTION_UPPER_BOUND = 2860;
+constexpr uint16_t PIPETTE_MULTI_A_CARRIER_DETECTION_LOWER_BOUND = 2844;
+
+constexpr uint16_t GRIPPER_DETECTION_UPPER_BOUND = 999;
+constexpr uint16_t GRIPPER_DETECTION_LOWER_BOUND = 536;
+
+enum Carrier : uint8_t {
+    UNDEFINED_CARRIER = 0x00,
```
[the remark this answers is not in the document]
sotaoverride: I feel like going into this I was confused what carrier the gripper goes on, or is it not carrier dependent / shares the carrier with one of the pipettes. Think this confusion may stem from how I don't see the voltage bounds as such for the gripper yet.

python/opentrons_ot3_firmware/constants.py
sotaoverride: threw in these other two values:
```
@unique
class ToolType(int, Enum):
    """Tool types detected on Head."""

    undefined_tool = 0x00
    pipette_96_chan = 0x01
    pipette_384_chan = 0x02
    pipette_single_chan = 0x03
    pipette_multi_chan = 0x04
    gripper = 0x05
    nothing_attached = 0x06
```

include/head/core/presence_sensing_driver.hpp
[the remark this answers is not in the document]
sfoster1: I think that's a good idea though it will probably become `common` since it turns out the pipette needs it too, but that can be a later pr

include/head/core/tasks/presence_sensing_driver_task.hpp
sotaoverride: I've tried putting in something to get this bit taken care of using the FreeRTOSTimer. Trying to figure out a way of writing a test for it

head/tests/CMakeLists.txt
```diff
@@ -28,6 +29,9 @@ target_compile_options(head
         -Wextra-semi
         -Wctor-dtor-privacy
         -fno-rtti)
+
+target_head_core_tool_list(head)
```
sfoster1, line 33: Do we need this? can't we use `target_head_core`?

head/firmware/main.cpp
```diff
@@ -185,6 +191,15 @@ auto main() -> int {
     head_tasks::start_tasks(can_bus_1, motor_left.motion_controller,
                             motor_left.driver, motor_right.motion_controller,
                             motor_right.driver, psd);
+    static auto timer_for_notifier =
```
sfoster1, line 192: having this be an in-function `static` will prevent it from being overwritten by freertos, but why have it be an in-function `static` as opposed to just be defined outside a function at all.

head/firmware/main.cpp
```diff
@@ -165,7 +169,9 @@ adc::ADC_interface ADC_intf1 = {
 
 static auto ADC_comms = adc::ADC(ADC_intf1, ADC_intf2);
 
-static auto psd = presence_sensing_driver::PresenceSensingDriver{ADC_comms};
+auto at = ot3_tool_list::AttachedTool{};
```
sfoster1, line 170: [nit] this is passed by-value to the `PresenceSensingDriver` so why not do it inline: `static auto psd = presence_sensing_driver::PresenceSensingDriver{ADC_comms, ot3_tool_list::AttachedTool{}};`

include/head/core/tool_list.hpp
```diff
@@ -0,0 +1,93 @@
...
+constexpr auto pipette_multiple_chan_z_bounds =
+    ToolCheckBounds{.upper = 945, .lower = 918};
+
+constexpr auto pipette_single_chan_a_bounds =
+    ToolCheckBounds{.upper = 2389, .lower = 2362};
+
+constexpr auto pipette_multiple_chan_a_bounds =
+    ToolCheckBounds{.upper = 2860, .lower = 2844};
+
+constexpr auto nothing_conected_z_bounds =
```
sfoster1, line 34:
```suggestion
constexpr auto nothing_connected_z_bounds =
```

include/head/core/tool_list.hpp
```diff
@@ -0,0 +1,93 @@
...
+    ToolCheckBounds{.upper = 3, .lower = 1};
+
+constexpr auto nothing_conected_a_bounds =
```
sfoster1, line 37:
```suggestion
constexpr auto nothing_connected_a_bounds =
```

include/head/core/tool_list.hpp
```diff
@@ -0,0 +1,93 @@
...
+    ToolCheckBounds{.upper = 54, .lower = 16};
+
+// revisit these, not sure if EE has a calculation for gripper carrier bounds
+constexpr auto nothing_conected_gripper_bounds =
```
sfoster1, line 41:
```suggestion
constexpr auto nothing_connected_gripper_bounds =
```

head/tests/test_presence_sensing_driver.cpp
```diff
@@ -18,4 +20,81 @@ SCENARIO("get readings called on presence sensing driver") {
...
+    }
+}
+
+SCENARIO("ot3 tools list for tool detection") {
+    GIVEN("ot3 tools lower and upper voltage bounds for tool detection") {
+        auto tools = ot3_tool_list::get_tool_list();
+        WHEN("when lower bound compared with upper bound") {
+            THEN("lower bound is always found lower or equal to upper bounds") {
+                for (auto& element : tools) {
+                    REQUIRE(element.bounds.lower <= element.bounds.upper);
```
sfoster1, line 95: nice, good idea!

sfoster1 (commented):
A couple typos, and we should test this on real hardware, but this looks really good now! Will approve once tested on hardware and the typos are fixed

head/tests/test_presence_sensing_driver.cpp
sotaoverride: was @amitlissack's idea. Thanks @amitlissack

head/core/tool_list.cpp
```diff
@@ -0,0 +1,39 @@
+
+
+#include "head/core/tool_list.hpp"
+using namespace ot3_tool_list;
+
+std::array<Tool, 10> _list{
```
amitlissack, line 6: Should this be `static`?

head/simulator/main.cpp
```diff
@@ -82,8 +82,10 @@ static motor_interrupt_driver::MotorInterruptDriver sim_interrupt_left(
 
 static auto adc_comms = adc::SimADC{};
 
+auto at = ot3_tool_list::AttachedTool{};
```
amitlissack, line 85: `static`? And `at` as a name is confusing to me.

head/tests/test_presence_sensing_driver.cpp
amitlissack: Hmmm. That wasn't really my idea. I was suggesting a concept that would require a Tool to have lower be less than upper.

This works too! SO it was all your idea, @sotaoverride !

head/core/tool_list.cpp
```diff
@@ -0,0 +1,39 @@
...
+     {.tool_type = can_ids::ToolType::nothing_attached,
+      .tool_carrier = Z_CARRIER,
+      .bounds = nothing_conected_z_bounds},
+     {.tool_type = can_ids::ToolType::nothing_attached,
+      .tool_carrier = A_CARRIER,
+      .bounds = nothing_conected_a_bounds},
+     {.tool_type = can_ids::ToolType::nothing_attached,
+      .tool_carrier = GRIPPER_CARRIER,
+      .bounds = nothing_conected_gripper_bounds}}};
+auto ot3_tool_list::get_tool_list() -> const std::array<Tool, 10>& {
```
amitlissack, line 37: `std::array<Tool, 10>`  is a type that is used in various places. 

Could we define an alias? 

Or perhaps use a `std::span` (not sure that is sensible) or a sensible type that doesn't require the size?

head/firmware/main.cpp
```diff
@@ -185,6 +189,15 @@ auto main() -> int {
     head_tasks::start_tasks(can_bus_1, motor_left.motion_controller,
                             motor_left.driver, motor_right.motion_controller,
                             motor_right.driver, psd);
+    static auto timer_for_notifier =
+        freertos_timer::FreeRTOSTimer<pdMS_TO_TICKS(100)>(
+            "timer for notifier",
```
amitlissack, line 194: I don't see that there's an equivalent timer in the simulator. 

There's no way you want to have a timer at 100ms in the simulator. But some way to exercise this would be nice.

Although it is completely sensible to move that to a different PR.

amitlissack (commented):
Thank you for listening to all the feedback. Looking good!

I agree that I'll approve once it's tested.

sfoster1 (changes requested):
Almost there! There's that one typo, and I agree with @amitlissack that we should `typedef` or `using` the array template specialization from the tool list. I also think we could make the constexpr values that populate the tool list live in a .cpp rather than a header, but that might make it annoying to test.

head/tests/CMakeLists.txt
sotaoverride: I was trying to keep FreeRTOS out of tests at the time I think - 
```
In file included from /Users/aatirmanzur/code/ot3-firmware/head/tests/../../include/can/core/message_writer.hpp:8,
                 from /Users/aatirmanzur/code/ot3-firmware/head/tests/../../include/head/core/tasks.hpp:4,
                 from /Users/aatirmanzur/code/ot3-firmware/head/core/tasks.cpp:1:
/Users/aatirmanzur/code/ot3-firmware/head/tests/../../include/common/core/freertos_message_queue.hpp:5:10: fatal error: FreeRTOS.h: No such file or directory
    5 | #include "FreeRTOS.h"
      |          ^~~~~~~~~~~~
compilation terminated.
```

sfoster1 (approved):
Nice! Looks great!

amitlissack (approved):

PR #227, opened 2022-02-01
fix(codegen): "Namespace" C enums.

amitlissack:
Prefix generated C enums with CAN to avoid clashes with C++. More consistent with C-style.

This is an issue for bootloader simulator as it uses both C++ and C generated types.


sfoster1 (approved):
oh, C

PR #228, opened 2022-02-01
feat(motor-control): Added GPIO addressing for all boards in motor control interrupt routine

caila-marashaj:
[RET-416]

[RET-416]: https://opentrons.atlassian.net/browse/RET-416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/head/firmware/za_board_gpio.c
+++ b/head/firmware/za_board_gpio.c
@@ -1,0 +1,68 @@
+#include "head/core/za_gpio_constants.h"
+
+/**
+ * @brief LED Driver GPIO Initialization Function
+ * @param None
+ * @retval None
+ */
+
+void LED_drive_gpio_init() {
+    /* GPIO Ports Clock Enable */
+    __HAL_RCC_GPIOB_CLK_ENABLE();
+
+    /*Configure GPIO pin : PB6 */
+    GPIO_InitTypeDef GPIO_InitStruct = {0};
+    GPIO_InitStruct.Pin = GPIO_PIN_6;
+    GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
+    GPIO_InitStruct.Pull = GPIO_NOPULL;
+    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
+    HAL_GPIO_Init(GPIOB, &GPIO_InitStruct);
+}
+
+/**
+ * @brief Limit Switch Z Initialization Function
+ * @param None
+ * @retval None
+ */
+void limit_switch_gpio_init() {
+    /* GPIO Ports Clock Enable */
+    __HAL_RCC_GPIOB_CLK_ENABLE();
+
+    /*Configure GPIO pin : PB7:z and PB9:a*/
+    GPIO_InitTypeDef GPIO_InitStruct = {0};
+    GPIO_InitStruct.Pin = GPIO_PIN_7 | GPIO_PIN_9;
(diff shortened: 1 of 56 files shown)


head/firmware/za_board_gpio.c
```diff
@@ -0,0 +1,68 @@
...
+
+void turn_on_LED_pin() {
+    HAL_GPIO_WritePin(GPIOB, GPIO_PIN_6, GPIO_PIN_SET);
+}
+
+void turn_off_LED_pin() {
+    HAL_GPIO_WritePin(GPIOB, GPIO_PIN_6, GPIO_PIN_RESET);
+}
+
+void check_limit_switch(MountType mount) {
```
sfoster1, line 47: I think this probably should be a different function for each switch. Since the head works by having one independent software stack for each side, each side will want a different function to call.

Also, we should probably make this return the value.

sfoster1 (changes requested):
One big thing and a couple small things.

Small things: this file is already identified as relating to the z/a board because it's in the `head` project. i don't think we need `za` in the names of the files or functions

Big thing: This does look like it will work, but it's not testable. We can't automatically test it because it's all in firmware and can't be host-compiled, and we can't test it on a board because it's not hooked up to any calling code. It's fine to implement something in firmware first, but we should add some place that calls the code, even if it's not the final location, so we can test that it works on real hardware.

include/head/core/za_gpio_constants.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
```
sotaoverride, line 1: something like this file should probably go under include/head/firmware

head/firmware/za_board_gpio.c
caila-marashaj: Should I initialize them in separate functions as well? Or only perform the check separately?

caila-marashaj:
> One big thing and a couple small things.
> 
> Small things: this file is already identified as relating to the z/a board because it's in the `head` project. i don't think we need `za` in the names of the files or functions
> 
> Big thing: This does look like it will work, but it's not testable. We can't automatically test it because it's all in firmware and can't be host-compiled, and we can't test it on a board because it's not hooked up to any calling code. It's fine to implement something in firmware first, but we should add some place that calls the code, even if it's not the final location, so we can test that it works on real hardware.

Couple questions: What would be a good name for these files instead? Maybe something like gpio_util? Or would that be too similar to the gpio files in the pipettes project?

What do you mean by calling code? What would I need to reference directly? Is a simpler solution for being able to test this code moving it into another directory, or adding calling code in here?

sfoster1:
> Couple questions: What would be a good name for these files instead? Maybe something like gpio_util? Or would that be too similar to the gpio files in the pipettes project?

`gpio_util` would be fine, and it would be actually good I think for the files to be similar to the ones in pipettes.

> What do you mean by calling code? What would I need to reference directly? Is a simpler solution for being able to test this code moving it into another directory, or adding calling code in here?

Just have some part of the code where these functions are actually called.

sfoster1 (approved):
Looks great to me!

One thing to do before merging though - change the PR description (it shows as the title of the commit in `main`) to something more descriptive of what this PR now is.

Laura-Danielle (approved):
Agree with Seth about the title. Changes look good.

PR #230, opened 2022-02-02
feat(can): Node IDs for left and right pipette

sfoster1:
When there are two pipettes attached, they need to have different node
IDs so they can both talk on the bus. The thing that makes them
different from each other is which mount they're on. We therefore can
use a different node id depending on the mount, which can be sensed from
a voltage expressed by a resistor net split between pipette and head
boards.

This commit only creates the new IDs and switches to using them. The pipette
application firmware and bootloader listen to left, arbitrarily.

The next step is to bring up the ADC to sense the voltage on the
resistor net and store its value so we know who we are.

### DIFF
--- a/include/can/core/message_writer.hpp
+++ b/include/can/core/message_writer.hpp
@@ -21,7 +21,6 @@ class MessageWriter {
 
     explicit MessageWriter(can_ids::NodeId node_id) : node_id(node_id) {}
 
-
     /**
      * Write a message to the can bus
      *
--- a/include/can/core/message_writer.hpp
+++ b/include/can/core/message_writer.hpp
@@ -21,6 +21,7 @@ class MessageWriter {
 
     explicit MessageWriter(can_ids::NodeId node_id) : node_id(node_id) {}
 
+
     /**
      * Write a message to the can bus
      *
@@ -44,9 +45,9 @@ class MessageWriter {
     }
 
     void set_queue(QueueType* q) { queue = q; }
+    can_ids::NodeId node_id;
 
   private:
     QueueType* queue{nullptr};
-    can_ids::NodeId node_id;
 };
 }  // namespace can_message_writer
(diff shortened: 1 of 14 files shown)


pipettes/core/can_task.cpp
```diff
@@ -92,7 +92,8 @@ void callback(void* cb_data, uint32_t identifier, uint8_t* data,
 [[noreturn]] void can_task::CanMessageReaderTask::operator()(
     can_bus::CanBus* can_bus) {
     can_bus->set_incoming_message_callback(nullptr, callback);
-    can_bus->setup_node_id_filter(can_ids::NodeId::pipette);
+    can_bus->setup_node_id_filter(can_ids::NodeId::pipette_left);
```
amitlissack, line 95: A TODO here too?

amitlissack (approved):

Laura-Danielle (approved):
Looks good

DerekMaggio:
Will the simulators be getting split in this PR?

sfoster1:
> Will the simulators be getting split in this PR?

Simulators (and firmwares) will not be split. One firmware will determine which mount it's on when it starts, and the simulators I suppose will need a parameter. I'll add it to the simulator when adding it to the firmware.

pipettes/core/can_task.cpp
sfoster1: yyesss

include/can/core/message_writer.hpp
```diff
@@ -44,9 +44,9 @@ class MessageWriter {
     }
 
     void set_queue(QueueType* q) { queue = q; }
+    can_ids::NodeId node_id;
```
amitlissack, line 47: Can we please keep this member private? Having a `set_node_id` makes it clear that nobody is reading this value. Just `set_queue`.

sfoster1: okayyyy

PR #232, opened 2022-02-03
feat(sim): simulators can handle disconnects. they retry connecting.

amitlissack:
A little bit of a user devx upgrade. Handle disconnects and try to reconnect.

Could limit the number of retries in the future. But this is (way) better than it was.


sfoster1 (approved):
Love to sim

PR #236, opened 2022-02-09
feat(pipettes): separate i2c into its own task

Laura-Danielle:
## Overview

Given there will be many different tasks which rely on writing to the i2c bus, we
needed to create the infrastructure where responsibility for reading and writing to the bus was
controlled by one entity.

This PR adds an i2c task in which polls a i2c command queue. This command queue can be written to by any task requiring an i2c bus (such as the eeprom task).

Closes RET 407

## Changelog
- Add i2c task
- Modify eeprom task to read/write to the i2c command queue
- Add i2c task initialization code
- Add tests to test the i2c queue writer, i2c task and eeprom task

## Follow-ups
- Adding tasks for the pressure sensor, capacitive sensor and miscellaneous sensors
- Adding the ability to 'poll' the i2c bus for data (capability is in the capacitive sensor PR)

## Review Requests

Please take a look at the callback function in `eeprom_task.hpp` and let me know if there is any way I can improve it. Also open to changes to tests and other such refactors.


include/pipettes/core/messages.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include <functional>
+
+#include "common/core/i2c.hpp"
+
+namespace pipette_messages {
+
+using Callback = std::function<void(uint8_t *, uint16_t)>;
+
+struct WriteToI2C {
```
DerekMaggio, line 11: Are these messages CAN messages?

pipettes/core/tasks.cpp
```diff
@@ -39,27 +43,36 @@ void pipettes_tasks::start_tasks(
     auto& tasks = pipettes_tasks::get_tasks();
```
DerekMaggio, line 43: Are all these pipette tasks in any particular order?

include/pipettes/core/i2c_writer.hpp
```diff
@@ -0,0 +1,72 @@
...
+class I2CWriter {
+    using QueueType = QueueImpl<TaskMessage>;
+
+  public:
+    I2CWriter() = default;
+
+    template <GenericByteBuffer Buffer, typename Data>
+    requires std::is_integral_v<Data>
+    void write(Buffer& buffer, Data data, uint16_t device_address) {
+        add_to_buffer(buffer, data);
```
amitlissack, line 30: I am a bit concerned with this messaging and memory ownership. `WriteToI2C` holds a pointer to the buffer rather than a copy of the data.  The caller must not deallocate the buffer. 

I think it's a dangerous pattern that could result in mystery crashes.

include/pipettes/core/i2c_writer.hpp
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    template <GenericByteBuffer Buffer>
+    void read(Buffer& buffer, uint16_t device_address,
+              const Callback
+                  callback) {  // NOLINT (performance-unnecessary-value-param)
+        // We want to copy the callback every time as opposed to passing a
+        // reference to it from the eeprom task.
+        uint16_t buff_size = buffer.size();
+        pipette_messages::ReadFromI2C read_msg{.address = device_address,
```
amitlissack, line 45: Same as above.

include/pipettes/core/tasks/i2c_task.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+    void visit(ReadFromI2C &m) {
+        auto callback = std::move(m.client_callback);
+        i2c_device.master_receive(m.buffer, m.size, m.address, TIMEOUT);
+        callback(m.buffer, m.size);
+    }
+    i2c::I2CDeviceBase &i2c_device;
+
+    // Default timeout should be 60 seconds
+    static constexpr auto TIMEOUT = 0xEA60;
```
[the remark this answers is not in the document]
amitlissack: Maybe we should write it in Decimal.

common/firmware/i2c/i2c_comms.cpp
```diff
@@ -13,12 +15,12 @@ using namespace i2c;
  */
 I2C::I2C(HAL_I2C_HANDLE handle) : handle{handle} {}
 
-auto I2C::master_transmit(uint16_t dev_address, uint8_t *data, uint16_t size,
+auto I2C::master_transmit(uint8_t* data, uint16_t size, uint16_t dev_address,
                           uint32_t timeout) -> bool {
     return hal_i2c_master_transmit(handle, dev_address, data, size, timeout);
```
fsinapi, line 20: This is definitely outside the scope of this PR, and possibly unnecessary, but it looks like the task is using the blocking `HAL_I2C_Master_Transmit` function. Moving to the `_IT` versions of the HAL functions and implementing a synchronization system to pause the task during I2C transactions would free up some CPU time for other tasks to run during a transaction. Not sure if this is an optimization you need though.

sfoster1 (changes requested):
Agree with @amitlissack that we shouldn't let the caller own the buffer - especially if these messages are small, let's just pack the data in.

Also especially with the callback I think we can get some safety wins by not using pure pointer arithmetic. we could either use a span or use an iterator pair (rather than an iterator + size).

common/firmware/i2c/i2c_comms.cpp
Laura-Danielle: yep was planning on transitioning to using either the `IT` or DMA versions of I2C transactions in a follow-up

include/pipettes/core/tasks/i2c_task.hpp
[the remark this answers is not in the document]
Laura-Danielle: Yeah can write it in decimal

pipettes/core/tasks.cpp
Laura-Danielle: Not really. Freertos has its own internal scheduling. If we cared about tasks being executed at a certain time, we would need to modify the priorities.

include/pipettes/core/tasks/i2c_task.hpp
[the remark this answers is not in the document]
DerekMaggio: OHHH I see. Thank you :)

include/pipettes/core/messages.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include <functional>
+
+#include "common/core/i2c.hpp"
+
+namespace pipette_messages {
+
+using MaxBufferSize = std::array<uint8_t, 5>;
+using Callback = std::function<void(MaxBufferSize)>;
```
sfoster1, line 10:
```suggestion
using Callback = std::function<void(const MaxBufferSize&)>;
```
though it doesn't really matter at 5 bytes

sfoster1 (approved):
Yup this looks great. Some nits around using refs for buffers rather than copying for _savage optimization_ but that's about it

common/firmware/i2c/i2c_comms.cpp
sfoster1: Doesn't the blocking version actually spin on HAL_Delay which is, i'm pretty sure, rtos-aware

PR #237, opened 2022-02-09
feat(bootloader): writing to flash

amitlissack:
This PR adds code to erase flash and then write to it in response to CAN messages.

It makes the assumption that the bootloader will reside at 0x0800000 and the application will reside at 0x08008000. That means that the first 32kb of flash are for the bootloader.

This is not consistent with how the subsystems are currently configured. Nor is it certain that this is how we'll partition flash.

All that is to say that this code will be useful after future PRs.


bootloader/core/util.c
```diff
@@ -78,3 +78,43 @@ uint16_t compute_checksum(const uint8_t * begin, const uint8_t * end) {
...
+ *
+ * Writes to G4 and L5 flash happen in 64 bit data increments.
+ * @param address The starting flash address
+ * @param buffer A byte buffer
+ * @param length Length in bytes of the buffer.
+ * @param callback A callback that will be called for each 64 bit address
+ * @return True on success
+ */
+bool dword_address_iter(uint32_t address, const uint8_t * buffer, uint8_t length, address_iter_callback callback) {
+    if (!buffer || length <= 0) {
```
DerekMaggio, line 95: What does `!buffer` mean in this context. Is this a conditional like `not variable` in python?
If so are you just verifying that it is not 0?

bootloader/firmware/updater.c
```diff
@@ -0,0 +1,99 @@
+#include "platform_specific_hal_conf.h"
+#include "bootloader/core/updater.h"
+#include "bootloader/core/util.h"
+#include "bootloader/firmware/constants.h"
+
```
DerekMaggio, line 5: Why is this file a .c file instead of a .cpp file? 
Is it because this is a HAL and it needs to be in c?

DerekMaggio (commented):
I left some questions.
For sure not qualified to approve.

bootloader/core/util.c
```diff
@@ -78,3 +78,43 @@ uint16_t compute_checksum(const uint8_t * begin, const uint8_t * end) {
...
+            // Clear word.
+            double_word = 0;
+            // Increment address
+            address += sizeof(uint64_t);
+        }
+        // The number of bytes to shift.
+        int shift_bytes = sizeof(uint64_t) - (double_word_index + 1);
+        int shift_amount = shift_bytes * 8;
+        // Or the shifted byte into the double word
+        double_word |= ((uint64_t)buffer[i] << shift_amount);
```
DerekMaggio, line 115: Is this what is adding the word to the double word? 
If so, what is the bitwise OR doing?

bootloader/core/util.c
DerekMaggio: Is there a reason that you don't use `!=0`?

bootloader/firmware/updater.c
amitlissack: In order to keep the binary small, all the bootloader code is to be written in C.

bootloader/core/util.c
```diff
@@ -78,3 +78,43 @@ uint16_t compute_checksum(const uint8_t * begin, const uint8_t * end) {
...
+ * @param buffer A byte buffer
+ * @param length Length in bytes of the buffer.
+ * @param callback A callback that will be called for each 64 bit address
+ * @return True on success
+ */
+bool dword_address_iter(uint32_t address, const uint8_t * buffer, uint8_t length, address_iter_callback callback) {
+    if (!buffer || length <= 0) {
+        return true;
+    }
+    uint64_t double_word = 0;
```
DerekMaggio, line 98: I was looking up what a double word is and I got a lot of conflicting reports.
Is it just a name for a unsigned 64 bit int? Kind of like longs and shorts?

bootloader/core/util.c
amitlissack: `double_word` is accumulating bytes until a full 8 bytes have been written into it. Shift amount goes from 7 to 0 repeatedly.

bootloader/firmware/updater.c
DerekMaggio: Dumb question. Why does the bootloader need to be small? Is it because it is actually stored on the physical memory of the device and we want to take up as little memory as possible?

bootloader/core/util.c
amitlissack: `word` is platform dependent. In the case  of this processor it is 32 bits. So a double word is 64 bits.

bootloader/core/util.c
sfoster1: idiomatic

bootloader/core/util.c
sfoster1, line 115: i think this is `memcpy`. can we use `memcpy`?

bootloader/firmware/updater.c
sfoster1: correct

amitlissack: The smaller the bootloader, the more space there is for the main applcation.

include/bootloader/firmware/constants.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
```
sfoster1, line 1: It may not be in this pr but you can define these values in the linker script (see the way eg `_estack` is set) and pull them into c files by `extern`-declaring them and single source of truth it

bootloader/firmware/updater.c
```diff
@@ -0,0 +1,99 @@
...
+    /** Running error detection value of update date. */
+    uint32_t error_detection;
+    bool erased;
+} UpdateState;
+
+static UpdateState update_state = {
+    .num_messages_received=0,
+    .error_detection=0,
+    .erased=false
+};
```
sfoster1, line 28: can we make this "c object-oriented" with a `UpdateState* get_state()` and have the other functions take an `UpdateState*`?

sfoster1 (approved):
Looks really good!

include/bootloader/firmware/constants.h
amitlissack: I'll get to that in the future. But good idea.

PR #238, opened 2022-02-09
Ret 485 motion stopping

caila-marashaj:


include/motor-control/core/motion_controller.hpp
```diff
@@ -50,13 +50,17 @@ class MotionController {
                  .velocity = velocity_steps,
                  .acceleration = acceleration_steps,
                  .group_id = can_msg.group_id,
-                 .seq_id = can_msg.seq_id};
+                 .seq_id = can_msg.seq_id,
+                 .stop_condition = static_cast<MoveStopCondition>(
```
Laura-Danielle, line 54: Think you can just pass this into the enum like: `MoveStopCondition(can_msg.request_stop_condition)`

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -163,17 +193,7 @@ class MotorInterruptHandler {
     void finish_current_move() {
```
Laura-Danielle, line 193: we should pass in the limit switch status here and default it to false.

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -163,17 +193,7 @@ class MotorInterruptHandler {
...
-                           .seq_id = buffered_move.seq_id,
-                           .current_position = static_cast<uint32_t>(
-                               position_tracker >>
-                               31),  // TODO (AA 2021-11-10): convert
-                                     // this value to mm instead of steps
-                           .ack_id = AckMessageId::complete};
-            static_cast<void>(
-                status_queue_client.send_move_status_reporter_queue(ack));
-        }
+        send_ack(false, AckMessageId::complete);
```
Laura-Danielle, line 196: Want to use the value passed into `finish_current_move` rather than setting it to false always.

include/motor-control/core/motor_messages.hpp
```diff
@@ -34,6 +37,7 @@ struct Ack {
     uint8_t seq_id;
     uint32_t current_position;
     AckMessageId ack_id;
+    bool lim_sw_triggered;
```
Laura-Danielle, line 40: The ack message gets put into a `MoveCompleted` command which is found in `move_status_reporter_task.hpp`

Laura-Danielle (changes requested):
Somehow you've got changes from your previous PR in here. Also have some change requests.

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -119,6 +124,31 @@ class MotorInterruptHandler {
...
+                status_queue_client.send_move_status_reporter_queue(ack));
+        }
+    }
+    auto limit_switch_triggered() -> bool {
+        if (hardware.check_limit_switch() && can_step()) {
+            send_ack(true, AckMessageId::error);
+            return true;
+        }
+        if (hardware.check_limit_switch() && !can_step()) {
+            send_ack(true, AckMessageId::complete);
```
Laura-Danielle, line 147: Now that I think about it, we want to have the move be "finished" regardless of what happens in the limit switch.


We should be calling `finish_move` either here or higher up in the `pulse` function

sfoster1 (changes requested):
I think we want the normal stop condition to ignore the limit switches. That way we don't have to retract and can tolerate being stopped with the limit switches triggered.

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -104,18 +104,37 @@ class MotorInterruptHandler {
...
                 }
+                return false;
             }
-            return false;
+        }
+        return false;
+    }
+
+    auto limit_switch_triggered() -> bool {
+        if (hardware.check_limit_switch() && can_step()) {
```
sfoster1, line 131: This functionality means that
- If the limit switch is triggered on the last _tick_ of the move, there's no error
- If the limit switch is triggered on any other tick, there's an error

There's a couple problems with this. First, we're going to use this for homing - we _want_ to hit the limit switch, that's not an error. Second, checking on the last _tick_ as opposed to the last _step_ basically ensures this won't happen - the last step would be the thing that hits the switch.

sfoster1 (changes requested):
The changes to `pulse` make sense, but I think that hitting the limit switch is _not_ an error. When the stop condition is `MoveStopCondition::limit_switch`, hitting the limit switch during the move is the expected outcome, the happy path. If we want an error, it's actually the case where the move completes _without_ hitting the switch.

So let's change `limit_switch_triggered` as mentioned in the inline comment.

Also, this is extremely core functionality - we need some tests.

include/motor-control/core/motor_interrupt_handler.hpp
Laura-Danielle: I think this all makes sense. One caveat that I think we should take into account is that if a stop condition that was specified is _not_ met, then we should consider that an error.

include/motor-control/core/motor_messages.hpp
```diff
@@ -17,23 +17,34 @@ struct MotionConstraints {
...
     uint8_t seq_id;
+    MoveStopCondition stop_condition = MoveStopCondition::none;
 };
-
 const uint8_t NO_GROUP = 0xff;
 
-enum class AckMessageId : uint8_t { complete = 0x1, error = 0x04 };
+enum class AckMessageId : uint8_t {
+    complete = 0x1,
+    triggered_lim_sw = 0x2,
```
sfoster1, line 38: if we stick with an approach like this, we'll keep having to add new elements when we add new stop conditions. what if we add two new elements, `stopped_from_condition` and `completed_without_condition` or something, and that'll hold us forever

include/motor-control/core/motor_messages.hpp
```diff
@@ -17,23 +17,34 @@ struct MotionConstraints {
...
+    error = 0x04
+};
 
 struct Ack {
     uint8_t group_id;
     uint8_t seq_id;
     uint32_t current_position;
     AckMessageId ack_id;
+    bool lim_sw_triggered;
```
sfoster1, line 47: I also think the `bool lim_sw_triggered` is redundant given that we have the `AckMessageId` already, I don't believe we need it

python/opentrons_ot3_firmware/messages/payloads.py
```diff
@@ -83,6 +83,7 @@ class AddToMoveGroupRequestPayload(MoveGroupRequestPayload):
 class AddLinearMoveRequestPayload(AddToMoveGroupRequestPayload):
```
sfoster1, line 83: We'll need to add the ack id mapping in here to `MoveCompletedPayload` in this file as well

sfoster1 (changes requested):
Almost there! I think we should change the way we do the ack ids and plumb the acknowledgement messaging all the way back out of canbus but the core functionality looks great.

motor-control/tests/test_limit_switch.cpp
```diff
@@ -46,7 +46,8 @@ TEST_CASE("Move with stop condition == limit switch") {
     GIVEN("the limit switch has not been triggered") {
         test_objs.hw.set_mock_lim_sw(false);
         WHEN("the move is finished") {
-            test_objs.handler.finish_current_move();
+            while (!test_objs.handler.pulse())
```
Laura-Danielle, line 49: Generally we try to stay away from using `while()` loops, especially in tests. If you make the move short enough you should only need to call pulse like 3x times or something like that.

python/opentrons_ot3_firmware/messages/payloads.py
caila-marashaj: are you referring to `request_stop_condition` inside `AddLinearMoveRequest` ? Or is there another parameter for ack_id mapping

include/motor-control/core/motor_messages.hpp
```diff
@@ -17,23 +17,33 @@ struct MotionConstraints {
...
+    complete_without_condition = 0x1,
+    stopped_by_condition = 0x2
+};
 
 struct Ack {
     uint8_t group_id;
     uint8_t seq_id;
     uint32_t current_position;
     AckMessageId ack_id;
+    //    bool lim_sw_triggered;
```
sfoster1, line 46: let's remove commented out code

include/motor-control/core/tasks/move_status_reporter_task.hpp
```diff
@@ -33,8 +33,8 @@ class MoveStatusMessageHandler {
             .group_id = message.group_id,
             .seq_id = message.seq_id,
             .current_position = message.current_position,
-            .ack_id =
-                static_cast<uint8_t>(motor_messages::AckMessageId::complete),
+            .ack_id = static_cast<uint8_t>(message.ack_id)
+            //                static_cast<uint8_t>(motor_messages::AckMessageId::complete_without_condition),
```
sfoster1, line 37: remove commented out code

sfoster1 (approved):
Functionality looks good. I think we may someday want "error" as a move complete id back, as well as some "there was no condition, and the move went ok", but I think this is great for now.

Let's just remove that commented out code before merging.

Laura-Danielle (approved):
Looks great to me aside from the commented out code 🥂

PR #240, opened 2022-02-10
feat(common): add sensor messages for all sensor reading requests

Laura-Danielle:
## Overview

Add generic sensor can message request and responses


sfoster1 (approved):
What's `WriteToSensorRequest` for?

PR #242, opened 2022-02-11
feat(bootloader): bootloader starts application on complete update

amitlissack:
See annotations too.

New abilities for bootloader:
- Fixes how we were building 64 bit words to write into flash. Endianness was wrong on byte and word level. 0% right!
- Responds to get fw update status request. This just returns the contents of `0x20000100` at the moment.
- Check power on reason in bootloader. We will look at `0x20000100` only if we aren't powering on. 
- Bootloader gets it's own LD and S scripts.
- Adjust flash location of our applications. This affects LD script and system_XXX.c vector table setup.


DerekMaggio:
Do we use the linker files (.ld) for anything besides setting up memory?

bootloader/firmware/updater.c
```diff
@@ -91,3 +90,35 @@ bool fw_write_to_flash(uint32_t address, uint64_t data) {
...
                              data) == HAL_OK;
 }
+
+
+#define SYSMEM_START APP_FLASH_ADDRESS
+#define SYSMEM_BOOT (SYSMEM_START + 4)
+
+const uint32_t *const sysmem_boot_loc = (uint32_t*)SYSMEM_BOOT;
+
+void fw_update_start_application() {
```
amitlissack, line 100: This is pretty much a copy paste of `app_update_start` in which we jump to the bootloader. 

I need to check if this can be parameterized.

common/firmware/STM32G491RETx/STM32G491RETx_FLASH.ld
```diff
@@ -39,7 +39,7 @@ MEMORY
 {
   CCMSRAM    (xrw)    : ORIGIN = 0x10000000,   LENGTH = 16K
   RAM    (xrw)    : ORIGIN = 0x20000000,   LENGTH = 112K
-  ROM    (rx)    : ORIGIN = 0x08000000,   LENGTH = 512K
+  ROM    (rx)    : ORIGIN = 0x08008000,   LENGTH = 512K
```
amitlissack, line 42: Changing where application resides in flash

common/firmware/system/app_update.c
```diff
@@ -23,9 +24,24 @@ void app_update_start() {
     // Notify bootloader that an update is requested by main application.
     firmware_update_flags = UPDATE_FLAG_REQUESTED;
 
+    // Clear reset flags for bootloader.
+    __HAL_RCC_CLEAR_RESET_FLAGS();
```
amitlissack, line 28: This line resets the `RCC_CSR` register. It hides the reason for a reset. 

Why here? Why now? 

We're about to jump to the bootloader. The bootloader will not check `firmware_update_flags` when on a system reset (`firmware_update_flags` is an uniitialized memory location).

bootloader/core/message_handler.c
```diff
@@ -17,9 +18,12 @@ static HandleMessageReturn handle_initiate_fw_update(const Message* request, Mes
 /** Handle a chunk of the firmware update data. */
 static HandleMessageReturn handle_fw_update_data(const Message* request, Message* response);
 
-/** Handle a firwmare ypdate complete message. */
+/** Handle a firmware update complete message. */
```
sfoster1, line 21: idk, i like the original it seemed like it was out of beowulf

common/firmware/system/app_update.c
sfoster1: Oh, I disagree with this at least for future work. The bootloader detects (or will detect) the reset type to determine if it's running a) because it's the first thing that runs after power up and everything is fine or b) if the app crashed or something. If we only reset this value here, then if we crash before this what will the bootloader see?

sfoster1 (changes requested):
The functionality looks good though I think there's more we can change if we also want the bootloader to be a supervisor that will handle, report, and force an update of bad firmware.

Definitely need to update some comments.

One question though - After this PR, what happens if you take a never-before-flashed X axis board and run e.g. `cmake --build --preset=gantry --target gantry-x-flash` (or however it's spelled) on it? will it be able to boot? 

If not, and forgive me if this already exists and I missed it, we should probably make some default flash targets to put both the bootloader and the application firmware on with the debugger.

sfoster1:
> Do we use the linker files (.ld) for anything besides setting up memory?

linker files are basically a required input to compilation (if you're not explicitly using one, it's using a default one in its install directory). the linker script docs are [here](https://sourceware.org/binutils/docs/ld/Scripts.html) (more or less, there are a lot of little variants). they govern memory layout but also can create or set the address of symbols, drop or add sections, set the location for the first function to run, a bunch of stuff. ours sets up maps for physical memory to symbolic names, then places compiled code elements within those symbolic regions and sets up some tags for locations within there.

common/firmware/system/app_update.c
amitlissack: Are you thinking that the register should be cleared by the bootloader or not at all?

amitlissack:
> The functionality looks good though I think there's more we can change if we also want the bootloader to be a supervisor that will handle, report, and force an update of bad firmware.
> 
> Definitely need to update some comments.
> 
> One question though - After this PR, what happens if you take a never-before-flashed X axis board and run e.g. `cmake --build --preset=gantry --target gantry-x-flash` (or however it's spelled) on it? will it be able to boot?
> 
> If not, and forgive me if this already exists and I missed it, we should probably make some default flash targets to put both the bootloader and the application firmware on with the debugger.

Not yet. In the future.

sfoster1:
> > One question though - After this PR, what happens if you take a never-before-flashed X axis board and run e.g. `cmake --build --preset=gantry --target gantry-x-flash` (or however it's spelled) on it? will it be able to boot?
> > If not, and forgive me if this already exists and I missed it, we should probably make some default flash targets to put both the bootloader and the application firmware on with the debugger.
> 
> Not yet. In the future.

Wait, which part's in the future? If the expectation now is that you have to make sure there's a bootloader on a board before flashing it we'd better pretty heavily document that expectation in readmes and so on.

amitlissack:
> e should probably make some default flash targets to put both the bootloader and the application firmware on with the debugger.

`should probably make some default flash targets to put both the bootloader and the application firmware on with the debugger.` is in the future.

sfoster1 (approved):
Looks great!

PR #243, opened 2022-02-14
fix(can): Wait for a free slot in the TX message fifo before trying sending.

amitlissack:


DerekMaggio (approved):
Lgtm

can/firmware/hal_can.c
```diff
@@ -100,6 +100,11 @@ void can_send_message(HAL_CAN_HANDLE handle, uint32_t arbitration_id, uint8_t* b
         .TxEventFifoControl = FDCAN_NO_TX_EVENTS,
         .MessageMarker=0
     };
+
+    // Wait for there to be room.
+    while (HAL_FDCAN_GetTxFifoFreeLevel(handle) == 0)
```
sfoster1, line 105: Can we [vTaskDelay(1)](https://www.freertos.org/a00127.html) here, little scared of a busyloop like this

sfoster1 (changes requested):
Can we delay instead of straight up looping? Otherwise looks great, will test on hardware today

sfoster1 (approved):
Looks good to me! Works great on hardware

PR #244, opened 2022-02-14
feat(pipettes): add humidity sensor task

Laura-Danielle:
## Overview

Adding in the task for the humidity sensor. There is a top level sensor message handler which will filter out which message queues to write to based on the sensor ID.

Changes will be tested on hardware tomorrow.

## Things not in this PR
- Pushing off MSB vs LSB data handling
- Pushing off sending a message about a sensor having an error

## Review Requests
Does the filtering for sensor ids seem OK? Need some ideas on how to handle checking whether a sensor is 'ready' for taking readings or not but I think that can be pushed to another PR.

All constants are taken from [this](https://www.ti.com/lit/ds/symlink/hdc2080.pdf?ts=1644850101397&ref_url=https%253A%252F%252Fwww.google.com%252F#:~:text=The%20HDC2080%20device%20is%20an,to%20dissipate%20condensation%20and%20moisture.) data sheet.

### DIFF
--- a/include/sensors/core/humidity_sensor.hpp
+++ b/include/sensors/core/humidity_sensor.hpp
@@ -1,0 +1,54 @@
+#pragma once
+
+#include "can/core/ids.hpp"
+
+/*
+ * Configurations and register information for the TI HDC2080 sensor
+ */
+
+namespace humidity_utils {
+
+static auto convert(uint16_t data, can_ids::SensorType type) -> float;
+
+// constants
+static constexpr float TEMP_CONST_MULTIPLIER = 165.0;
+static constexpr float TEMP_CONST = 40.5;
+static constexpr float HUMIDITY_CONST = 100.0;
+// max bits on this sensor is 2^8
+static constexpr float MAX_SIZE = 65536.0;
+
+static constexpr uint16_t ADDRESS = 0x41 << 1;
+static constexpr uint16_t DEVICE_ID = 0x07D0;
+
+// Registers to read from
+static constexpr uint8_t INTERRUPT_REGISTER = 0x07;
+static constexpr uint8_t DRDY_CONFIG = 0x0E;
+// Low configurations for both temperature and humidity
+// this records the status when a reading goes below
+// a certain threshold.
+static constexpr uint8_t TEMPERATURE_REGISTER = 0x00;
+static constexpr uint8_t HUMIDITY_REGISTER = 0x02;
+
+// humidity sensor configurations
+static constexpr uint8_t SAMPLE_RATE =
+    (5 << 4) | (1 << 2) | (1 << 1);  // 1 sample/second, positive DRDY output
+static constexpr uint8_t SET_DATARDY = 1 << 7;
+static constexpr uint8_t BEGIN_MEASUREMENT_RECORDING = 1;
+
+static auto convert(uint16_t data, can_ids::SensorType type) -> float {
+    switch (type) {
+        case can_ids::SensorType::humidity: {
+            // returns humidity in
+            return HUMIDITY_CONST * ((float)data / MAX_SIZE);
+        }
+        case can_ids::SensorType::temperature: {
+            // returns temperature in celsius
+            return TEMP_CONST_MULTIPLIER * ((float)data / MAX_SIZE) -
+                   TEMP_CONST;
+        }
+        default:
+            return 0.0;
+    }
+}
+
+}  // namespace humidity_utils
--- a/include/sensors/core/message_handlers/sensors.hpp
+++ b/include/sensors/core/message_handlers/sensors.hpp
@@ -1,0 +1,56 @@
+#pragma once
+
+#include "can/core/ids.hpp"
+#include "can/core/messages.hpp"
+#include "sensors/core/utils.hpp"
+
+namespace sensor_message_handler {
+
+template <sensor_task_utils::TaskClient SensorTaskClient>
+class SensorHandler {
+  public:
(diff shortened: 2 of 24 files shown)


include/sensors/core/message_handlers/sensors.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+    void visit(can_messages::ReadFromSensorRequest &m) {
+        send_to_queue(can_ids::SensorType(m.sensor), m);
+    }
+
+    void send_to_queue(can_ids::SensorType type,
+                       sensor_task_utils::TaskMessage m) {
+        switch (type) {
+            case can_ids::SensorType::temperature: {
+                client.send_humidity_queue(m);
```
amitlissack, line 44: Is this really supposed to be `send_humidity_queue`?

include/sensors/core/tasks/humidity_task.hpp
```diff
@@ -0,0 +1,165 @@
...
+using namespace humidity_utils;
+using namespace can_ids;
+
+template <message_writer_task::TaskClient CanClient>
+struct HumidityReadingCallback {
+    CanClient &can_client;
+
+    void operator()(const pipette_messages::MaxMessageBuffer &buffer) {
+        uint16_t data = 0x0;
+        const auto *iter = buffer.cbegin();
```
sotaoverride, line 24: should iter be declared a const if it's getting reassigned? what am I missing here?

include/sensors/core/humidity_sensor.hpp
```diff
@@ -0,0 +1,54 @@
...
+namespace humidity_utils {
+
+static auto convert(uint16_t data, can_ids::SensorType type) -> float;
+
+// constants
+static constexpr float TEMP_CONST_MULTIPLIER = 165.0;
+static constexpr float TEMP_CONST = 40.5;
+static constexpr float HUMIDITY_CONST = 100.0;
+// max bits on this sensor is 2^8
+static constexpr float MAX_SIZE = 65536.0;
```
sfoster1, line 18: that's not 2^8 though that's 2^16, either the comment or value here isn't right

include/sensors/core/tasks/humidity_task.hpp
sfoster1: classic cdecl stuff - when defining this `const auto *iter`, it means "an iterator dereferencable to a `const` value whose value is deduced". The way to spell "an iterator whose value cannot be changed" is like `auto *const iter` or something

include/sensors/core/tasks/humidity_task.hpp
```diff
@@ -0,0 +1,165 @@
...
+        auto humidity = humidity_utils::convert(data, SensorType::humidity);
+
+        auto message = can_messages::ReadFromSensorResponse{
+            {}, SensorType::humidity, static_cast<uint16_t>(humidity)};
+        can_client.send_can_message(can_ids::NodeId::host, message);
+    }
+};
+
+template <message_writer_task::TaskClient CanClient>
+struct TemperatureReadingCallback {
```
sfoster1, line 35: building on @amitlissack 's comment above - can we just leave this unimplemented rather than making it the temperature sensor? kind of confusing

include/sensors/core/message_handlers/sensors.hpp
sfoster1: Oh wait this is a combo humidity/temp sensor isn't it

include/sensors/core/humidity_sensor.hpp
```diff
@@ -0,0 +1,54 @@
+#pragma once
+
+#include "can/core/ids.hpp"
+
+/*
+ * Configurations and register information for the TI HDC2080 sensor
```
sfoster1, line 6: this is a good comment but it's the only point in any of this code where you're reminded what this sensor is. for sensor or remote ic bindings, I often find it helpful to have files like this, that store values related to not just humidity or enviro sensors but _this specific chip_, be named after the chip and maybe what it does - like `enviro_sensor_hdc2080.hpp` and the namespace be like `hdc2080`

pipettes/core/tasks.cpp
```diff
@@ -9,6 +9,7 @@
 #include "pipettes/core/can_task.hpp"
 #include "pipettes/core/tasks/eeprom_task_starter.hpp"
 #include "pipettes/core/tasks/i2c_task_starter.hpp"
+#include "sensors//core/tasks/humidity_task_starter.hpp"
```
sfoster1, line 12:
```suggestion
#include "sensors/core/tasks/humidity_task_starter.hpp"
```

sfoster1 (changes requested):
The functionality looks good but 
1. Some of the details about this sensor are a little unclear and lead to confusion. This is an environment sensor combining temperature and RH sensing elements, and we use both of them. Let's refer to it as an `environmental_sensor` or something in the task and object names because otherwise it's weird when you ask the humidity sensor what the temperature is. I also personally like calling what here is `include/sensors/core/humidity_sensor.hpp` something more like `include/sensors/core/hdc2080.hpp` or `enviro_hdc2080.hpp` or `environment/hdc2080.hpp` or something
2. Tests of some sort are probably in order
3. Needs to be in the simulator - a quick implementation that returns hardcoded values of like 25C, 25%RH is probably fine

include/sensors/core/message_handlers/sensors.hpp
Laura-Danielle: yes it is. I'll change the naming to make it less confusing, sorry!

sfoster1 (approved):
Looks good to me!

PR #245, opened 2022-02-15
refactor(can): Use enum typedefs

amitlissack:
Small refactor to typedef enums.

### DIFF
--- a/common/firmware/system/app_update.c
+++ b/common/firmware/system/app_update.c
@@ -26,6 +26,18 @@ void app_update_start() {
     // Disable irqs.
     __disable_irq();
 
+    // systick should be off at boot
+    SysTick->CTRL = 0;
+    SysTick->LOAD = 0;
+    SysTick->VAL = 0;
+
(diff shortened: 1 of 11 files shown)


common/firmware/system/app_update.c
```diff
@@ -26,6 +26,18 @@ void app_update_start() {
     // Disable irqs.
     __disable_irq();
 
+    // systick should be off at boot
+    SysTick->CTRL = 0;
+    SysTick->LOAD = 0;
+    SysTick->VAL = 0;
+
+    /* Clear Interrupt Enable Register & Interrupt Pending Register */
+    for (int i=0;i<8;i++)
```
Laura-Danielle, line 35: Should we not be doing this in the `SystemInit` function of each `system_stm32l5xx.c` file?

Laura-Danielle (approved):
Enum typedef changes look good, but I had a question about the interrupt enable reg and pending reg. I'm not quite sure if this needs to be done in each subproject's individual `system_stm32...c` file.

sfoster1 (approved):
enum away

PR #246, opened 2022-02-15
feat(pipette): Detect mount at boot and assign node id

sfoster1:
Bring up the ability for the pipette to decide which node ID it should use based on the tool mount voltage it senses, reusing code from the similar implementation for head.

For code reusability, the first couple commits commonize first the tool detection (using some fun ranges and filter views which don't work with clang at all, see the comments and the commit message) and then a basic adc interface since the needs are so similar. Once those are done, the work to actually do the sensing on the pipette is surprisingly small.

Still some todos, but definitely reviewable.

TODO:
- [ ] Test on hardware
- [ ] Figure out what to do about the bootloaders (probably just some more expansive voltage ranges or something, but I'm not quite sure where the best customization point for setting the node id is- @amitlissack any ideas there?).


common/core/tool_detection.cpp
```diff
@@ -0,0 +1,74 @@
+#include "common/core/tool_detection.hpp"
+
+#include <array>
+#include <span>
+
+using namespace tool_detection;
+
+constexpr static auto pipette_384_chan_z_bounds =
```
Laura-Danielle, line 8: I thought we were only going to be detecting differences between gripper vs pipette vs no instrument attached

sotaoverride: We have the means to figure out the type of pipette -[this sheet shows the different expected voltages ](https://opentrons.slack.com/files/U01SBA3K485/F02PHGXC9GU/slot_id_resistor_calculations)

sfoster1: @Laura-Danielle we'll probably go that direction eventually but this part was already implemented so I stuck with it

common/tests/test_tool_detection.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <ranges>
+#include <vector>
+
+#include "can/core/ids.hpp"
+#include "catch2/catch.hpp"
+#include "common/core/tool_detection.hpp"
+
+// Filter tests - no BDD here because these macros don't have
+// BDD equivalents
+TEMPLATE_TEST_CASE_SIG("tool list by pipette type", "",
```
amitlissack, line 11: Neato

include/common/core/tool_detection.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+// A full specification of an attached tool, including its bounds
+struct Tool {
+    can_ids::ToolType tool_type;
+    Carrier tool_carrier;
+    ToolCheckBounds bounds;
+    [[nodiscard]] auto within_bounds(millivolts_t reading) const -> bool {
+        return ((reading >= this->bounds.lower) &&
+                (reading < this->bounds.upper));
+        return false;
```
amitlissack, line 36: Why two returns? Am I crazy?

include/common/core/tool_detection.hpp
```diff
@@ -0,0 +1,189 @@
...
+concept ToolList = std::ranges::view<T> &&
+    std::is_same_v<std::iter_value_t<std::ranges::iterator_t<T>>, Tool>;
+
+// Lookup table retrieval with filtering. Does the same thing as lookup_table(),
+// but with a return type that is annoying enough to deal with that we let it be
+// fully auto deduced.
+//
+// Can specify your own base lookup table (including, possibly, the output of
+// another filter function call) or load from the default one.
+template <ToolOrCarrier FilterTarget>
```
amitlissack, line 76: You are a real macher, aren't you?

include/common/core/tool_detection.hpp
sfoster1: What how did that get there why did no tooling care about it gah

pipettes/firmware/main.cpp
```diff
@@ -136,7 +139,7 @@ auto main() -> int {
     pipettes_tasks::start_tasks(can_bus_1, pipette_motor.motion_controller,
                                 pipette_motor.driver, i2c_comms,
                                 // TODO: Load from mount interface
```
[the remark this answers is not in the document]
sfoster1: sure don't!

amitlissack (approved):
Lovely

common/tests/test_tool_detection.cpp
```diff
@@ -0,0 +1,107 @@
...
+            // can't modify the lookup table, gotta copy
+            auto filtered = tool_detection::lookup_table_filtered(which_tool);
+            auto copied = std::vector(filtered.begin(), filtered.end());
+            std::sort(copied.begin(), copied.end(),
+                      [](const tool_detection::Tool& a,
+                         const tool_detection::Tool& b) {
+                          return a.bounds.upper < b.bounds.upper;
+                      });
+            auto trailer =
+                std::ranges::subrange(copied.begin(), --copied.end());
```
Laura-Danielle, line 75: cool tests

Laura-Danielle (approved):
Looks great

common/tests/test_tool_detection.cpp
sfoster1: home, home on the `std::range`

PR #247, opened 2022-02-16
fix(i2c): create I2CWriter globally.

amitlissack:
Fixing a little bug found in simulation. `I2CWriter` was allocated on the stack causing eeprom communication to fail.


Laura-Danielle (approved):
Thanks!

PR #248, opened 2022-02-16
fix(sim): logging uses mutex

amitlissack:
Our simulators were randomly locking up. 

It was due to use of `printf`. See https://www.freertos.org/FreeRTOS-simulator-for-Linux.html `Known Issues`


common/simulation/logging.cpp
```diff
@@ -0,0 +1,26 @@
...
+void log_message(const char * format, ...) {
+
+    // See "Known Issues" section:
+    //  https://www.freertos.org/FreeRTOS-simulator-for-Linux.html
+    auto lock = synchronization::Lock{mutex};
+
+    va_list argp;
+    va_start(argp, format);
+
+    vprintf(format, argp);
```
sfoster1, line 21: [vprintf_s](https://en.cppreference.com/w/c/io/vfprintf) please

common/simulation/CMakeLists.txt
```diff
@@ -22,3 +27,13 @@ target_compile_options(common-simulation
         -Wctor-dtor-privacy
         -fno-rtti
         )
+
+
+# Logging module requires FreeRTOS, which means it cannot be compiled into "common-simulation" shared lib.
+function(target_enable_logging TARGET)
```
sfoster1, line 33: this is cool

sfoster1 (changes requested):
Let's use vprintf_s please (https://en.cppreference.com/w/c/io/vfprintf), `*print*` makes me twitch

common/simulation/logging.cpp
sfoster1: Oh I guess this isn't available in c++.... that's really too bad

sfoster1 (approved):
Looks good to me!

PR #250, opened 2022-02-17
fix: small fixes

amitlissack:
A couple sneaky things.

One was causing `pipettes` executable to be gigantic.


DerekMaggio:
Out of curiosity, what part was causing the executable to be huge?

sotaoverride (approved):

sotaoverride:
> Out of curiosity, what part was causing the executable to be huge?

could be the iostream header

sfoster1 (approved):
wish i could figure out how to get include what you use working with this

PR #252, opened 2022-02-17
feat(i2c): add polling capabilities to the I2C task

Laura-Danielle:
## Overview

Add the ability to poll the i2c bus, which is used by both the capacitive sensor and pressure sensor.

### DIFF
--- a/include/pipettes/core/tasks/i2c_task.hpp
+++ b/include/pipettes/core/tasks/i2c_task.hpp
@@ -35,6 +35,17 @@ class I2CMessageHandler {
                                    TIMEOUT);
         m.client_callback(m.buffer);
     }
+
+    void visit(PollReadFromI2C &m) {
+        auto empty_array = m.buffer;
+        for (int i = 0; i < m.polling; i++) {
+            i2c_device.central_receive(m.buffer.data(), m.buffer.size(),
+                                       m.address, TIMEOUT);
+            m.client_callback(m.buffer);
+            m.buffer = empty_array;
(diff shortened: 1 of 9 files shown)


include/pipettes/core/tasks/i2c_task.hpp
```diff
@@ -35,6 +35,17 @@ class I2CMessageHandler {
                                    TIMEOUT);
         m.client_callback(m.buffer);
     }
+
+    void visit(PollReadFromI2C &m) {
+        auto empty_array = m.buffer;
+        for (int i = 0; i < m.polling; i++) {
+            i2c_device.central_receive(m.buffer.data(), m.buffer.size(),
```
sfoster1, line 42: We should make sure there's a wait or a yield or something somewhere in this loop so we don't busywait the entire cpu. it's also a bit weird to me that the timing isn't defined and isn't saved anywhere, but I don't know what this is for so maybe it's fine.

sfoster1 (approved):
Functionality looks good but let's make sure we're not eating all the cpu spinning through the poll, and also consider either specifying a frequency to poll or saving the times of the responses or something if this is for data we're going to later do digital signal processing things to

include/pipettes/core/tasks/i2c_task.hpp
Laura-Danielle: Yep that was an oversight on my side.

sfoster1 (approved):
looks good!

PR #255, opened 2022-02-18
refactor: generate canbus definitions manually

sfoster1:
Commit the outputs of the header generation, move the python source to
the monorepo,  and add a script and cmake target that will generate
headers from a separately-checked-out monorepo.

The workflow for dealing with canbus definitions mirroring between
python and c/++ was kind of a pain - not here, but in the monorepo. The
monorepo's use of pipenv really isn't friendly to this sort of thing,
and updating the version of this repo used in the monorepo's python was
just as manual as anything this commit does.

After this change, check out the monorepo to a sibling directory and run
cmake --build --preset=<anything> --target update-headers and you'll
write the changes locally.


amitlissack (approved):
Very nice! Way better than what I was thinking.

include/bootloader/core/ids.h
```diff
@@ -0,0 +1,112 @@
+/********************************************
+* This is a generated file. Do not modify.  *
+********************************************/
+#pragma once
+
+/** Can bus arbitration id function code. */
```
Laura-Danielle, line 6: Should this file be here?

include/can/core/ids.hpp
```diff
@@ -0,0 +1,114 @@
...
+};
+
+/** Can bus arbitration id node id. */
+enum class NodeId {
+    broadcast = 0x0,
+    host = 0x10,
+    pipette_left = 0x60,
+    pipette_right = 0x70,
+    gantry_x = 0x30,
+    gantry_y = 0x40,
```
Laura-Danielle, line 76: same for this file

Laura-Danielle (approved):
Had a question about the id files, but otherwise looks good.

sfoster1:
Yeah @Laura-Danielle the concept here is that rather than automatically generating those files when we build, we now manually generate them from an unrelated checkout of the monorepo.

PR #256, opened 2022-02-18
feat(boot): target that flashes app and bootloader

amitlissack:
It "works" on gantrys and head. By "work" i mean that it flashes correctly. It doesn't however know how to build the bootloader bin files.

This is a work in progress and I could use insight from people who know CMake and/or OpenOCD better than me. And that means just about anyone in the world. Thank you.


gantry/firmware/CMakeLists.txt
```diff
@@ -215,17 +215,6 @@ add_custom_target(gantry-debug-y
         )
 
 
+target_stm32G4_flash(gantry-x $<TARGET_FILE_DIR:gantry-x>/gantry-x.bin $<TARGET_FILE_DIR:bootloader-gantry-x>/bootloader-gantry-x.bin)
```
sfoster1, line 218: does this properly pick up the dependency of building the binary firmware files? it seems a bit awkward to haveto use the generator like this

amitlissack: It didn't pick up the binary dependencies correctly. 

I've rewritten this without a function.

amitlissack (commented):
I've redone this using hex files (no need to specify offsets as they're already in the hex file) and not using a function.

head/firmware/CMakeLists.txt
```diff
@@ -144,9 +144,9 @@ add_custom_target(head-debug
         )
 
 
-# Runs openocd to flash the board (without using a debugger)
+# A target to flash bootloader and firmware
 add_custom_target(head-flash
-        COMMAND "${OpenOCD_EXECUTABLE}" "-f" "${COMMON_EXECUTABLE_DIR}/STM32G491RETx/stm32g4discovery.cfg" "-c" "program $<TARGET_FILE:head>;reset;exit"
+        COMMAND "${OpenOCD_EXECUTABLE}" "-f" "${COMMON_EXECUTABLE_DIR}/STM32G491RETx/stm32g4discovery.cfg" "-c" "init;reset init;halt;stm32g4x mass_erase 0;flash write_image $<TARGET_FILE_DIR:bootloader-head>/bootloader-head.hex;flash write_image head.hex;reset;exit"
```
Laura-Danielle, line 149: I think it might be nice/more readable to make the stuff past the `-c` a template file and configure the file in `cmakelists.txt` (similar to the gdbinit file).

amitlissack: Good idea. I'll check it out.

sfoster1 (approved):
lgtm! nice work!

Laura-Danielle (approved):

PR #258, opened 2022-02-22
Ret 501 message handling

caila-marashaj:


include/can/core/messages.hpp
```diff
@@ -124,7 +124,6 @@ struct AddLinearMoveRequest : BaseMessage<MessageId::add_move_request> {
     ticks duration;
     um_per_tick_sq acceleration;
     um_per_tick velocity;
-    uint8_t request_stop_condition;
```
Laura-Danielle, line 127: I think you accidentally removed this here? We'll want to be able to specify a stop condition for linear moves sometimes (i.e. for capacitance).

include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -132,6 +132,22 @@ class MotorInterruptHandler {
         return false;
     }
 
+    auto homing_stopped() -> bool {
+        if (!can_step() && !limit_switch_triggered()) {
+            finish_current_move(AckMessageId::timeout);
+            return true;
+        }
+        if (can_step() && limit_switch_triggered()) {
```
sfoster1, line 140: When you're homing, you don't know where you are. In practice, you'll be commanding a move that is longer than your motion extents. You _expect_, and want, to hit the limit switch _before_ you're done moving, because you don't know where you are. It's completely fine to hit the limit switch before you're done.

sfoster1 (changes requested):
The checks aren't quite right - almost all the time, we'll hit the switch before we're done moving.

include/motor-control/core/motor_messages.hpp
```diff
@@ -34,8 +34,9 @@ struct Move {        // NOLINT(cppcoreguidelines-pro-type-member-init)
 const uint8_t NO_GROUP = 0xff;
 
 enum class AckMessageId : uint8_t {
-    complete_without_condition = 0x1,
-    stopped_by_condition = 0x2
+    complete = 0x1,
```
sfoster1, line 37: I'm a little confused by this - I think `complete_without_condition` and `stopped_by_condition` are perfectly adequate for this, right? `complete_without_condition` is the default completion for moves without condition, but in the context of home it probably means there was an error; `stopped_by_condition` is a good value for when a condition, well, stopped the move.

Laura-Danielle:
I think we should also add a test in `test_move_group.cpp` for `HomeRequest`

Laura-Danielle (approved):
Looks great to me!

sfoster1 (approved):
Nice!

PR #259, opened 2022-02-23
fix(head): presence sensing fix(?) 

amitlissack:
`adc_setup` was not being called which led to `adc_set_chan` to enter the error handler because the handle's `Instance` member was `NULL`.

I've tested this on my nucleo board and nothing breaks.


sotaoverride (approved):
Thank you!!

sfoster1 (approved):
lgtm!

PR #261, opened 2022-02-24
feat(pipettes): add capacitive sensor support

Laura-Danielle:
## Overview

Apologies for the PR being all over the place. I tried to split up the commits, at least, to make it more readable.

Overall, this PR adds support for the fdc1004 capacitive sensor. There is an ability to set an offset -- depending on the environment the sensor is in. The idea here for the offset specifically is that you would be able to get the new offset by sending a `BaselineSensorRequest`, and then you would send over a `WriteToSensor` request to update the current sensor offset on the capacitive sensor itself.

## Changelog
- Added in an `offset` parameter to the `Read/Baseline` requests which will determine whether you are taking a sensor read or an offset read (this is used for both the pressure and capacitive sensor)
- Added the ability to poll from multiple registers
- Modified polling so that it sends write/read requests to the i2c bus for a given register
- Created abstract base classes for the two callback types used for the sensor tasks
- overloaded the operator function of the callback so that you can send a final can message once the firmware has taken an average read for polling
- Added a test for the capacitive sensor

## Review Requests
Please let me know if anything is confusing!

### DIFF
--- a/include/pipettes/core/i2c_writer.hpp
+++ b/include/pipettes/core/i2c_writer.hpp
@@ -41,7 +41,7 @@ class I2CWriter {
         queue->try_write(write_msg);
     }
 
-    /*
+    /**
      * A single read to the i2c bus.
      *
      * @param device_address: device address on the i2c bus
@@ -68,7 +68,7 @@ class I2CWriter {
         queue->try_write(read_msg);
     }
 
-    /*
+    /**
      * Single Register Poll.
      *
      * This function will poll the results of a single register.
@@ -104,7 +104,7 @@ class I2CWriter {
         queue->try_write(read_msg);
     }
 
-    /*
+    /**
      * Multi Register Poll.
      *
      * This function will poll the results of two registers. Typically this is
@@ -138,11 +138,11 @@ class I2CWriter {
         pipette_messages::MultiRegisterPollReadFromI2C read_msg{
             .address = device_address,
             .polling = number_reads,
+            .delay_ms = delay,
             .register_buffer_1 = buffer_1,
             .register_buffer_2 = buffer_2,
             .client_callback = std::move(client_callback),
-            .handle_buffer = std::move(handle_callback),
-            .delay_ms = delay};
+            .handle_buffer = std::move(handle_callback)};
         queue->try_write(read_msg);
     }
 
--- a/include/pipettes/core/i2c_writer.hpp
+++ b/include/pipettes/core/i2c_writer.hpp
@@ -12,6 +12,7 @@
 namespace i2c_writer {
 
 using namespace pipette_messages;
+using namespace sensor_callbacks;
 
 using TaskMessage =
     std::variant<std::monostate, WriteToI2C, ReadFromI2C,
@@ -40,51 +41,98 @@ class I2CWriter {
         queue->try_write(write_msg);
     }
 
-    void read(uint16_t device_address,
-              sensor_callbacks::SingleRegisterCallback* callback,
-              uint8_t reg = 0x0) {
-        // We want to copy the callback every time as opposed to passing a
-        // reference to it from the eeprom task.
-        std::array<uint8_t, MAX_SIZE> max_buffer{reg};
-        pipette_messages::ReadFromI2C read_msg{.address = device_address,
-                                               .buffer = max_buffer,
-                                               .client_callback = callback};
-        queue->try_write(read_msg);
-    }
-
     /*
-     * This function will poll the results of a single register.
+     * A single read to the i2c bus.
+     *
+     * @param device_address: device address on the i2c bus
+     * @param client_callback: the function that will write a CAN return message
+     * to the client queue.
+     * @param handle_callback: the function that will handle any data
+     * manipulation required for the specific task using the i2c bus. Takes in
+     * two buffers.
+     * @param reg: the register to read from, if relevant.
      *
      * A poll function automatically initiates a read/write to the specified
      * register on the i2c bus.
      */
-    void single_register_poll(
-        uint16_t device_address, uint8_t number_reads, uint16_t delay,
-        sensor_callbacks::SingleRegisterCallback* callback, uint8_t reg = 0x0) {
+    void read(uint16_t device_address, SingleBufferTypeDef handle_callback,
+              SendToCanFunctionTypeDef client_callback, uint8_t reg = 0x0) {
+        // We want to copy the callback every time as opposed to passing a
+        // reference to it from the eeprom task.
+        std::array<uint8_t, MAX_SIZE> max_buffer{reg};
+        pipette_messages::ReadFromI2C read_msg{
+            .address = device_address,
+            .buffer = max_buffer,
+            .handle_buffer = handle_callback,
+            .client_callback = client_callback};
+        queue->try_write(read_msg);
+    }
+
+    /*
+     * Single Register Poll.
+     *
+     * This function will poll the results of a single register.
+     *
+     * @param device_address: device address on the i2c bus
+     * @param number_reads: how many data points to take from the sensor
+     * @param delay: the frequency to send the read/writes to the register
+     * provided.
+     * @param client_callback: the function that will write a CAN return message
+     * to the client queue.
+     * @param single_callback: the function that will handle any data
+     * manipulation required for the specific task using the i2c bus.
+     * @param reg: the register to read from, if relevant.
+     *
+     * A poll function automatically initiates a read/write to the specified
+     * register on the i2c bus.
+     */
+    void single_register_poll(uint16_t device_address, uint8_t number_reads,
+                              uint16_t delay,
+                              SendToCanFunctionTypeDef client_callback,
+                              SingleBufferTypeDef handle_callback,
+                              uint8_t reg = 0x0) {
         // We want to copy the callback every time as opposed to passing a
         // reference to it from the eeprom task.
         std::array<uint8_t, MAX_SIZE> max_buffer{reg};
         pipette_messages::SingleRegisterPollReadFromI2C read_msg{
             .address = device_address,
             .polling = number_reads,
+            .delay_ms = delay,
             .buffer = max_buffer,
-            .client_callback = callback,
-            .delay_ms = delay};
+            .client_callback = client_callback,
+            .handle_buffer = handle_callback};
         queue->try_write(read_msg);
     }
 
     /*
+     * Multi Register Poll.
+     *
      * This function will poll the results of two registers. Typically this is
      * relevant when data is stored in two separate registers.
      *
+     * @param device_address: device address on the i2c bus
+     * @param number_reads: how many data points to take from the sensor
+     * @param delay: the frequency to send the read/writes to the register
+     * provided.
+     * @param client_callback: the function that will write a CAN return message
+     * to the client queue.
+     * @param multi_callback: the function that will handle any data
+     * manipulation required for the specific task using the i2c bus. Takes in
+     * two buffers.
+     * @param register_1: the register to read from, if relevant.
+     * @param register_2: the register to read from, if relevant.
+     *
+     * Registers will be written to/read from in the order they are listed. This
+     * means that `register_1` will always be read from before `register_2`.
+     *
      * A poll function automatically initiates a read/write to the specified
-     * registers on the i2c bus.
+     * register on the i2c bus.
      */
-    void multi_register_poll(uint16_t device_address, uint8_t number_reads,
-                             uint16_t delay,
-                             sensor_callbacks::MultiRegisterCallback* callback,
-                             uint8_t register_1 = 0x0,
-                             uint8_t register_2 = 0x0) {
+    void multi_register_poll(
+        uint16_t device_address, uint8_t number_reads, uint16_t delay,
+        sensor_callbacks::SendToCanFunctionTypeDef client_callback,
+        sensor_callbacks::MultiBufferTypeDef handle_callback,
+        uint8_t register_1 = 0x0, uint8_t register_2 = 0x0) {
         std::array<uint8_t, MAX_SIZE> buffer_1{register_1};
         std::array<uint8_t, MAX_SIZE> buffer_2{register_2};
         pipette_messages::MultiRegisterPollReadFromI2C read_msg{
@@ -92,7 +140,8 @@ class I2CWriter {
             .polling = number_reads,
             .register_buffer_1 = buffer_1,
             .register_buffer_2 = buffer_2,
-            .client_callback = callback,
+            .client_callback = client_callback,
+            .handle_buffer = handle_callback,
             .delay_ms = delay};
         queue->try_write(read_msg);
     }
--- a/include/pipettes/core/i2c_writer.hpp
+++ b/include/pipettes/core/i2c_writer.hpp
@@ -13,7 +13,7 @@ namespace i2c_writer {
 using namespace pipette_messages;
 
 using TaskMessage =
-    std::variant<std::monostate, WriteToI2C, ReadFromI2C, PollReadFromI2C>;
+    std::variant<std::monostate, WriteToI2C, ReadFromI2C, SingleRegisterPollReadFromI2C>;
 
 template <template <class> class QueueImpl>
 requires MessageQueue<QueueImpl<TaskMessage>, TaskMessage>
@@ -39,7 +39,7 @@ class I2CWriter {
     }
 
     void read(uint16_t device_address,
-              const Callback
+              const SingleRegisterCallback
                   callback,  // NOLINT (performance-unnecessary-value-param)
               uint8_t reg = 0x0) {
         // We want to copy the callback every time as opposed to passing a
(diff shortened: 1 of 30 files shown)


include/sensors/core/callback_types.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+namespace sensor_callbacks {
+using MaxMessageBuffer = std::array<uint8_t, 5>;
+
+// NOLINTNEXTLINE(cppcoreguidelines-special-member-functions)
+class SingleRegisterCallback {
+  public:
+    virtual void operator()(const MaxMessageBuffer &buffer) = 0;
```
amitlissack, line 9: One of the expectations of an interface like this is that someone who needs to create an implementation will know what the methods need to do. 

I don't know when `operator()()` is called verses the other `operator`. What should happen in these functions?

Laura-Danielle: I'll add documentation

include/sensors/core/callback_types.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+namespace sensor_callbacks {
+using MaxMessageBuffer = std::array<uint8_t, 5>;
+
+// NOLINTNEXTLINE(cppcoreguidelines-special-member-functions)
+class SingleRegisterCallback {
```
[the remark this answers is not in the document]
Laura-Danielle: I chose to do a pure virtual function because I honestly didn't know how else to type this callback. The reason for having two overloaded operator functions is so that the i2c task can separately choose to send a CAN message back. This is most relevant for polling a specific i2c bus where you'll want to accumulate the data without actually sending a message back.

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
```diff
@@ -0,0 +1,129 @@
...
+        can_client.send_can_message(can_ids::NodeId::host, message);
+    }
+
+    void reset() {
+        number_of_reads = 1;
+        measurement = 0;
+    }
+
+    void set_number_of_reads(uint16_t number_of_reads) {
+        number_of_reads = number_of_reads;
```
amitlissack, line 54: This is a weird line to look at. Can you please make it 
```this->number_of_reads = number_of_reads;```?

File: include/sensors/core/tasks/capacitive_sensor_callbacks.hpp [hunk 4]
```diff
@@ -0,0 +1,129 @@
...
+        can_client.send_can_message(can_ids::NodeId::host, message);
+    }
+
+    void reset() {
+        measurement = 0;
+        number_of_reads = 1;
+    }
+
+    void set_number_of_reads(uint16_t number_of_reads) {
+        number_of_reads = number_of_reads;
```
amitlissack, line 100: Same as above

include/sensors/core/tasks/capacitive_sensor_task.hpp
```diff
@@ -0,0 +1,154 @@
...
+        THRESHOLD = m.threshold;
+        auto message = can_messages::SensorThresholdResponse{
+            {}, SensorType::capacitive, THRESHOLD};
+        can_client.send_can_message(can_ids::NodeId::host, message);
+    }
+
+    InternalCallback internal_callback{};
+    sensor_task_utils::BitMode mode = sensor_task_utils::BitMode::MSB;
+    // 8 pF
+    uint32_t THRESHOLD = 0x8;
```
amitlissack, line 107: These three are in all caps. Are they constants?

include/pipettes/core/tasks/i2c_task.hpp
```diff
@@ -33,18 +33,46 @@ class I2CMessageHandler {
     void visit(ReadFromI2C &m) {
         i2c_device.central_receive(m.buffer.data(), m.buffer.size(), m.address,
                                    TIMEOUT);
-        m.client_callback(m.buffer);
+        m.handle_buffer(m.buffer);
+        m.client_callback();
     }
 
-    void visit(PollReadFromI2C &m) {
+    void visit(SingleRegisterPollReadFromI2C &m) {
```
amitlissack, line 40: It looks like processing `SingleRegisterPollReadFromI2C` and `MultiRegisterPollReadFromI2C`  messages will keep this task busy for the duration of the poll. 

Is there a concern that blocking all other I2C communication during the poll can be a problem?

Should the polling be happening from a different task?

amitlissack (commented):
A serious question about polling within the i2c task. Otherwise rock solid.

include/sensors/core/tasks/capacitive_sensor_task.hpp
Laura-Danielle: Sort of..you can change the threshold to something else from the python side but it shouldn't be changed frequently.

include/pipettes/core/tasks/i2c_task.hpp
Laura-Danielle: I wanted it to block i2c communication because you can't read/write from two things at the same time on the same bus which is why I had it block in the i2c task.

include/sensors/core/tasks/capacitive_sensor_task.hpp
amitlissack: Ok. Can we make them lower case or `const`?

include/pipettes/core/i2c_writer.hpp
```diff
@@ -38,32 +41,108 @@ class I2CWriter {
...
+     * manipulation required for the specific task using the i2c bus. Takes in
+     * two buffers.
+     * @param reg: the register to read from, if relevant.
+     *
+     * A poll function automatically initiates a read/write to the specified
+     * register on the i2c bus.
+     */
+    void read(uint16_t device_address, SingleBufferTypeDef handle_callback,
+              SendToCanFunctionTypeDef client_callback, uint8_t reg = 0x0) {
         // We want to copy the callback every time as opposed to passing a
```
sfoster1, line 60: can we remove this comment? it's not quite accurate (we're moving, not copying) and it also doesnt mention why we'd want to do this

include/pipettes/core/i2c_writer.hpp
```diff
@@ -38,32 +41,108 @@ class I2CWriter {
...
+     * @param client_callback: the function that will write a CAN return message
+     * to the client queue.
+     * @param single_callback: the function that will handle any data
+     * manipulation required for the specific task using the i2c bus.
+     * @param reg: the register to read from, if relevant.
+     *
+     * A poll function automatically initiates a read/write to the specified
+     * register on the i2c bus.
+     */
+    void single_register_poll(uint16_t device_address, uint8_t number_reads,
```
sfoster1, line 89: let's give `number_reads` a bigger datatype just to get ahead of ourselves

include/pipettes/core/i2c_writer.hpp
```diff
@@ -38,32 +41,108 @@ class I2CWriter {
...
+                              uint16_t delay,
+                              SendToCanFunctionTypeDef client_callback,
+                              SingleBufferTypeDef handle_callback,
+                              uint8_t reg = 0x0) {
         // We want to copy the callback every time as opposed to passing a
```
sfoster1, line 94: remove or update

include/pipettes/core/tasks/i2c_task.hpp
```diff
@@ -33,18 +33,52 @@ class I2CMessageHandler {
...
         auto empty_array = m.buffer;
         for (int i = 0; i < m.polling; i++) {
+            i2c_device.central_transmit(m.buffer.data(), m.buffer.size(),
+                                        m.address, TIMEOUT);
             i2c_device.central_receive(m.buffer.data(), m.buffer.size(),
                                        m.address, TIMEOUT);
-            m.client_callback(m.buffer);
+            m.handle_buffer(m.buffer);
             m.buffer = empty_array;
             i2c_device.wait_during_poll(m.delay_ms);
```
sfoster1, line 52: we also might want to delay this to another pr but this is not how you want to do fixed-frequency polling since it adds the time taken for the message to the wait. If you pass 100ms here, then you'll be reading every 100ms + transmit/receive time. Usually what you want to do is take a timestamp before transmission begins and then use something like [`vTaskDelayUntil`](https://www.freertos.org/vtaskdelayuntil.html) to delay until `timestamp+100ms`, which gives you more consistent timing (or use a timer)

include/sensors/core/callback_types.hpp
```diff
@@ -0,0 +1,71 @@
...
+    {cb.handle_data(buffer, buffer)};
+    {cb.send_to_can()};
+};
+
+template <SingleRegister scb>
+struct SingleRegisterCallback {
+    SingleRegisterCallback(scb &callback) : callback{callback} {}
+    scb &callback;
+
+    void operator()(const MaxMessageBuffer &buffer) {
```
sfoster1, line 45: Let's not do this thing with the `operator()()` overloads. It's a really bad idea to make function overloads do totally different things, it makes code unreadable. Function overloads are best used to do fundamentally the same thing from somewhat different data - maybe a level of indirection, maybe a cast. Having two function overloads that do two entirely different things is _really_ hard to parse.

include/sensors/core/fdc1004.hpp
```diff
@@ -0,0 +1,78 @@
...
+ * We are currently only using 1 out of the 4 measurement channels.
+ *
+ * By default, CAPDAC is enabled. The max resolution of the capdac is 5 bits.
+ *
+ * Capacitance measurement is a total of 24 bits long. You must do consecutive
+ * reads of the MSB and LSB registers to get the data.
+ */
+
+namespace fdc1004_utils {
+static auto convert(uint32_t measurement, uint16_t accumulation,
```
sfoster1, line 23: `static` here is not a good idea IMO - what it will do is force the function definition to live in every single translation unit that includes this file, even transitively. These are just plain old pure functions, just declare them without static i.e. `auto convert(...) -> sq14_15;`

include/sensors/core/fdc1004.hpp
```diff
@@ -0,0 +1,78 @@
...
+                    float current_offset) -> sq14_15;
+static auto calculate_capdac(uint16_t measurement, uint16_t accumulation,
+                             float current_offset) -> uint16_t;
+
+// Capacitance Sensor Address and Registers
+static constexpr uint16_t ADDRESS = 0x50 << 1;
```
sfoster1, line 29: [`constexpr` implies `const` implies `static`](https://stackoverflow.com/a/50489130) on variables at namespace scope, can just be `constexpr`

include/sensors/core/fdc1004.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+// Constants
+static constexpr int MAX_CAPDAC_RESOLUTION = 5;
+static constexpr int MSB_SHIFT = 16;
+static constexpr uint16_t MAX_CAPDAC_OFFSET = 0x1F << MAX_CAPDAC_RESOLUTION;
+static constexpr float MAX_MEASUREMENT = 524288.0;
+// single ended capdac offset measurement in pF
+static constexpr float CAPDAC_OFFSET = 3.125;
+
+[[maybe_unused]] static auto convert(uint32_t measurement,
```
sfoster1, line 54: why `[[maybe_unused]]`?

sfoster1, line 54: as with the above functions, this really shouldn't do `static` - it's pretty confusing in a header. If you want to keep the definition in the header, use `inline`

include/sensors/core/fdc1004.hpp
```diff
@@ -0,0 +1,78 @@
...
+                                     uint16_t accumulation,
+                                     float current_offset) -> sq14_15 {
+    // shift the 32 bit value number down to 24 bits
+    auto average =
+        static_cast<float>(measurement >> 8) / static_cast<float>(accumulation);
+    float capacitance = average / MAX_MEASUREMENT + current_offset;
+    return convert_to_fixed_point(capacitance, 15);
+}
+
+[[maybe_unused]] static auto calculate_capdac(uint16_t measurement,
```
sfoster1, line 64: why `[[maybe_unused]]`?

sfoster1, line 64: `inline` instead of `static` please

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
```diff
@@ -0,0 +1,128 @@
...
+                     const sensor_callbacks::MaxMessageBuffer &LSB_buffer) {
+        uint32_t msb_data = 0x0;
+        uint32_t lsb_data = 0x0;
+        const auto *MSB_iter = MSB_buffer.cbegin();
+        MSB_iter =
+            bit_utils::bytes_to_int(MSB_iter, MSB_buffer.cend(), msb_data);
+        const auto *LSB_iter = LSB_buffer.cbegin();
+        LSB_iter =
+            bit_utils::bytes_to_int(LSB_iter, LSB_buffer.cend(), lsb_data);
+        measurement += (msb_data << MSB_SHIFT | lsb_data);
```
sfoster1, line 36: why `+=`? shouldn't this just be `measurement = (...)`?

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
```diff
@@ -0,0 +1,128 @@
...
+struct ReadCapacitanceCallback {
+  public:
+    ReadCapacitanceCallback(CanClient &can_client, float current_offset)
+        : can_client{can_client}, current_offset{current_offset} {}
+
+    void handle_data(const sensor_callbacks::MaxMessageBuffer &MSB_buffer,
+                     const sensor_callbacks::MaxMessageBuffer &LSB_buffer) {
+        uint32_t msb_data = 0x0;
+        uint32_t lsb_data = 0x0;
+        const auto *MSB_iter = MSB_buffer.cbegin();
```
sfoster1, line 30: the implied bit/byte order here sort of worries me; can we use `uint16_t` for the data elements here, and widen when we combine the two values with `MSB_SHIFT`?

motor-control/core/utils.cpp
```diff
@@ -25,3 +25,8 @@ sq0_31 fixed_point_multiply(sq31_31 a, sq0_31 b) {
     int64_t result = a * static_cast<int64_t>(b);
     return static_cast<sq0_31>((result >> 31) & 0xFFFFFFFF);
 }
+
+float fixed_point_to_float(uint32_t data, int to_radix) {
+    const uint32_t power_of_two = 2 << to_radix;
```
sfoster1, line 30: feels like it should be a float, right?

include/sensors/core/fdc1004.hpp
```diff
@@ -0,0 +1,78 @@
...
+                                              uint16_t accumulation,
```
sfoster1, line 65: can we call this `read_count` or something? `accumulation` doesn't really give you a lot to go on, and I think it's just the number of values that we'll be reading right

include/sensors/core/fdc1004.hpp
```diff
@@ -0,0 +1,78 @@
...
+                                              float current_offset)
+    -> uint16_t {
+    float average =
+        static_cast<float>(measurement) / static_cast<float>(accumulation);
+    float offset = (average + current_offset) / CAPDAC_OFFSET;
```
sfoster1, line 70: This is kind of a weird calculation, can we comment what it's up to? `CAPDAC_OFFSET` above is noted as being in pF, and the average + current offset values are presumably somehow related to pF, so is the output here unitless? Bit confused.

include/sensors/core/fdc1004.hpp
```diff
@@ -0,0 +1,78 @@
...
+// Constants
+static constexpr int MAX_CAPDAC_RESOLUTION = 5;
+static constexpr int MSB_SHIFT = 16;
+static constexpr uint16_t MAX_CAPDAC_OFFSET = 0x1F << MAX_CAPDAC_RESOLUTION;
+static constexpr float MAX_MEASUREMENT = 524288.0;
+// single ended capdac offset measurement in pF
+static constexpr float CAPDAC_OFFSET = 3.125;
+
+[[maybe_unused]] static auto convert(uint32_t measurement,
+                                     uint16_t accumulation,
```
sfoster1, line 55: let's change this name too

include/sensors/core/fdc1004.hpp
```diff
@@ -0,0 +1,78 @@
...
+                                     float current_offset) -> sq14_15 {
+    // shift the 32 bit value number down to 24 bits
```
sfoster1, line 57: this is weird - if this is a 24 bit value, it should really be in the 24 least significant bits

sfoster1 (changes requested):
Had some inline notes and change requests. The structure is there but I think there's some stuff we need to change, mostly for readability.

Aside from that, it feels a bit weird to me to have this request-response thing with the host having to read a new offset, get it from can, then send it back over can, then request data, which has the offset baked into it - so why does the host ever need to see the offset value? why not have some `update_offset` can command that, sure, maybe it returns the offset, but for informational purposes only, it updated the stored value on its own.

include/pipettes/core/tasks/i2c_task.hpp
Laura-Danielle: Put in a TODO comment for a follow-up PR.

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
```diff
@@ -0,0 +1,128 @@
...
+struct ReadOffsetCallback {
+  public:
+    ReadOffsetCallback(CanClient &can_client, float current_offset)
+        : can_client{can_client}, current_offset{current_offset} {}
+
+    void handle_data(const sensor_callbacks::MaxMessageBuffer &buffer) {
+        uint16_t data = 0x0;
+        const auto *iter = buffer.cbegin();
+        iter = bit_utils::bytes_to_int(iter, buffer.cend(), data);
+        measurement += data;
```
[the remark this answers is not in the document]
Laura-Danielle: You might take multiple readings of read offset, or you might not. Rather than having two separate callbacks for a single read of capacitance/capdac respectively, I decided to just combine it into one.

include/sensors/core/fdc1004.hpp
Laura-Danielle: Clang was getting mad at me previously saying that the callback wasn't always used -- I think changing things from using a virtual base class fixed it.

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
```diff
@@ -0,0 +1,94 @@
+#pragma once
+#include <iostream>
```
sfoster1, line 2: use `LOG`
```suggestion
```

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
```diff
@@ -0,0 +1,94 @@
...
+        measurement += (msb_data << MSB_SHIFT | lsb_data);
+    }
+
+    void send_to_can() {
+        uint32_t capacitance =
+            convert_capacitance(measurement, number_of_reads, current_offset);
+        auto message = can_messages::ReadFromSensorResponse{
+            {}, SensorType::capacitive, capacitance};
+        can_client.send_can_message(can_ids::NodeId::host, message);
+        std::cout << "CAPACITANCE " << static_cast<int>(capacitance) << "\n";
```
sfoster1, line 46:
```suggestion
        LOG("CAPACITANCE %d\n", static_cast<int>(capacitance));
```

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
```diff
@@ -0,0 +1,94 @@
...
+        if (capacitance > zero_threshold || capacitance < -zero_threshold) {
+            std::cout << "greater than threshold " << "\n";
```
sfoster1, line 48:
```suggestion
            LOG("greater than threshold\n");
```

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
```diff
@@ -0,0 +1,94 @@
...
+            auto capdac = update_capdac(capacitance, current_offset);
+            current_offset = capdac;
+            std::cout << "current offset " << static_cast<int>(current_offset)  << "\n";
```
sfoster1, line 51:
```suggestion
            LOG("current offset %d\n", static_cast<int>(current_offset));
```

include/sensors/core/tasks/capacitive_sensor_task.hpp
```diff
@@ -0,0 +1,155 @@
...
+        /**
+         * The FDC1004 sensor has an offset register and
+         * a measurement register. The CAPDAC offset is used
+         * to account for any additional baseline noise that
+         * may occur on the capacitve sensor.
+         *
+         */
+        LOG("Received request to read from %d sensor\n", m.sensor);
+        capdac_offset = capacitance_handler.get_offset();
+        std::cout << "capdac  " << static_cast<int>(capdac_offset) << "\n";
```
sfoster1, line 63:
```suggestion
        LOG("capdac  %d\n", static_cast<int>(capdac_offset));
```

include/sensors/core/tasks/capacitive_sensor_task.hpp
```diff
@@ -0,0 +1,155 @@
...
+        if (bool(m.offset_reading)) {
+            std::cout << "offset reading  " << "\n";
```
sfoster1, line 65:
```suggestion
            LOG("offset reading \n");
```

sfoster1 (changes requested):
Looks great, gotta remove `#include <iostream>` and use of `cout` in a couple places though

sfoster1 (approved):
Awesome!

PR #262, opened 2022-02-24
refactor(motor-driver): update tmc2130 communications

ahiuchingau:
This copies @fsinapi's TMC2130 implementation in [opentrons-modules](https://github.com/Opentrons/opentrons-modules/pull/232) for the OT3. We now have a nice `TMC2130RegisterMap` so the driver configuration is more readable and can be modified easily. We're keeping the way SPI works in the OT3 for now, and we should think about doing refactor there and make SPI its own task.

### DIFF
--- a/gantry/core/utils.cpp
+++ b/gantry/core/utils.cpp
@@ -26,12 +26,12 @@ auto utils::register_config_by_axis(enum GantryAxisType which)
                                .run_current = 0x10,
                                .hold_current_delay = 0x7},
                 .thigh = {.threshold = 0xFFFFF},
-                .chopconf = {.toff = 0b101,
-                             .hstrt = 0b101,
-                             .hend = 0b11,
-                             .tbl = 0b10,
-                             .mres = 0b100},
-                .coolconf = {.sgt = 0b110}};
+                .chopconf = {.toff = 0x5,
+                             .hstrt = 0x5,
+                             .hend = 0x3,
+                             .tbl = 0x2,
+                             .mres = 0x4},
+                .coolconf = {.sgt = 0x6}};
 
         case GantryAxisType::gantry_y:
             return tmc2130::TMC2130RegisterMap{
--- a/gantry/core/utils.cpp
+++ b/gantry/core/utils.cpp
@@ -17,25 +17,41 @@ auto utils::get_node_id() -> can_ids::NodeId {
 }
 
 auto utils::register_config_by_axis(enum GantryAxisType which)
-    -> motor_driver_config::RegisterConfig {
+    -> tmc2130::TMC2130RegisterMap {
     switch (which) {
         case GantryAxisType::gantry_x:
-            return motor_driver_config::RegisterConfig{.gconf = 0x04,
-                                                       .ihold_irun = 0x71002,
-                                                       .chopconf = 0x40101D5,
-                                                       .thigh = 0xFFFFF,
-                                                       .coolconf = 0x60000};
+            return tmc2130::TMC2130RegisterMap{
+                .gconfig = {.en_pwm_mode = 1},
+                .ihold_irun = {.hold_current = 0x2,
+                               .run_current = 0x10,
+                               .hold_current_delay = 0x7},
+                .thigh = {.threshold = 0xFFFFF},
+                .chopconf = {.toff = 0b101,
+                             .hstrt = 0b101,
+                             .hend = 0b11,
+                             .tbl = 0b10,
+                             .mres = 0b100},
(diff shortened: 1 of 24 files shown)


include/common/core/spi.hpp
```diff
@@ -26,8 +26,8 @@ class TMC2130Spi {
      * @param transmit The transmit buffer.
      * @param receive The receive buffer.
      */
-    virtual void transmit_receive(const TMC2130Spi::BufferType& transmit,
-                                  TMC2130Spi::BufferType& receive) = 0;
+    virtual auto transmit_receive(const TMC2130Spi::BufferType& transmit,
+                                  TMC2130Spi::BufferType& receive) -> bool;
```
sotaoverride, line 30: should this still be a pure virtual function?virtual auto transmit_receive(const ```
TMC2130Spi::BufferType& transmit,
                                  TMC2130Spi::BufferType& receive) -> bool = 0;
```

File: gantry/core/utils.cpp [hunk 2]
```diff
@@ -17,25 +17,41 @@ auto utils::get_node_id() -> can_ids::NodeId {
...
-                                                       .coolconf = 0x60000};
+            return tmc2130::TMC2130RegisterMap{
+                .gconfig = {.en_pwm_mode = 1},
+                .ihold_irun = {.hold_current = 0x2,
+                               .run_current = 0x10,
+                               .hold_current_delay = 0x7},
+                .tpowerdown = {},
+                .tcoolthrs = {.threshold = 0},
+                .thigh = {.threshold = 0xFFFFF},
+                .chopconf = {.toff = 0x5,
```
Laura-Danielle, line 45: Can we choose to either represent numbers for each register either in binary or hex? It's a bit confusing that chopconf here is in hex whereas above it's in binary.

include/common/core/spi.hpp
Laura-Danielle: Since both the spi and i2c devices only get built once, I think it's fine to keep this virtual.

include/common/core/spi.hpp
```diff
@@ -26,8 +26,8 @@ class TMC2130Spi {
      * @param transmit The transmit buffer.
      * @param receive The receive buffer.
      */
-    virtual void transmit_receive(const TMC2130Spi::BufferType& transmit,
-                                  TMC2130Spi::BufferType& receive) = 0;
+    virtual auto transmit_receive(const TMC2130Spi::BufferType& transmit,
```
Laura-Danielle, line 29: As we are already refactoring in this PR, I think it's time we call this thing something other than `TMC2130Spi` cus it's getting a bit confusing. We could instead call it `SpiDeviceBase` similar to `I2CDeviceBase`. For now, let's keep the `build_command` as is. If we find we need to change it, we can.

Laura-Danielle (approved):
In general looks good. Just had some nits. Also pls make sure the pipette run current is updated to something other than 0x2 everywhere in the project.

include/common/core/spi.hpp
ahiuchingau: I think `build_command` definitely needs to be updated in the future, but I don't care to do it in this PR. A good time to work on that might be when we refactor spi to make it its own task.

Laura-Danielle: Yep I agree. I think it would just be nice to at least change the name of this base class in this PR b/c we now have a class called TMC2130 and it's a bit confusing.

include/common/core/spi.hpp
sfoster1: I don't think I see an implementation of `TMC2130Spi::transmit_receive`, just in the subclasses. In that, case we definitely need to keep it pure virtual and add an `= 0`.

include/common/core/spi.hpp
sfoster1: I think that's a good idea - no code changes other than making it `class SpiBase`.

include/common/core/spi.hpp
sfoster1, line 30:
```suggestion
                                  TMC2130Spi::BufferType& receive) -> bool= 0;
```

sfoster1 (changes requested):
This is really good! I think it's probably a good idea to change the spi base class name like @Laura-Danielle suggests, but I don't mind if it's in another PR; but like @sotaoverride says we really need that `TMC2130Spi::transmit_receive` class to be pure virtual, not just virtual, if it does not have an implementation.

sfoster1 (approved):
LGTM once tested on hardware and passing tests

PR #263, opened 2022-02-25
feat(can): crc 32 checksum of entire fw update

amitlissack:
The download complete message now includes a crc32 of the entire application data.  We are using the HAL crc module to compute the crc. 

The crc is generated by python's `binascii.crc32`. There was a bit of monkeying around to get the HAL CRC to match the python generated.  This makes me a little nervous and I'm kind of wondering whether we should use a plain old checksum.

Anyway. It works. Tested on G4.

### DIFF
--- a/bootloader/firmware/crc32.c
+++ b/bootloader/firmware/crc32.c
@@ -2,6 +2,10 @@
 #include "platform_specific_hal_conf.h"
 #include "common/firmware/errors.h"
 
+/**
+ * Used this to match results from python's zlib.crc32.
+ * https://m0agx.eu/2021/04/09/matching-stm32-hardware-crc-with-standard-crc-32/
+ */
 
 /**
  * Handle to CRC module.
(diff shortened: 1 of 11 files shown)


bootloader/firmware/crc32.c
```diff
@@ -0,0 +1,69 @@
...
+void crc32_reset_accumulator() {
+    __HAL_CRC_DR_RESET(&hcrc);
+}
+
+/**
+ * Initialize the CRC unit.
+ */
+void MX_CRC_Init(void) {
+    hcrc.Instance = CRC;
+    hcrc.Init.DefaultPolynomialUse = DEFAULT_POLYNOMIAL_ENABLE;
```
Laura-Danielle, line 60: How did we decide on these CRC values?

Laura-Danielle (approved):
Had one question but looks good to me.

bootloader/firmware/crc32.c
amitlissack: I did what I could to have the fw match what Python produces.

sfoster1: Definitely looks good, the key to me is using a standard polynomial - everything else is just kind of matching bit ordering.

sfoster1 (approved):
lgtm!

PR #264, opened 2022-02-28
feat: enable IWDG in all applications

amitlissack:
This PR enables the IWDG. It expires after 1 second. When the watchdog expires the system will be reset.

A task is introduced that will refresh the watchdog timer every 950ms.

The bootloader detects when a restart occurred because of IWDG and does not automatically jump to the main application.

### DIFF
--- a/common/firmware/system/iwdg.c
+++ b/common/firmware/system/iwdg.c
@@ -1,0 +1,27 @@
+#include "platform_specific_hal_conf.h"
+#include "common/firmware/errors.h"
+#include "common/firmware/iwdg.h"
+
+
+static IWDG_HandleTypeDef hiwdg;
+
+
+void MX_IWDG_Init(void) {
+    /* USER CODE END IWDG_Init 1 */
+    hiwdg.Instance = IWDG;
+    hiwdg.Init.Prescaler = IWDG_PRESCALER_32;
+    hiwdg.Init.Window = IWDG_WINDOW_DISABLE;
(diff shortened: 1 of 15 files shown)


common/firmware/system/iwdg.c
```diff
@@ -0,0 +1,27 @@
...
+#include "common/firmware/iwdg.h"
+
+
+static IWDG_HandleTypeDef hiwdg;
+
+
+void MX_IWDG_Init(void) {
+    /* USER CODE END IWDG_Init 1 */
+    hiwdg.Instance = IWDG;
+    hiwdg.Init.Prescaler = IWDG_PRESCALER_32;
```
amitlissack, line 12: 32khz clock prescaled by 32 (1khz clock).

fsinapi:
Have you tested how the watchdog behaves when you hit a breakpoint in the debugger? Not sure if you need to set the `DBGMCU` registers to pause the timer during breakpoints

amitlissack:
> Have you tested how the watchdog behaves when you hit a breakpoint in the debugger? Not sure if you need to set the `DBGMCU` registers to pause the timer during breakpoints

I have. Sitting at a breakpoint doesn't trigger a watchdog reset.

sfoster1 (approved):
Nice!

I'm interested to see how this approach with a dedicated, prioritized task goes going forward - I'm more used to using the idle task for this. I think it'll be fine, just not quite what I'm used to.

PR #265, opened 2022-02-28
feat(sim): use boost for socket comms

amitlissack:


Laura-Danielle (approved):
Wow definitely simplifies things.

sfoster1 (approved):
Love boost

PR #267, opened 2022-03-02
feat(motor-control): adjustable motor current through CAN

ahiuchingau:
This PR allows us to change the TMC2130 driver motor currents via CAN in mA (in the form of a `uint32_t`)!

- Added new CAN message & ID for `WriteMotorCurrentRequest` that takes both the hold current (standstill) and the run current
- Updated motor driver task to handle the new can message
- Reconfigured the TMC2130 class so each subsystem can customize the driver's `R sense` and `Vsf` values that are crucial to motor current calculations


include/motor-control/core/tasks/motor_driver_task.hpp
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <cmath>
```
sfoster1, line 3: do we need these includes?

include/motor-control/core/tmc2130.hpp
```diff
@@ -126,6 +132,21 @@ class TMC2130 {
         return true;
     }
 
+    [[nodiscard]] auto convert_to_tmc2130_current_value(uint32_t c) const
+        -> uint32_t {
+        constexpr const float SMALL_R = 0.02;
+        const auto SQRT_TWO = static_cast<float>(std::sqrt(2));
```
sfoster1, line 138:
```suggestion
        constexpr auto SQRT_TWO = static_cast<float>(std::sqrt(2));
```

sfoster1 (approved):
awesome! maybe an include or two that isn't needed but this is great

include/motor-control/core/tmc2130.hpp
```diff
@@ -126,6 +132,21 @@ class TMC2130 {
...
+        auto FLOAT_CONSTANT = static_cast<float>(
+            SQRT_TWO * 32.0 * (_current_config.r_sense + SMALL_R) /
+            _current_config.v_sf);
+        auto fixed_point_constant = static_cast<uint32_t>(
+            FLOAT_CONSTANT * static_cast<float>(1LL << 16));
```
Laura-Danielle, line 143: we have functions that allow you to multiply and divide fixed point

ahiuchingau: Will add this unsigned fixed point multiply func to that file

PR #268, opened 2022-03-02
feat(can): add freertos task info request/response

amitlissack:
Add request to read FreeRTOS task info. The request results in a sequence of responses; one for each running task.

Using the https://freertos.org/uxTaskGetSystemState.html function we return each tasks:
- name
- current priority
- run time counter
- stack high water mark
- current state

This could be useful in fine tuning stack sizes, balancing priorites, etc.


sfoster1 (approved):
Awesome!

PR #270, opened 2022-03-03
feat(bootloader): erase flash message

amitlissack:
Rather than erasing the flash on the first data message, we will explicitly erase flash in response to a message.


sfoster1 (approved):
looks good and makes sense

PR #271, opened 2022-03-04
refactor(bootloader):  use interrupt version of flash erase

amitlissack:
Instead of using `HAL_FLASHEx_Erase` we are now using `HAL_FLASHEx_Erase_IT` to erase flash. 

`HAL_FLASHEx_Erase` was taking about 5 seconds which didn't give us an opportunity to reset the watchdog resulting in reset.

This PR enables flash interrupt and implements the complete and error callbacks called from the HAL Flash IRQ handler.

This has been tested on G4, but not L5.

### DIFF
--- a/bootloader/core/update_state.c
+++ b/bootloader/core/update_state.c
@@ -3,6 +3,7 @@
 static UpdateState update_state = {
     .num_messages_received=0,
     .error_detection=0,
+    .erase_state=erase_state_idle
 };
 
 
@@ -14,5 +15,6 @@ void reset_update_state(UpdateState * state) {
     if (state) {
         state->num_messages_received = 0;
         state->error_detection = 0;
+        state->erase_state = erase_state_idle;
     }
 }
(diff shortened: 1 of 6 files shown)


sfoster1 (changes requested):
Not sure but I don't think this usage of the flash state as a cross-execution-context signaling mechanism is safe, you could get fetches optimized out. In similar use cases in modules we use freertos task notifications; obviously that won't work here, but maybe this is a [c atomics to the rescue](https://en.cppreference.com/w/c/atomic) type of situation

amitlissack:
> Not sure but I don't think this usage of the flash state as a cross-execution-context signaling mechanism is safe, you could get fetches optimized out. In similar use cases in modules we use freertos task notifications; obviously that won't work here, but maybe this is a [c atomics to the rescue](https://en.cppreference.com/w/c/atomic) type of situation

Didn't know about c atomics so I'm happy to make a change, but for what it's worth this PR does work even in a release build.

sfoster1 (approved):
nice!

PR #272, opened 2022-03-07
refactor: Update run_emulation.yaml

DerekMaggio:
# Overview

Add basic sanity check action to run current version of ot3-firmware with emulation and make sure, it builds and runs

# Review Requests

* Is running on every PR and push too often? That is what it was before. It actually takes less time to run now, 4.5 minutes instead of 8
* I just put a super basic curl request sanity check, is there anything else we want to run?


DerekMaggio:
Will need to update opentrons-emulation github action refs to `main` after this PR

amitlissack (approved):

PR #273, opened 2022-03-07
refactor(motor-current): change ihold irun values for gantrys and heads

ahiuchingau:
Updating the gantry & head's irun value to `0x18` which sets the motor current to be roughly 1.4 A.


Laura-Danielle (approved):
Alt text

sfoster1 (approved):

PR #275, opened 2022-03-08
refactor(gantry): update motion system config

ahiuchingau:
Updating the numbers for steps/mm for Gantry X and Gantry Y based on the following values from HW:

> For X:
12.7mm pulley diameter
200 steps / revolution
1/32 microstepping
0.006234097922 mm per microstep

>For Y:
12.7254mm pulley diameter
200 steps / revolution
1/32 microstepping
0.006246566118 mm per microstep

This requires us to update the linear motion system belt config to take in `pulley diameter` instead of pulley tooth pitch. We also need to update the MRES value for the chopconf register from 0x4 to 0x3 to set microstep resolution at 32.


sfoster1 (approved):
nice, looks good to me!

PR #277, opened 2022-03-08
feat(motor control): reset position after home

caila-marashaj:


sfoster1 (changes requested):
needs a test for this one I think

PR #280, opened 2022-03-10
feat: add version int and git sha to get device info

sfoster1:
We need versioning in our firmware, and this adds it.

Specifically,
- Use a monotonically incrementing version that looks for tags like `v###` 
  - this is as opposed to a semver because this is not a separable component - it isn't something users should ever handle.
- Pull other useful flags and stuff from git, like the commit sha

Works with [github.com/opentrons/opentrons/pulls/9679](https://github.com/Opentrons/opentrons/pull/9679)


amitlissack (approved):
Lovely

PR #281, opened 2022-03-11
fix: simulator instability

amitlissack:
Address a number of flaky issues noticed when running tests.
- Our blocking socket reads were locking out message processing. We now poll the socket for available data before doing the blocking read. 
- FreeRTOS docs for posix port recommend using `nanosleep`. 
- `vTaskDelay` is a real cpu hog.  Simulators were at 99% cpu even when not doing nothing.
- As prescribed by FreeRTOS, disable all signals before a blocking call (like printf in logging).


sfoster1 (approved):
Looks good to me. Kind of weirded out by the high usage in idle, I'm not sure what should be causing that - at some point we should probably trace it or something

PR #282, opened 2022-03-11
feat(gripper): add support for brushed motor driver

ahiuchingau:
This PR allows us to set the Vref for the brushed motor driver to a default value using HAL_DAC.


sfoster1 (approved):
Looks good; I think in the future, this degree of commonality between the stepper and dc motor systems might bite us, but we can leave that for a future problem.

PR #283, opened 2022-03-11
fix(FreeRTOS): use Freertos systick

ahiuchingau:
This fixes HAL_Delay hanging issues by updating `SysTick_Handler` to (1) increment tick and (2) call the freeRTOS xPortSysTickHandler.


Laura-Danielle (approved):

sfoster1 (approved):
What's here looks good, but should we also do this for pipette and gripper?

ahiuchingau:
@sfoster1 I already added the changes to gripper in the other PR. And turns out the L5 board already took care of this [issue](https://github.com/STMicroelectronics/STM32CubeL5/blob/master/Middlewares/Third_Party/FreeRTOS/Source/portable/GCC/ARM_CM33_NTZ/non_secure/port.c#L736).

But I did open a ticket #284 to remind ourselves to make sure we update the pipettes subproject when we switch back to the G4 board.

PR #288, opened 2022-03-14
fix(motor-control): homing fixups

sfoster1:
- Set position to 0 after homing concludes
- Initialize the position tracker to a large positive value before homing starts, since homing will move negative and if we start the position at 0 and move negative we'll run into trouble

This is hardware tested and works great but we should probably have some unit testing

### DIFF
--- a/include/motor-control/core/motor_interrupt_handler.hpp
+++ b/include/motor-control/core/motor_interrupt_handler.hpp
@@ -168,6 +168,9 @@ class MotorInterruptHandler {
 
     void update_move() {
         has_active_move = queue.try_read_isr(&buffered_move);
+        if (buffered_move.stop_condition == MoveStopCondition::limit_switch) {
+            position_tracker = 0x7FFFFFFFFFFFFFFF;
+        }
         // (TODO: lc) We should check the direction (and set respectively)
         // the direction pin for the motor once a move is being pulled off the
         // queue stack. We'll probably want to think about moving the hardware
--- a/include/motor-control/core/motor_interrupt_handler.hpp
+++ b/include/motor-control/core/motor_interrupt_handler.hpp
@@ -106,7 +106,8 @@ class MotorInterruptHandler {
         }
         if (has_active_move) {
             if (buffered_move.stop_condition ==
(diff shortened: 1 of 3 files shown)


include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -136,6 +129,7 @@ class MotorInterruptHandler {
     auto homing_stopped() -> bool {
         if (limit_switch_triggered()) {
             finish_current_move(AckMessageId::stopped_by_condition);
+            position_tracker = 0;
```
caila-marashaj, line 132: I have a question about the reset function: does queue.reset() reset the entire queue, or just pop the current move off of it? If it's just the last move, could we just use the reset() in this file and start the next move?

sfoster1: Looks like it [removes all the messages](https://www.freertos.org/a00018.html#xQueueReset)

caila-marashaj: gotcha

motor-control/tests/test_limit_switch.cpp
```diff
@@ -91,6 +95,8 @@ TEST_CASE("Move with stop condition != limit switch") {
         test_objs.hw.set_mock_lim_sw(true);
         CHECK(!test_objs.handler.pulse());
         WHEN("the move is finished") {
+            REQUIRE(test_objs.handler.get_current_position() !=
```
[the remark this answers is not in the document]
caila-marashaj: yes, this check is just to ensure that the position doesn't get set to some arbitrary very high number at the beginning of any move that _isn't_ a home move though. Are you referring to the home completed in the other test case?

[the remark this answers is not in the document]
Laura-Danielle: I see. I didn't understand what this was trying to test upon looking at it. It might make more sense to verify that the position is equal to distance * 2 (for example) if you did two simultaneous non-homing moves back to back. Or, you could do a sequence of moves (such as non-home, home, non-home, non-home) to feel confident.

Laura-Danielle (approved):

PR #289, opened 2022-03-14
fix(head): Head detect bug

amitlissack:
`AttatchToolRequest` should always respond with attached tools. Regardless of whether there's a change from the last poll.


sfoster1 (approved):
Nice, looks correct to me

PR #291, opened 2022-03-14
Ret512 encoder firmware c configuration

Carlos-fernandez:
This PR configures the motor encoders for Z axis and A axis on the Head of the OT3. The current configuration enables TIMER2 and TIMER3 for the motors to read the encoders as inputs. This currently works with svd by looking at the CNT register on the TIM2 and TIM3.


motor-control/firmware/motor_encoder_hardware.h
```diff
@@ -0,0 +1,19 @@
+
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif  // __cplusplus
+
+#include <stdbool.h>
```
amitlissack, line 8: I can't see a use of `bool` type. Are you sure this include is needed?

Carlos-fernandez: you're right, I removed it. I don't need it. Thanks!

motor-control/firmware/motor_encoder_hardware.c
```diff
@@ -0,0 +1,48 @@
...
+
+#include "FreeRTOS.h"
+#include "task.h"
+
+#include "platform_specific_hal_conf.h"
+
+void motor_encoder_set_pin(void* port, uint16_t pin, uint8_t active_setting)
+{
+    if(!port) {return;}
+    // Hall Function here
```
sfoster1, line 12: let's avoid comments like this that restate exactly what the next line does or are notes to self. A typical rule is that comments are 
- in the general case, about _why_ you're doing it
- in some cases where the _what_ is weird, the what, but this should be pretty rare (HAL interaction is a pretty normal place to have a "what" comment, like "configure this timer for encoder input per manual section xx.yy.zz"
- in extreme cases, _how_ you're doing something, but usually it's better to figure out a way to do something that doesn't need a how comment in the first place. 

An example is this function: https://github.com/Opentrons/ot3-firmware/blob/4698d1eb0f01f991d476cf8e575e073e4ea7f2c9/bootloader/firmware/updater.c#L190-L219 which is what's responsible for moving from bootloader to application.
- It maybe could use a "why" comment, but given that the function is called `fw_update_start_application` you can kind of see what it does
- the "what" comment about the systick is good because those are kind of cryptic calls; same with the MSP setting
- The assembler snippet has "what" and "why"; it could also have "how", but at that point it would just be restating arm thumb assembler docs since it's literally one assembler directive

motor-control/firmware/motor_encoder_hardware.c
```diff
@@ -0,0 +1,48 @@
...
+        case GPIO_PIN_SET:
+            return GPIO_PIN_RESET;
+        case GPIO_PIN_RESET:
+            return GPIO_PIN_SET;
+        default:
+            return GPIO_PIN_SET;
+    }
+}
+
+void motor_encoder_reset_pin(void* port, uint16_t pin, uint8_t active_setting) {
```
sfoster1, line 27: do we need this function, since these are all input pins?

head/firmware/motor_hardware.c
```diff
@@ -189,6 +194,103 @@ void MX_GPIO_Init(void) {
...
+void MX_TIM2_Init(void){
+
+    TIM_Encoder_InitTypeDef sConfig = {0};
+    TIM_MasterConfigTypeDef sMasterConfig = {0};
+    TIMEx_EncoderIndexConfigTypeDef sEncoderIndexConfig = {0};
+    __HAL_RCC_TIM2_CLK_ENABLE();
+    htim2.Instance = TIM2;
+    htim2.Init.Prescaler = 0;
+    htim2.Init.CounterMode = TIM_COUNTERMODE_UP;
+    htim2.Init.Period = 4.294967295E9;
```
sfoster1, line 218: this is autogenerated, right? this is a very odd way to write this - i assume what it's getting at is "the largest value you can have in a 32 bit register". Instead, we can use the macro `UINT32_MAX`, defined in [stdint.h](https://en.cppreference.com/w/c/types/integer)
```suggestion
    htim2.Init.Period = UINT32_MAX;
```

head/firmware/motor_hardware.c
```diff
@@ -189,6 +194,103 @@ void MX_GPIO_Init(void) {
...
+void MX_TIM3_Init(void){
+
+    TIM_Encoder_InitTypeDef sConfig = {0};
+    TIM_MasterConfigTypeDef sMasterConfig = {0};
+    TIMEx_EncoderIndexConfigTypeDef sEncoderIndexConfig = {0};
+    __HAL_RCC_TIM3_CLK_ENABLE();
+    htim3.Instance = TIM3;
+    htim3.Init.Prescaler = 0;
+    htim3.Init.CounterMode = TIM_COUNTERMODE_UP;
+    htim3.Init.Period = 65535;
```
sfoster1, line 262: similarly, we could use `UINT16_MAX` here, but i find `65535` equally legible - up to you
```suggestion
    htim3.Init.Period = UINT16_MAX;
```

head/firmware/motor_hardware.c
```diff
@@ -189,6 +194,103 @@ void MX_GPIO_Init(void) {
     GPIO_InitStruct.Pull = GPIO_NOPULL;
     GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
     HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
+
+    /*Configure GPIO pin : B1_Pin */
```
sfoster1, line 198: What are we doing with this pin? We don't seem to be using it anywhere else

head/firmware/motor_hardware.c
```diff
@@ -181,6 +184,8 @@ void MX_GPIO_Init(void) {
     __HAL_RCC_GPIOF_CLK_ENABLE();
     __HAL_RCC_GPIOA_CLK_ENABLE();
     __HAL_RCC_GPIOB_CLK_ENABLE();
+    __HAL_RCC_GPIOD_CLK_ENABLE();
```
sfoster1, line 187: what do we use portd for?

sfoster1 (changes requested):
Some stuff to change, but overall looks good.

One thing that's not in the inline comments - where do we initialize the pins that the encoder is connected to, A0/1 and A6/7? I don't see it here, is there a missing file?

head/firmware/motor_hardware.c
Carlos-fernandez: that value was autogenerated but the cubemx software. That looks better.

motor-control/firmware/motor_encoder_hardware.c
Carlos-fernandez: nope we don't need this.

head/firmware/motor_hardware.c
Carlos-fernandez: we need port D for ENCI_Z_Axis pin. That's the only pin on port D. That's why i enabled the CLK on this port.

head/firmware/motor_hardware.c
```diff
@@ -189,6 +194,103 @@ void MX_GPIO_Init(void) {
...
     HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
+
+    /*Configure GPIO pin : B1_Pin */
+    GPIO_InitStruct.Pin = GPIO_PIN_13;
+    GPIO_InitStruct.Mode = GPIO_MODE_IT_RISING;
+    GPIO_InitStruct.Pull = GPIO_NOPULL;
+    HAL_GPIO_Init(GPIOC, &GPIO_InitStruct);
+
+    /* EXTI interrupt init*/
+    HAL_NVIC_SetPriority(EXTI15_10_IRQn, 0, 0);
```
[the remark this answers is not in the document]
Carlos-fernandez: this was autogenerated from the CUBEMX software. Removed

motor-control/firmware/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 function(target_ot_motor_control_firmware TARGET)
   target_sources(${TARGET} PUBLIC
     ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/motor_control_hardware.c
+    #{CMAKE_CURRENT_FUNCTION_LIST_DIR}/motor_encoder_hardware.c
```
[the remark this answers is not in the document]
Carlos-fernandez: removing these files for this pull request. This is silly of me to do.

head/firmware/motor_hardware.c
Carlos-fernandez: removed, not being used.

Carlos-fernandez:
I added the GPIO pin configuration for the motor encoders this time. That way it's more clear that the pins were defined.

PR #294, opened 2022-03-16
fix(pipettes): prevent I2C device object from being destroyed too soon

Laura-Danielle:
## Overview

The i2c device was being created inside the main function which gets destroyed by the time the i2c
task wants to use that device. Tested on hardware.


common/firmware/i2c/i2c_comms.cpp
```diff
@@ -27,4 +26,8 @@ auto I2C::central_receive(uint8_t* data, uint16_t size, uint16_t dev_address,
     return hal_i2c_master_receive(handle, dev_address, data, size, timeout);
 }
 
-auto I2C::wait_during_poll(uint16_t delay) -> void { vTaskDelay(delay); }
\ No newline at end of file
+auto I2C::wait_during_poll(uint16_t delay) -> void { vTaskDelay(delay); }
+
+auto I2C::set_handler(HAL_I2C_HANDLE i2c_handle) -> void {
```
amitlissack, line 31: Minor nit. It should be called `set_handle` instead, no?

amitlissack (approved):
Classic  C/C++ learning opportunity!!! 

A minor thing (that I am ambivalent about) is that we have tended to make our HAL_XXX_handles global. If that were the case you could still construct the `I2C` class with a handle.

pipettes/firmware/main.cpp
```diff
@@ -131,12 +132,9 @@ auto main() -> int {
     adc_init();
     auto id = pipette_mounts::detect_id();
 
-    // TODO(lc 01/04/2021): We should move this setup to another object
-    // responsible for sensor control.
-    auto i2chandler_struct = I2CHandlerStruct{};
+    I2CHandlerStruct i2chandler_struct{};
```
sfoster1, line 135: This is gonna have exactly the same problem - `i2chandler_struct` also must be on the free store rather than main's stack, unless it is copied to somewhere else.

sfoster1 (changes requested):
i2 chandler has to be on the free store also I think

sfoster1 (approved):
nice

PR #295, opened 2022-03-17
fix(pipettes): use correct i2c bus

Laura-Danielle:
## Overview
The main sensors were actually on i2c3 bus which would explain why nothing was responding on bus 1.
The schematic labels were incorrect 😭


sfoster1 (approved):
Big Oof

PR #297, opened 2022-03-18
fix(pipettes): fix typo in sensor CAN messages and change data type

Laura-Danielle:
## Overview
Can message typo was causing the message to not send over all of the expected data. Also changed the
type of the data to support positive or negative values.


sfoster1 (approved):
Nice. Might want an integration test for this at some point

amitlissack (approved):
👍

Laura-Danielle:
> Nice. Might want an integration test for this at some point

Yep working on that for today.

PR #298, opened 2022-03-18
ZA Encoder Pulse Counter and reset Pulse Counter

Carlos-fernandez:
This PR adds a few things to the motor hardware of the head.
-Motor Encoder pulse count can now be retrieved
-A reset Motor Encoder pulse count
-Separated the Encoder GPIO Configuration from MX_GPIO_Init() function. 

These changes were tested with gdb with setting breaking points and reading the enc_pulse_counter when rotating the motor manually.

### DIFF
--- a/head/firmware/motor_hardware.c
+++ b/head/firmware/motor_hardware.c
@@ -287,8 +287,6 @@ void MX_TIM2_Init(void){
     {
     Error_Handler();
     }
-    HAL_NVIC_SetPriority(TIM3_IRQn, 4, 0);
-    HAL_NVIC_EnableIRQ(TIM3_IRQn);
     HAL_TIM_Encoder_Start_IT(&htim2, TIM_CHANNEL_ALL);
 }
 
@@ -333,8 +331,6 @@ void MX_TIM3_Init(void){
     {
     Error_Handler();
     }
-    HAL_NVIC_SetPriority(TIM3_IRQn, 4, 0);
-    HAL_NVIC_EnableIRQ(TIM3_IRQn);
     HAL_TIM_Encoder_Start_IT(&htim3, TIM_CHANNEL_ALL);
 }
 
@@ -375,7 +371,7 @@ void HAL_TIM_Base_MspInit(TIM_HandleTypeDef *htim) {
     }
 }
 
-uint16_t encoder_pulse_count(TIM_HandleTypeDef *htim){
+uint32_t encoder_pulse_count(TIM_HandleTypeDef *htim){
     if (htim == &htim2){
         enc_pulse_counter = __HAL_TIM_GET_COUNTER(&htim2);
     }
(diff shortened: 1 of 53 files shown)


head/firmware/motor_hardware.c
```diff
@@ -356,11 +375,32 @@ void HAL_TIM_Base_MspInit(TIM_HandleTypeDef *htim) {
     }
 }
 
+uint16_t encoder_pulse_count(TIM_HandleTypeDef *htim){
```
amitlissack, line 378: `enc_pulse_counter` is `uint32_t`.

head/firmware/motor_hardware.c
```diff
@@ -5,6 +5,7 @@
 TIM_HandleTypeDef htim7;
 TIM_HandleTypeDef htim2;
 TIM_HandleTypeDef htim3;
+uint32_t enc_pulse_counter;
```
amitlissack, line 8: Is this variable meant to be accessible outside of this module using `extern`? If not, can you please make it `static`

head/firmware/motor_hardware.c
```diff
@@ -356,11 +375,32 @@ void HAL_TIM_Base_MspInit(TIM_HandleTypeDef *htim) {
...
+    if (htim == &htim2){
+        enc_pulse_counter = __HAL_TIM_GET_COUNTER(&htim2);
+    }
+    else if(htim == &htim3){
+        enc_pulse_counter = __HAL_TIM_GET_COUNTER(&htim3);
+    }
+    return enc_pulse_counter;
+}
+
+void reset_encoder_counter(TIM_HandleTypeDef *htim){
```
amitlissack, line 388: I don't see where this function is being called.

sfoster1: it's not being called anywhere - this is a first implementation intended to work basically only under a debugger, and make sure we got the hardware handling stuff right. it may get discarded though, we might want to call it as a dummy thing from somewhere

head/firmware/motor_hardware.c
```diff
@@ -274,6 +287,9 @@ void MX_TIM2_Init(void){
     {
     Error_Handler();
     }
+    HAL_NVIC_SetPriority(TIM3_IRQn, 4, 0);
```
sfoster1, line 290: We're not using these interrupts, are we? We shouldn't enable them

head/firmware/motor_hardware.c
```diff
@@ -317,6 +333,9 @@ void MX_TIM3_Init(void){
...
```
sfoster1, line 336: Shouldn't enable these interrupts

sfoster1 (changes requested):
We shouldn't enable interrupts we're not using

head/firmware/motor_hardware.c
Carlos-fernandez: yep, that variable is meant to be accessible outside of the module. Add this to motor_hardware.h as `extern enc_pulse_counter;` ?

sfoster1 (changes requested):

PR #300, opened 2022-03-21
feat(pipettes): add pipette info messages

sfoster1:
The host needs to be able to query what kind of pipette is attached so
it can display this information to the user, load appropriate settings,
and similar. This information doesn't really belong in the device info
because it's only present on pipettes.

This PR carves out messages and enums for the information and hooks up a
basic message handler that hardcodes a response. In the future, this
information will be either compiled in or stored on the EEPROM.


amitlissack (approved):

Laura-Danielle (approved):
Thanks! And looks great :)

PR #302, opened 2022-03-22
feat(gripper): generate pwm signals

ahiuchingau:
We're using two separate timers to generate two PWM signals to drive the brushed motor. One line is connected to the PC0 pin (TIM1) and the other PA6 (TIM3). 

Next steps are to implement these features in the motor control module and allow to adjust the Vref, duty cycle and test for and use the optimum PWM frequency.

### DIFF
--- a/gripper/firmware/stm32g4xx_it.c
+++ b/gripper/firmware/stm32g4xx_it.c
@@ -45,6 +45,8 @@
 DMA_HandleTypeDef hdma_spi1_tx;
 DMA_HandleTypeDef hdma_spi1_rx;
 extern TIM_HandleTypeDef htim7;
+extern TIM_HandleTypeDef htim1;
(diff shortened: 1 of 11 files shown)


gripper/firmware/stm32g4xx_it.c
```diff
@@ -150,4 +155,9 @@ void SysTick_Handler(void) {
     HAL_IncTick();
     xPortSysTickHandler();
 }
+
+/**
+ * @brief This function handles EXTI line[15:10] interrupts.
+ */
+void EXTI15_10_IRQHandler(void) { HAL_GPIO_EXTI_IRQHandler(GPIO_PIN_13); }
```
sfoster1, line 162: what's this for?

sfoster1 (approved):
Looks good though I think some unused interrupt defs have snuck in there

PR #303, opened 2022-03-24
fix(pipettes): specify the individual targets in cmake presets

Laura-Danielle:
## Overview

Do not compile all of the pipettes when you want to only compile one target.


DerekMaggio (approved):
Looks pretty straightforward to me

PR #304, opened 2022-03-24
feat(*): real position in move-complete

sfoster1:
We've always been sending the current position back after a move
completes, but it's also always been in steps. That isn't a useful unit
when the host doesn't know the steps/mm of each device.

Instead, let's use our knowledge of the mechanical configuration to
translate the current position from steps to micrometers before sending
it back over the canbus.


Laura-Danielle (approved):
lgtm

amitlissack (approved):

PR #305, opened 2022-03-25
chore: Add workflow dispatch, remove push to run_emulation workflow

DerekMaggio:
Add `workflow_dispatch` remove `push` from `run_emulation` workflow


sfoster1 (approved):
Approved but please make sure that what goes into the git log when you merge this mentions which workflow it's talking about

PR #307, opened 2022-03-28
added pipette, gantry hardware configurations

caila-marashaj:


sfoster1 (changes requested):
The `sync_pin` function looks good, but I don't really love the way the `spi_pins_[b,c,d]` functions work. By defining the function this way, you're relying on the programmer to call the right function and setting aside the tools the language has to make it impossible to call it wrong - no static analysis can catch that you used `spi_pins_b` to initialize `GPIOD`.

Instead, can we have just one `spi_pins` function that takes a gpio handle as an argument and switches on it? If we do that it's a lot easier to check.


Something like this:
```c
uint8_t spi_pins(GPIO_Handle* for_handle) {
    switch(for_handle->Instance) {
      case GPIOB: return GPIO_PIN_13 | GPIO_PIN_14 | GPIO_PIN_15;
      case GPIOC: // etc
      default: return 0;
    }
}
```

gantry/firmware/axis_x_hardware_config.c
```diff
@@ -0,0 +1,17 @@
+#include "axis_hardware_config.h"
+
+GantryHardwarePin gantry_hardware_get_gpio(GantryHardwareDevice device) {
+    GantryHardwarePin pinout;
+    switch(device) {
+        case gantry_hardware_device_LED_drive:
```
sfoster1, line 6: This will fall through every time and always return GPIOB pin 7, needs a `break` statement in each case or a `return`. The compiler will then complain that there's a path to returning an uninitialized value, so you'll need a `default`.

gantry/firmware/axis_y_hardware_config.c
```diff
@@ -0,0 +1,17 @@
+#include "axis_hardware_config.h"
+
+GantryHardwarePin gantry_hardware_get_gpio(GantryHardwareDevice device) {
+    GantryHardwarePin pinout;
+    switch(device) {
+        case gantry_hardware_device_LED_drive:
```
sfoster1, line 6: This will fall through every time and always return GPIOB pin 7, needs a break statement in each case or a return. The compiler will then complain that there's a path to returning an uninitialized value, so you'll need a default.

pipettes/firmware/high_throughput_config.c
```diff
@@ -0,0 +1,29 @@
+#include "hardware_config.h"
+#include "stm32l5xx_hal_gpio.h"
+
+
+PipetteHardwarePin pipette_hardware_get_gpio(PipetteHardwareDevice device) {
+    PipetteHardwarePin pinout;
+    switch(device) {
```
sfoster1, line 7: check fallthrough

pipettes/firmware/low_throughput_config.c
```diff
@@ -0,0 +1,28 @@
+#include "hardware_config.h"
+#include "stm32l5xx_hal_gpio.h"
+
+
+PipetteHardwarePin pipette_hardware_get_gpio(PipetteHardwareDevice device) {
+    PipetteHardwarePin pinout;
+    switch(device) {
+        case pipette_hardware_device_limit_switch:
```
sfoster1, line 8: check fallthrough

sfoster1 (changes requested):
These new `switch` statements don't work, they'll fallthrough all the way. Other than that, looks good

sfoster1 (approved):
Looks good to me! Nice work

PR #308, opened 2022-03-29
fix(pipettes): Fix builds and workflows

sfoster1:
The commit that added the different pipette build types also removed the
`pipette` CMake target, so now cmake silently does nothing in the
pipette build workflows, which means we didn't notice accidentally
breaking the pipette build.

Add the target back to cmake, fix the build issure, and update the
workflow to use presets.


Laura-Danielle (approved):
lgtm

PR #310, opened 2022-03-29
feat(pipettes): add pressure sensor

Laura-Danielle:
## Overview

This adds the pressure to the pipettes project.

## Review Requests

This is still a WIP. I need to finish handling the data storage of a given register once the data is received back. I would like some eyes on:
1. How the callbacks are handled (is it really bad that they're on the driver class now?)
2. The general structure of the sensor driver.

Will test on hardware once I finish the data handling part.


include/pipettes/core/i2c_writer.hpp
```diff
@@ -33,9 +33,9 @@ class I2CWriter {
 
     template <typename Data>
     requires std::is_integral_v<Data>
-    void write(Data data, uint16_t device_address) {
+    void write(Data data, uint16_t device_address, uint8_t reg = 0x0) {
```
amitlissack, line 36: Why a default of `0x0`? Why any default?

include/sensors/core/mmr920C04.hpp
```diff
@@ -0,0 +1,281 @@
...
+#include "sensors/core/sensors.hpp"
+
+namespace mmr920C04 {
+
+using namespace can_ids;
+using namespace mmr920C04_registers;
+using namespace sensors_registers;
+
+template <class I2CQueueWriter, message_writer_task::TaskClient CanClient>
+class MMR92C04 {
```
amitlissack, line 18: Looks like we do this in other instances, so if I'm alone in thinking that naming classes and namespaces solely after a part is alienating just ignore me.

include/sensors/core/mmr920C04.hpp
```diff
@@ -0,0 +1,281 @@
...
+            case Registers::MACRAM_WRITE:
+                break;
+        }
+    }
+
+  private:
+    MMR920C04RegisterMap _registers{};
+    bool _initialized = false;
+    int32_t threshold = 0x8;
+    uint16_t DELAY = 20;
```
amitlissack, line 269: Is this a constant?

include/sensors/core/mmr920C04.hpp
```diff
@@ -0,0 +1,281 @@
...
+            case Registers::MEASURE_MODE_4:
+            case Registers::MACRAM_WRITE:
+                break;
+        }
+    }
+
+  private:
+    MMR920C04RegisterMap _registers{};
+    bool _initialized = false;
+    int32_t threshold = 0x8;
```
amitlissack, line 268: Is this a constant? What is 8? What are the units?

include/sensors/core/mmr920C04.hpp
```diff
@@ -0,0 +1,281 @@
...
+    auto poll_read(Registers reg, uint16_t number_reads) -> void {
+        writer.single_register_poll(
+            ADDRESS, number_reads, DELAY,
+            [this, reg]() { send_to_can(this, reg); },
+            [this, reg](auto message_a) { handle_data(message_a, this, reg); },
+            static_cast<uint8_t>(reg));
+    }
+
+    auto write_config() -> bool {
+        if (!reset(_registers.reset)) {
```
amitlissack, line 62: If these functions fail should `_initialized` become false?

include/sensors/core/mmr920C04.hpp
```diff
@@ -0,0 +1,281 @@
...
+            return false;
+        }
+        if (!set_measure_mode4(_registers.measure_mode_4)) {
+            return false;
+        }
+        _initialized = true;
+        return true;
+    }
+
+    auto set_measure_mode1(MeasureMode1 reg) -> bool {
```
amitlissack, line 72: Feels like there could be one public `set_measure_mode` method.

include/sensors/core/mmr920C04.hpp
```diff
@@ -0,0 +1,281 @@
...
+
+    auto set_measure_mode4(MeasureMode4 reg) -> bool {
+        if (set_register(reg)) {
+            _registers.measure_mode_4 = reg;
+            return true;
+        }
+        return false;
+    }
+
+    auto get_pressure(bool poll = false, uint16_t sample_rate = 0) -> void {
```
amitlissack, line 104: These three get_pressure methods look about the same. Can there be a function that takes the register as input and reduces the duplication of logic?

include/sensors/core/mmr920C04_registers.hpp
```diff
@@ -0,0 +1,313 @@
...
+            default:
+                return SensorStatus::UNKNOWN;
+        }
+    }
+};
+
+struct MMR920C04RegisterMap {
+    Reset reset = {};
+    Idle idle = {};
+    MeasureMode1 measure_mode_1 = {};
```
amitlissack, line 298: What are the differences between these measuring modes?

amitlissack (commented):
Looks good. Just some code style and naming questions.

common/simulation/i2c_sim.cpp
```diff
@@ -1,16 +1,22 @@
...
     uint8_t reg = data[0];
     uint8_t write_bit = data[1];
+    LOG("Received a register bit: %d", static_cast<int>(reg));
+
     if (write_bit == 1) {
+        LOG("Received a write bit");
         uint16_t store_in_register = 0;
         auto *iter = data + 1;
         iter = bit_utils::bytes_to_int(iter, data + size, store_in_register);
+        LOG("Storing %d in register %d", static_cast<int>(store_in_register),
```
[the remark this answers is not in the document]
Laura-Danielle: Probably not for logging. I will double check. They are necessary in `std::cout` because it can't interpret unsigned ints.

include/pipettes/core/i2c_writer.hpp
Laura-Danielle: For now I have a default until I've modified the other sensors to match the driver refactor since the `write` command previously did not take in a register. I will add a TODO comment.

include/sensors/core/mmr920C04.hpp
Laura-Danielle: Yeah generally we name the "drivers" the part number, but I can add an additional comment in this file to say that this is a pressure driver class.

include/sensors/core/mmr920C04.hpp
Laura-Danielle: This is not a constant, it can be set from CAN. I'll put the units in a comment.

amitlissack: Better to name the variable with the unit.

include/sensors/core/mmr920C04.hpp
amitlissack: Then type it so.

include/sensors/core/mmr920C04_registers.hpp
Laura-Danielle: It affects the low pass filter option of pressure readings

include/pipettes/core/i2c_writer.hpp
```diff
@@ -33,9 +33,11 @@ class I2CWriter {
...
```
sfoster1, line 36: I think that rather than doing this at the level of the i2c writer, it's better to have some independent data formatters in a utility. Like you've noticed here, almost - _almost_ - every single i2c device has a register model where "read-register" transactions look like "WRITE regaddr + stop READ ", and write-register transactions look like "WRITE regaddr+continue DATA...", but not quite all of them do. So I think having some `namespace i2c_registers` that has functions like `auto build_register_write_message(uint8_t regaddr, std::ranges Data) -> MaxMessageBuffer;` and `auto build_register_read_message(uint8_t reg) -> MaxMessageBuffer` or whatever, and you can pass the result of those into i2c write.

sfoster1: Or actually another good way to do this would be to have a second overload of `write`: `void write(uint16_t device_address, uint8_t reg, Data data)` that lives alongside this one

include/sensors/core/mmr920C04.hpp
```diff
@@ -0,0 +1,275 @@
+#pragma once
```
sfoster1, line 1: You mentioned in your PR comment that you were worried about putting some of this functionality here. I kind of agree with that assessment, but honestly I think it's about the naming. This does what `capacitive_sensor_callbacks.hpp` does, not what `hdcwhatever` does. I think a file named like this should just have conversion and utility functions with little state, and have something that's more closely associated with the task to handle the actual data routing. So moving this to `sensors/tasks` and putting the actual conversion functions back into `sensors/core` seems perfect.

include/sensors/core/tasks/capacitive_sensor_task.hpp
```diff
@@ -32,7 +32,7 @@ class CapacitiveMessageHandler {
     }
 
     void initialize() {
-        writer.write(DEVICE_ID_REGISTER, ADDRESS);
+        writer.write(0x0, ADDRESS, DEVICE_ID_REGISTER);
```
sfoster1, line 35: this sort of adds to what I mean about the comments about `write` above - this is _really_ weird looking, and it would be nice if this could use either the old signature or a new signature for like `prep_read` that means you don't have to send data that will always be 0.

sfoster1 (changes requested):
I think some restructuring of the i2c changes will help a little. Also one of us is going to have some brutal conflicts

include/pipettes/core/i2c_writer.hpp
```diff
@@ -34,13 +34,27 @@ class I2CWriter {
     template <typename Data>
     requires std::is_integral_v<Data>
     void write(Data data, uint16_t device_address) {
+        // TODO do not make register default once all of the sensor
```
sfoster1, line 37: out of date comment

sfoster1 (approved):
Looks good, one leftover comment tho

PR #311, opened 2022-03-29
feat(can): add gripper messages to adjust vref, pwm, movements

ahiuchingau:
Making the gripper grip.

Relative changes in `opentrons_hardware` [here](https://github.com/Opentrons/opentrons/pull/9841)


include/can/core/messages.hpp
```diff
@@ -572,6 +572,44 @@ struct SensorThresholdResponse
...
+        body = bit_utils::bytes_to_int(body, limit, duty_cycle);
+        return SetBrushedMotorPwmRequest{.freq = freq,
+                                         .duty_cycle = duty_cycle};
+    }
+
+    auto operator==(const SetBrushedMotorPwmRequest& other) const
+        -> bool = default;
+};
+
+using GripperGripRequest = Empty<MessageId::gripper_grip_request>;
```
sfoster1, line 609: we should parameterize this with a `force_mn` (millinewtons, uint32_t) value and maybe an `expected_grip_width_um` (micrometers, uint32_t) value too

sfoster1 (approved):
Looks good, we should probably change how we parameterize that grip message but also we can do it later

sfoster1:
Pretty awesome!

PR #313, opened 2022-03-31
Ret 668 query sync in motor interrupt

caila-marashaj:
- added a check inside `motor_interrupt_handler()` that finishes the current move if the sync line is high and the stop condition is `cap_sensor`
- added a test file for this


sfoster1 (approved):
looks good to me!

PR #316, opened 2022-04-04
feat(pipette): add sensor bindings

sfoster1:
Add the ability for the i2c capacitive sensor attached to the pipette to control the global system sync line, which we can use to stop motion when that sensor triggers - important for automatic calibration. At some point we'll need the same thing for the pressure sensor for liquid level sensing, but it should be easier to add.

There is... a lot here. The core reason for that is twofold:
1. The way we did polling seemed like it would be a little hard to extend, and extending it in place would leave things messy
2. This is the first thing outside motor control that needs gpios, so those had to be generalized
3. The way we were handling responses to i2c messages actually took advantage of undefined behavior. Messages passed through freertos queues are moved around via memcpy. This is not allowed for types that are not [TriviallyCopyable](https://en.cppreference.com/w/cpp/named_req/TriviallyCopyable), and [closure types](https://en.cppreference.com/w/cpp/language/lambda) generated for lambdas and held in std::functions need not be. That means that every time you memcpy a `std::function` with a closure type, you're in UB-land, which segfaults at least in simulators. The fix for this was to switch to responding with messages, which is quite a lot of code.
4. While doing all the above, the organization was getting to me. 

So, the overall set of changes looks like:
- I2C read results now are communicated via response messages to the commanding task. To accommodate this, read-request messages have a handle on a queue that the sender specifies that will get responses.
- Read-requests and responses now share a struct that identifies a transaction; this is specified by the caller as whatever it needs to be, with some handy additional members but mostly relying on a uint32_t.
- Added a transact-request method (and this is the only actual message) that does a write and then immediate read, to match the usual needs of i2c communication
- Added a new "polling task", which hoists the polling out of the i2c task and into its own, and triggers it with timers. There is a simiilar interface to the i2c writer for the poller; and responses work similarly, with response messages forwarded upstream.
- Other ancillary refactors as needed: timers now have their periods set by methods, not by template arguments, so they can be rescheduled; queues can handle any convertible message
- Sensors adapted to the i2c changes, as the major consumers of the i2c interface
- The actual changes to enable setting the sync line - continuous polling of sensors and hardware capture

## Reviewing
The best way to review this is probably to approach it as if this is the initial commit for everything in `include/i2c` and `i2c/`. Start with `include/i2c/core/tasks/i2c_task.hpp` and `include/i2c/core/writer.hpp`, going over to `include/i2c/core/messages.hpp`. After that, take a look at the poller: `include/i2c/core/tasks/i2c_poller_task.hpp`, `include/i2c/core/poller.hpp`, `include/i2c/core/poller_impl.hpp`. These have tests in `i2c/tests/`. Then you can move to the sensor changes, and finally the integration into the pipette main in firmware and simulation.

## Testing
This is tested in simulation but needs to be tested in hardware before merge. Testing should include
- ~Baselining~
- ~Reading~
- ~Sync-binding~
The core functionality works in hardware but some hardware bindings about which busses to use need to change and I really want to do that in a followup pr. this doesn't break any functionality but it does continue what's supposed to work not quite working as designed.


pipettes/simulator/main.cpp
```diff
@@ -54,6 +55,8 @@ std::map<uint16_t, sensor_simulator::SensorType> sensor_map = {
 
 static auto i2c3_comms = sim_i2c::SimI2C{sensor_map};
 
+static sensor_hardware::SimulatedSensorHardware fake_sensor_hw{};
```
Laura-Danielle, line 58: can we call this `sim_sensor_hw` instead?

pipettes/simulator/main_multi_i2c.cpp
```diff
@@ -55,6 +56,8 @@ std::map<uint16_t, sensor_simulator::SensorType> sensor_map = {
 static auto i2c3_comms = sim_i2c::SimI2C{sensor_map};
 static auto i2c1_comms = sim_i2c::SimI2C{sensor_map};
 
+static sensor_hardware::SimulatedSensorHardware fake_sensor_hw{};
```
Laura-Danielle, line 59: same here

Laura-Danielle (approved):
looks awesome!

include/common/core/timer.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+template <typename TimerType>
+concept Timer = requires(TimerType tt, Callback cb) {
+    // Query that the timer is running
+    { tt.is_running() } -> std::same_as<bool>;
+    {tt.update_callback(std::move(cb))};
+    {tt.update_period(static_cast<uint32_t>(10))};
+    {tt.start()};
+    {tt.stop()};
+    {TimerType(
```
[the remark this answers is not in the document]
sfoster1: yeah that's correct - have to be able to give it a name, a callback, and a time

include/common/tests/mock_message_queue.hpp
```diff
@@ -59,6 +74,6 @@ class MockMessageQueue {
     uint8_t get_size() { return queue_data_structure.size(); }
 
   private:
-    std::vector<Message> queue_data_structure;
+    std::deque<Message> queue_data_structure;
```
Laura-Danielle, line 77: deque always feels fancy to me lol

Laura-Danielle (approved):
Looks great to me

caila-marashaj (approved):

PR #317, opened 2022-04-05
Adding Encoder Position to Move Complete

Carlos-fernandez:
This PR adds the encoder pulse count to move complete. The move complete returns a addition line for the encoder position. I have tested this method with the can_mon.py script that returns a value. This firmware works on all the axes. In the future we can convert the pulse counts to mm for prediction or make a pulses/mm. 

Note:  Currently the robot doesn't have encoders on XY axes, we can remove the condition that controls this in the future.


include/motor-control/core/motor_interrupt_handler.hpp
```diff
@@ -241,6 +247,7 @@ class MotorInterruptHandler {
 
   private:
     uint64_t tick_count = 0x0;
+    // uint32_t pulses = 0x0;
```
sfoster1, line 250: remove commented out code

sfoster1 (changes requested):
We should fix up the existing tests and add a new test for the encoder pulse functionality but this is looking great!

can/tests/test_messages.cpp
```diff
@@ -49,8 +49,9 @@ SCENARIO("message serializing works") {
         auto message = MoveCompleted{.group_id = 1,
                                      .seq_id = 2,
                                      .current_position_um = 0x3456789a,
+                                     .encoder_position = 0x05803931,
                                      .ack_id = 1};
-        auto arr = std::array<uint8_t, 8>{};
+        auto arr = std::array<uint8_t, 11>{};
```
sfoster1, line 54:
```suggestion
        auto arr = std::array<uint8_t, 12>{};
```

Adding the extra byte here implicitly tests that the serialization only writes the necessary data even if there is more available space in the buffer.

sfoster1 (changes requested):
Looks good overall but I think some recent changes mean you need to change what happens in `include/motor-control/core/motion_controller.hpp` - I think that file moved

sfoster1 (approved):
Nice!

PR #318, opened 2022-04-05
refactor(motor-control): separate stepper motor & brushed motor files

ahiuchingau:
This PR only changes the file locations - no code is being touched.


sfoster1 (commented):
Isn't some of the motor-control core stuff shared? Should some be not in `stepper_motor` or `brushed_motor` but in a core? and actually there doesn't seem to be a `motor-control/core/brushed_motor`

ahiuchingau:
@sfoster1 There's already a `motor-control/core/brushed_motor` folder on main, but it only has one file as far for the brushed motor driver. The motion controller for the brushed motor is not fully built out yet, but it'll be my next step

motor-control/firmware/brushed_motor/brushed_motor_driver_hardware.cpp
```diff
@@ -1,5 +1,5 @@
+#include "../motor_control_hardware.h"
```
Laura-Danielle, line 1: can we not do relative paths pls

motor-control/firmware/brushed_motor/brushed_motor_hardware.cpp
```diff
@@ -1,6 +1,6 @@
-#include "motor-control/firmware/brushed_motor_hardware.hpp"
+#include "motor-control/firmware/brushed_motor/brushed_motor_hardware.hpp"
 
-#include "motor_control_hardware.h"
+#include "../motor_control_hardware.h"
```
Laura-Danielle, line 3: Same here

Laura-Danielle (approved):
LGTM, I just have a strong preference that we don't use relative paths w/ the `..` due to past traumas from python

PR #319, opened 2022-04-08
feat(sensors): update simulation values

caila-marashaj:


include/can/core/messages.hpp
```diff
@@ -530,7 +530,7 @@ struct BaselineSensorRequest : BaseMessage<MessageId::baseline_sensor_request> {
 
 struct ReadFromSensorResponse : BaseMessage<MessageId::read_sensor_response> {
     can_ids::SensorType sensor{};
-    int32_t sensor_data = 0;
+    int32_t sensor_data = 0x10;
```
Laura-Danielle, line 533: Let's remove this and change it back to 0.

Laura-Danielle (changes requested):
As we discussed in our call, let's actually just change the size of the number allowed in the register map for simulation.

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
```diff
@@ -42,8 +42,9 @@ struct ReadCapacitanceCallback {
     }
 
     void send_to_can() {
-        int32_t capacitance =
+        int16_t capacitance =
```
Laura-Danielle, line 45: This should be `int32_t`

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
```diff
@@ -78,7 +79,7 @@ struct ReadCapacitanceCallback {
     float current_offset;
     int32_t zero_threshold;
     int32_t measurement = 0;
-    uint16_t number_of_reads = 1;
+    uint32_t number_of_reads = 1;
```
Laura-Danielle, line 82: Keep this at `uint16_t`

include/sensors/core/tasks/environment_sensor_callbacks.hpp
```diff
@@ -33,7 +33,7 @@ struct HumidityReadingCallback {
 
   private:
     CanClient &can_client;
-    int32_t humidity = 0;
+    int16_t humidity = 0;
```
Laura-Danielle, line 36: This should be `int32_t`

include/sensors/core/tasks/environment_sensor_callbacks.hpp
```diff
@@ -59,7 +59,7 @@ struct TemperatureReadingCallback {
 
   private:
     CanClient &can_client;
-    int32_t temperature = 0;
+    int16_t temperature = 0;
```
Laura-Danielle, line 62: Same here

include/sensors/tests/callbacks.hpp
```diff
@@ -24,8 +24,8 @@ struct UpdateCallback {
 
 struct MultiUpdateCallback {
   public:
-    uint8_t register_a_value;
-    uint8_t register_b_value;
+    uint16_t register_a_value;
```
Laura-Danielle, line 27: The registers are `uint8_t`, the data they send back are `uint16_t`

Laura-Danielle (changes requested):
What you should be changing is the size of the data _returned_ from the registers, not what's sent over CAN. We want to keep that value at `int32_t`.

You should be changing back the variables data is saved to in all of the `handle_data` functions (if you haven't already).

The diff for this PR should only contain 3 file changes:
- `simulation/sensors.hpp`
- `simulation/hdc2080.hpp`
- `simulation/fdc1004.hpp`

Laura-Danielle (approved):
lgtm but you do have some conflicts you need to resolve here now. Let me know if you need help walking through it.

PR #320, opened 2022-04-12
Feat(Encoders): Add pipette encoder functionality

Carlos-fernandez:
This PR enables the encoder for each Pipette type. The encoder index pin is a different pin configuration based on the low throughput pipettes vs the high throughput pipettes. I have made a condition to compensate the encoder index pin configuration based on pipette_type. This firmware is functional on the pipette by just reading the count register when rotating the pipette. This should bring up encoder pulse count to the CANbus messages that was enable on PR [#317](https://github.com/Opentrons/ot3-firmware/pull/317).


sfoster1 (changes requested):
I think some autogenerated stuff got in there, we should change that; otherwise looks good

sfoster1 (approved):
Looks good to me!

Laura-Danielle (approved):

PR #321, opened 2022-04-12
Gripper encoder configuration

Carlos-fernandez:
This PR enables the Encoder on the Gripper. The pin configuration is based on the schematic details from the gripper PCB. I haven't tested a gripper encoder on hardware until we get the gripper PCB through electrical validation first. I'm confident that this encoder configuration will work based off the other boards on the OT3.  This should bring up encoder pulse count to the CANbus messages that was enable on PR https://github.com/Opentrons/ot3-firmware/pull/317.


sfoster1 (approved):
Looks good to me if checks pass!

PR #323, opened 2022-04-14
feat(common): Separate SPI freertos task

Laura-Danielle:
## Overview

To support multiple items using the same SPI bus, we need to separate SPI into its own task. This lays groundwork needed for multiple drivers using the same SPI bus on the 96 channel.

## Changelog
- Updated all of the task files to add in a spi task
- Edited the driver task to accept a spi writer
- Moved spi to its own project
- Added more tests on the spi code changes

## Review Requests
Please take a look at how I handle write request responses. I couldn't think of what else to do except handle a failure to write to the driver.

## TODO
Making this a draft because there is still some work that needs to be done:
- [x] Get tests passing
- [x] Add cohesive comments where needed
- [x] Decide what to do with a `send data` response for the tmc2130 driver
- [x] Test on hardware (Gantry/Head/Pipettes)
- [x] Fix linter
- [x] Get simulator working


.github/workflows/spi.yaml
```diff
@@ -0,0 +1,75 @@
+name: "spi build/test"
+on:
+  pull_request:
+    paths:
+      - "spi/**"
```
sfoster1, line 5: probably wants to trigger on changes to `common/**` as well, both here and for the push trigger

gantry/firmware/axis_hardware_config.h
```diff
@@ -1,11 +1,14 @@
+#pragma once
+
+#include "stdint.h"
 #include "platform_specific_hal_conf.h"
 
-typedef struct {
+typedef struct PinType {
```
sfoster1, line 6: i'm perfectly fine with c anonymous structs and enums, since the `struct PinType` is never explicitly used - any reason for this change?

gantry/firmware/axis_x_hardware_config.c
```diff
@@ -21,6 +21,8 @@ GantryHardwarePin gantry_hardware_get_gpio(GantryHardwareDevice device) {
             return pinout;
 
         default:
+            pinout.port = GPIOB;
```
sfoster1, line 24: we definitely want the port at least to stay null, i'd think. maybe even force a crash here.

gantry/firmware/axis_y_hardware_config.c
```diff
@@ -19,7 +19,9 @@ GantryHardwarePin gantry_hardware_get_gpio(GantryHardwareDevice device) {
             pinout.port = GPIOB;
             pinout.pin = GPIO_PIN_6;
             return pinout;
-
-        default: return pinout;
+        default:
+            pinout.port = GPIOB;
```
sfoster1, line 23: ditto with keeping port null or forcing a crash

head/firmware/motor_hardware.c
```diff
@@ -334,6 +334,7 @@ void MX_TIM7_Init(void) {
     TIM_MasterConfigTypeDef sMasterConfig = {0};
 
     htim7.Instance = TIM7;
+//    htim7.Init.Prescaler = 2000;
```
sfoster1, line 337: probably remove this

include/spi/core/messages.hpp
```diff
@@ -0,0 +1,106 @@
...
+/**
+ *
+ * @brief Transaction identifier struct
+ *
+ * @param[token] id of the message (generally the register reading/writing to)
+ * @param[command_type] whether it's a read or write command
+ * @param[requires_response] whether the message requires a response
+ */
+struct TransactionIdentifier {
+    uint8_t token;
```
sfoster1, line 19: probably a good idea to make this a little bigger; it can be used as a bitmap by calling tasks to differentiate different kinds of calls at a higher level, and it's nice to have the space. uint32 maybe?

include/spi/core/writer.hpp
```diff
@@ -0,0 +1,122 @@
...
+     *
+     * @tparam[RQType] The originating response queue
+     * @param[addr] The address to write to
+     * @param[command_data] The contents to write to the address
+     * @param[response_queue] The queue with which the SPI task will write a
+     * response
+     * @return A success boolean
+     */
+    template <OriginatingResponseQueue RQType>
+    auto write(uint8_t addr, uint32_t command_data, RQType& response_queue)
```
sfoster1, line 77: the address thing here is a register address on the remote device, right?

sfoster1 (changes requested):
Couple little things inline; some things that didn't fit inline are

- Change the github workflows for executables that use spi (all of them i guess?) to also trigger on changes to the new `include/spi/**` and `spi/**` paths
- This is definitely a great step forward but as I read it, it still requires more changes to actually support multiple devices on the same spi bus; specifically, these transmit and receive commands will need to be able to encode which chip select to use for the transaction, right?

include/spi/core/writer.hpp
Laura-Danielle: yes it is. Do you prefer I call it something else like 'remote_addr'?

Laura-Danielle:
> This is definitely a great step forward but as I read it, it still requires more changes to actually support multiple devices on the same spi bus; specifically, these transmit and receive commands will need to be able to encode which chip select to use for the transaction, right?

Yes this is true. I was planning to add it in when I add the additional chip select pins for the 96 channel, but I can do it for this PR.

gantry/firmware/axis_hardware_config.h
Laura-Danielle: No reason really. I can keep it anonymous.

sfoster1:
> > This is definitely a great step forward but as I read it, it still requires more changes to actually support multiple devices on the same spi bus; specifically, these transmit and receive commands will need to be able to encode which chip select to use for the transaction, right?
> 
> Yes this is true. I was planning to add it in when I add the additional chip select pins for the 96 channel, but I can do it for this PR.

Nope definitely good without it, we can do it later

include/spi/core/writer.hpp
sfoster1: yeah or register_addr or something

sfoster1 (approved):
Nice, looks good to me!

PR #326, opened 2022-04-18
refactor(*): remove start args from task template

sfoster1:
The types of the arguments that you use to start a task aren't actually
relevant to the FreeRTOSTask _class_ template, just its start() method.
If that start method becomes templated instead of the class, then its
argument types can be inferred from the call.

This requires a minor refactor to curry the argument holding tuples
through a lambda so that the task starter data doesn't need a
class-level type definition.


amitlissack (approved):
Neat

PR #327, opened 2022-04-18
refactor(*): commonize task starters

sfoster1:
We currently have dedicated task starters for each task, and they don't do much - they're really just boilerplate. They exist to have some storage for task stacks and to hold some freertos-specific types we don't want in the tests. We can turn that into a generic template, especially if we refactor tasks to have fewer class template arguments in favor of deduction-friendly templated `operator()` methods.

This PR is best reviewed by going through each commit in series.


include/common/core/freertos_task.hpp
```diff
@@ -73,4 +74,28 @@ class FreeRTOSTask {
     std::function<void()> starter{};
 };
 
+template <uint32_t StackDepth,
+          template <template <typename> typename> typename TaskObj>
```
sotaoverride, line 78: one liner comment here to explain that `template <template <typename> typename> typename TaskObj` part of the template parameter list takes care of `template <template <class> class QueueImpl>` (or whatever it actually maps to) sounds helpful to me.

amitlissack, line 78: Is there a way to express this in a more easily understandable manner?

amitlissack (commented):
Really sweet improvement. 

Looks like it's just lint errors. 

Can you manually run integration tests against this PR?

include/motor-control/core/tasks/move_group_task.hpp
```diff
@@ -126,6 +124,8 @@ class MoveGroupTask {
     /**
      * Task entry point.
      */
+    template <motion_controller_task::TaskClient MotionControllerClient,
+              message_writer_task::TaskClient CanClient>
```
ahiuchingau, line 128: I don't understand how & when this is getting called

include/common/core/freertos_task.hpp
```diff
@@ -73,4 +74,28 @@ class FreeRTOSTask {
...
+struct TaskStarter {
+    TaskStarter() : queue{}, task_entry{queue}, task{task_entry} {}
```
amitlissack, line 80: You could go further and remove the queue from this class. Rather than this class creating the `TaskEntry` instance it can be given a reference.

include/common/core/freertos_task.hpp
sfoster1: Yeah that's a good idea, will do

sfoster1: Unfortunately I'm pretty sure there's not, but a comment should help a lot

include/motor-control/core/tasks/move_group_task.hpp
sfoster1: It gets called in `FreeRTOSTask` in the lambda in the `start` method, here: https://github.com/Opentrons/ot3-firmware/blob/5b1dd97a56ac7732d693c5926a819ff706a9dd6c/include/common/core/freertos_task.hpp#L49

The [std::apply](https://en.cppreference.com/w/cpp/utility/apply)+tuple pattern is the c++ way of doing a call like this in python:
```python

class Deferrer:
   def __init__(self, f, *defer_args, **defer_kwargs):
      self._f = f
      self._defer_args = defer_args
      self._defer_kwargs = defer_kwargs

  def run(self):
      self._f(*self._defer_args, **self._defer_kwargs)

def foo(a, b, c=1):
    print(f"{a} {b} {c}")

d = Deferrer(foo, 1, 2, c=34)
d.run() # prints "1, 2, 34"
```

include/common/core/freertos_task.hpp
sfoster1: Something like this?
```c++
// freertos_task.hpp
template <size_t StackDepth, typename TaskObj>
struct TaskStarter {
    explicit TaskStarter(TaskObj& task_entry) : task_entry{task_entry}... {}
    ...
};
// e.g. pipettes/core/tasks.cpp
auto capacitive_queue = freertos_queue::FreeRTOSMessageQueue<sensors::tasks::TaskMessage>{};
auto capacitive_task = sensors::tasks::CapacitiveSensorTask(capacitive_queue);
auto capacitive_starter = freertos_task::TaskStarter<512>(capacitive_task);
```

If so, I'd really like to not do that - it means that there's now three lines of boilerplate per task in tasks.cpp rather than just one.

amitlissack: That's fine.

sfoster1:
@amitlissack Integration tests all pass, though I'm not sure what a good way to indicate that is

amitlissack (approved):

sfoster1:
Rebased and ran the integration tests again, still all good

PR #328, opened 2022-04-22
feat(actions): run CI actions with the debug flag

Laura-Danielle:
## Overview

 For some reason, certain things get optimized out of the code using _RelWithDebInfo_ that causes compilation to be successful when it shouldn't be. There isn't a good way to switch between D flags in CMAKE, so for now we should just run CI using the debug flag to try and catch any things that compiling locally did not catch.


sfoster1 (approved):
Looks good; we should probably do both, honestly, but this is solid.

PR #329, opened 2022-04-22
feat(eeprom): eeprom can messages generalized

amitlissack:
Modify eeprom can message payloads. These can specify a target address, data length, and buffer.

The messages don't do anything yet. That's for a future PR.


Laura-Danielle (approved):
Nice. Looks good to me

sfoster1 (approved):
Nice

PR #330, opened 2022-04-22
feat(pipettes): add tip handling CAN messages

Laura-Danielle:
## Overview

Add the firmware side of the tip handling commands. Implementation of these commands will happen in
a separate PR.


amitlissack (approved):
Thanks for the tip! (argggg)

sotaoverride (approved):

PR #332, opened 2022-04-25
refactor(pipettes): move pin configurations around for 96 channel

Laura-Danielle:
## Overview

Thorough pass of GPIO pins for the low and high throughput pipettes.


sfoster1 (approved):
Nice

PR #333, opened 2022-04-25
feat(spi): Support multiple cs lines

Laura-Danielle:
## Overview

Closes #Ret-917. Last supporting component before multiple motor drivers can be supported on the same SPI bus.

## Changelog

- Add chip select interface struct that gets passed around using the driver configs (there might be a better place to put this)
- Modify all of the subprojects to support this change
- Modify tests as needed

## Tasks

- [x] Test on hardware

### DIFF
--- a/gantry/simulator/interfaces.cpp
+++ b/gantry/simulator/interfaces.cpp
@@ -30,7 +30,27 @@ static auto motor_interface = sim_motor_hardware_iface::SimMotorHardwareIface();
 static freertos_message_queue::FreeRTOSMessageQueue<motor_messages::Move>
     motor_queue("Motor Queue");
 
-static auto driver_configs = utils::driver_config();
+static tmc2130::configs::TMC2130DriverConfig driver_configs{
+    .registers = {.gconfig = {.en_pwm_mode = 1},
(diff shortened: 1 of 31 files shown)


gantry/simulator/interfaces.cpp
```diff
@@ -30,7 +30,27 @@ static auto motor_interface = sim_motor_hardware_iface::SimMotorHardwareIface();
...
+                               .tbl = 0x2,
+                               .mres = 0x3},
+                  .coolconf = {.sgt = 0x6}},
+    .current_config =
+        {
+            .r_sense = 0.1,
+            .v_sf = 0.325,
+        },
+    .chip_select{
+        .cs_pin = 0,
```
sfoster1, line 51: i think you probably want separate configs here with separate chip selects so the simulator can tell them apart as a hook for future simulator improvements

sfoster1 (approved):
Looks good to me!

gantry/simulator/interfaces.cpp
Laura-Danielle: I'll make a ticket to track that

ahiuchingau (approved):
LGTM

PR #335, opened 2022-04-26
refactor(common): do not use abort for pin configuration switch cases

Laura-Danielle:
## Overview

Rather than sending back bogus values, we should simply use `abort` from the standard library to
exit out of a switch statement that doesn't meet the required case.

Also fixed an issue where I accidentally deleted a return statement for the `sync_out` pin response.

### DIFF
--- a/gantry/core/utils.cpp
+++ b/gantry/core/utils.cpp
@@ -1,15 +1,16 @@
 #include "gantry/core/utils.hpp"
 
-#include <cstdlib>
-
 auto utils::get_node_id_by_axis(enum GantryAxisType which) -> can_ids::NodeId {
     switch (which) {
         case GantryAxisType::gantry_x:
             return can_ids::NodeId::gantry_x;
         case GantryAxisType::gantry_y:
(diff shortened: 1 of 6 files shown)


sfoster1 (changes requested):
We should not do this.

If the code reaches this point, it's a programmer error, right? That means two things
1. It's not gonna get fixed on its own, so restarting the system won't help
2. It _must_ leave the system in a natively recoverable state
3. It _should_ leave enough traces of what happened for a developer to be able to fix the problem

`abort()` will _probably_, but i dunno, do 1. 2 is best served by bouncing out to the bootloader and marking the firmware bad, though I don't know if we have the capability.  It's probably best to ticket that capability and add it later. And the best way to do 3 is probably to drop into a debug loop, an infinite loop, so you can debug attach and backtrace to see the problem.

sfoster1 (changes requested):
Ok this is now a pr that has a couple entirely separatable tiny things - can we split it?

PR #336, opened 2022-04-27
refactor(eeprom):  pull out the eeprom task from include pipettes

amitlissack:
- Moving files
- Renaming namespaces


include/eeprom/core/types.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
-namespace eeprom_types {
+namespace eeprom {
+namespace types {
```
sotaoverride, line 5: is this the same as `namespace eeprom::types{}` ?

sotaoverride (approved):

include/eeprom/core/types.hpp
amitlissack: Yup

amitlissack: It is. I'll rename. Thanks

PR #338, opened 2022-04-28
changed pressure and lowpasspressure reg values, i2c_sim change

caila-marashaj:


i2c/simulation/i2c_sim.cpp
```diff
@@ -14,12 +14,15 @@ auto SimI2C::central_transmit(uint8_t *data, uint16_t size,
...
     transmit_count++;
     if (!sensor_map.empty()) {
-        uint16_t store_in_register = 0;
-        auto *iter = data + 1;
-        iter = bit_utils::bytes_to_int(iter, data + size, store_in_register);
-        LOG("Storing %d in register %d", store_in_register, reg);
-        sensor_map[dev_address].REGISTER_MAP[reg] = store_in_register;
         next_register_map[dev_address] = reg;
+        if (size > 1) {
+            uint16_t store_in_register = 0;
```
sfoster1, line 19: we almost certaiinly need the special storage logic here as well as in read()

sfoster1 (approved):
This looks good though I think you need to have the special-cased sizing thing in the writer as well as the reader

i2c/simulation/i2c_sim.cpp
```diff
@@ -14,12 +14,20 @@ auto SimI2C::central_transmit(uint8_t *data, uint16_t size,
...
+            auto *iter = data + 1;
+            if (size == 3) {
+                iter = bit_utils::bytes_to_int(iter, data + size,
+                                               store_in_register);
+            } else if (size == 5) {
+                uint32_t store_in_register = 0;
```
sfoster1, line 25: This definitely won't do what you think. It will save the value to the variable defined on line 25 and not the one defined on line 19, so you'll write 0 on line 29 always. I think you have to pull the assignment into the `if` clauses.

sfoster1 (changes requested):

include/motor-control/core/types.hpp
```diff
@@ -3,7 +3,7 @@
 #include <cstdint>
 
 using sq0_31 = int32_t;   // 0: signed bit,  1-31: fractional bits
-using sq14_15 = int32_t;  // 0: signed bit,  1-31: fractional bits
+using sq15_16 = int32_t;  // 0: signed bit,  1-31: fractional bits
```
sfoster1, line 6: let's update the comment here

sfoster1 (approved):
Nice, looks good to me. Let's just update the comment on the sq15_16 type.

Laura-Danielle (approved):
lgtm, just one comment request.

PR #340, opened 2022-04-29
refactor(i2c): I2c read register

amitlissack:
`i2c::writer::Writer` has a `write` method that accepts a register address. This is a nice convenience for writing a buffer to a device that expects the first byte of a transaction to be a memory or register address.

This PR adds the `read` equivalent. When using this method, a transaction is created that initially writes a single byte address followed by a read.


Laura-Danielle (approved):
lgtm

PR #341, opened 2022-04-29
feat(eeprom): create eeprom hardware driver

amitlissack:


include/eeprom/core/write_protect.hpp
```diff
@@ -0,0 +1,72 @@
...
+     * Constructor
+     * @param pin
+     */
+    explicit WriteProtector(WriteProtectPin& pin) : pin{pin} {}
+
+    /**
+     * Disable the write protect
+     */
+    void disable() {
+        if (count++ == 0) {
```
ahiuchingau, line 47: Could you explain the logic of the disable count?

amitlissack: The eeprom needs to have write protection disabled for all writes. I want to disable write protect at the start of a write and re-enable it when there are no more writes. 

Due to the asynchronous nature of the i2c task it can be tricky to know when this happens. There could be multiple write requests in the queue simultanously. 

The eeprom task calls disable whenever it gets a write to eeprom request. It then calls enable whenever the write request completes. 

So image two write requests arrive:
eeprom task processes first request -> disables write protect (changes pin state, disable count is 1)
eeprom task processes second request -> disables write protect (does nothing to pin, only increments disable count to 2)
first request completes -> enables write protect (does nothing to pin, decrement count to 1)
second request completes -> enables write protect (changes pin state, decrement count to 0)

include/eeprom/core/write_protect.hpp
```diff
@@ -0,0 +1,72 @@
+#pragma once
+
+namespace eeprom {
+namespace write_protect {
+
+/**
+ * Interface to write protect pin. Must be implemented in FW and Simulation
+ */
+class WriteProtectPin {
```
Laura-Danielle, line 9: nit: can we separate the `WriteProtectPin` class and call it something like `EepromHardwareIface` (following the pattern of `motor_hardware_interface.hpp`?

pipettes/core/tasks.cpp
```diff
@@ -113,7 +114,7 @@ void pipettes_tasks::start_tasks(
     spi_task_client.set_queue(&spi_task.get_queue());
 
     auto& eeprom_task =
-        eeprom_task_builder.start(5, "eeprom", i2c3_task_client, queues);
+        eeprom_task_builder.start(5, "eeprom", i2c3_task_client, eeprom_wp_pin);
```
Laura-Danielle, line 117: Mistake? Don't we still need to give the queue client to the eeprom task builder?

eeprom/tests/test_eeprom_task.cpp
```diff
@@ -1,43 +1,191 @@
...
+                        address);
+                REQUIRE(transact_message.transaction.write_buffer[1] ==
+                        data[0]);
+                REQUIRE(transact_message.transaction.write_buffer[2] ==
+                        data[1]);
+                REQUIRE(transact_message.transaction.write_buffer[3] ==
+                        data[2]);
+                REQUIRE(transact_message.transaction.write_buffer[4] ==
+                        data[3]);
+                REQUIRE(transact_message.id.token == static_cast<uint32_t>(-1));
```
Laura-Danielle, line 57: Can we just set it to message handler's `WRITE_TOKEN` type?

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -1,7 +1,6 @@
 #pragma once
 
-#include <optional>
-
+#include "can/core/can_writer_task.hpp"
```
Laura-Danielle, line 3: Mistake?

Laura-Danielle (approved):
Small comments. Main thing I would like verified is the comment about the queue client/memcpy.

eeprom/tests/test_eeprom_task.cpp
```diff
@@ -1,43 +1,191 @@
...
+
+                auto transact_message =
+                    std::get<i2c::messages::Transact>(i2c_message);
+                REQUIRE(transact_message.transaction.address == 0xA0);
+                REQUIRE(transact_message.transaction.bytes_to_read ==
+                        data_length);
+                REQUIRE(transact_message.transaction.bytes_to_write == 1);
+                REQUIRE(transact_message.transaction.write_buffer[0] ==
+                        address);
+                REQUIRE(transact_message.id.token == 0);
```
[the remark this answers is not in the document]
amitlissack: No. They are generated using the IdMap.

include/eeprom/core/messages.hpp
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+namespace eeprom {
+namespace message {
+
+/** Eeprom message */
+struct EepromMessage {
+    eeprom::types::address memory_address{0};
```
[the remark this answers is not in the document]
amitlissack: Good question. Perhaps we don't.

include/eeprom/core/write_protect.hpp
amitlissack: Ok

pipettes/core/tasks.cpp
amitlissack: We do not. The eeprom task only communicates with an i2c task.

include/sensors/core/tasks/pressure_driver.hpp
amitlissack: I don't think so. I believe that it solved a build error.

include/eeprom/core/messages.hpp
```diff
@@ -0,0 +1,31 @@
...
+    eeprom::types::data_length length{0};
+    eeprom::types::EepromData data{};
+
+    auto operator==(const EepromMessage&) const -> bool = default;
+};
+
+using ReadResponseCallback = void (*)(const EepromMessage&, void*);
```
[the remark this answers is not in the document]
amitlissack: I'm not using std::function but a plain old "c" style function pointer. So that issue doesn't factor in.

My initial idea was to use a pointer to an interface like:

```
class ResponseHandler {
    public:
        virtual void handle(const EepromMessage&) = 0;
}
```

The plain old function pointer + param makes it possible to bind to an instance function without using a virtual function.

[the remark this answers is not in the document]
Laura-Danielle: Got it. That makes sense.

include/eeprom/core/hardware_iface.hpp
```diff
@@ -0,0 +1,73 @@
...
+ * Class to manages the state of the write protect pin across multiple
+ * transactions.
+ *
+ * Consecutive write requests will want to have write protection disabled. The
+ * enable should happen when all the write transactions have completed.
+ *
+ * This class keeps the eeprom from having to keep track of the write protection
+ * state.
+ */
+class WriteProtector {
```
sfoster1, line 36: this definitely will work as long as it's called from one task, and I don't think it'll be a problem, but this is definitely a semaphore which [freertos has an implementation of](https://www.freertos.org/xSemaphoreCreateCountingStatic.html) - maybe we should back this class with one of those?

sfoster1 (approved):
Looks good to me overall, though I think we should consider using a semaphore directly or as a backing data structure for the write protect counter.

PR #342, opened 2022-04-29
feat(gripper): Enabling gripper motor homing

ahiuchingau:
This PR does a couple of things:
1. Set up separate move group and move status report tasks for the G motor
2. Update `GripperGrip` and `GripperHome` CAN message so they conform to the move group message format
3. Update Brushed Motion Controller to handle CAN messages correctly
4. Add brushed motor interrupt handler to take care of reading from the: 
   i. homing switch (implemented here)
  ii. encoder (tbd)
5. Have gripper send back MoveCompleted message whenever the limit switch is triggered
6. Enable the correct TIM1 interrupt 

#### TODO
~~Change the actual pwm freq & duty cycle based on the CAN message params~~  (DONE in this PR)


include/motor-control/core/brushed_motor/brushed_motion_controller.hpp
```diff
@@ -20,6 +25,36 @@ class MotionController {
...
 
+    void move(const can_messages::GripperGripRequest& can_msg) {
+        BrushedMove msg{.duration = can_msg.duration,
+                        .freq = can_msg.freq,
+                        .duty_cycle = can_msg.duty_cycle,
+                        .group_id = can_msg.group_id,
+                        .seq_id = can_msg.seq_id,
+                        .stop_condition = MoveStopCondition::none};
+        queue.try_write(msg);
+        hardware.grip();
```
Laura-Danielle, line 36: I thought we were going to be writing to a queue for handling the motor interrupts, so do we still need these `grip` and `home` functions?

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -0,0 +1,121 @@
...
+    void set_buffered_move(BrushedMove new_move) { buffered_move = new_move; }
+
+    [[nodiscard]] auto get_buffered_move() const -> BrushedMove {
+        return buffered_move;
+    }
+
+    bool has_active_move = false;
+
+  private:
+    q31_31 position_tracker = 0x0;
```
Laura-Danielle, line 115: we probably don't need a position tracker here right?

Laura-Danielle (approved):
Looks great to me! Just two comments/questions

include/motor-control/core/brushed_motor/brushed_motion_controller.hpp
ahiuchingau: Essentially, the `grip` and `home` functions actually begin the PWM generation. Only after the PWM generation has started, the motor interrupt handler could try to read from the limit switch. 

I struggled to think of another better way to handle this, let me know what you think!

sfoster1: Based on previous discussions - you're right, this is a problem. There's two solutions I can see:
1. Only turn the PWM channel off (so, CC1) when you're not running PWM; keep the timer core running, and generating overflow interrupts, all the time.
2. Use more of the hardware. The PWM pulse callback does two things: count down time remaining, and check the limit switch. The limit switch can be set up as a gpio external interrupt that sends a message to this task. The time remaining could be loaded in a timer compare register of its own. Then you wouldn't have any logic running in interrupts - you'd stop the PWM if you got an EXTI or if you got the timer duration countdown ping.

I think these all make sense; when we spoke earlier, I recommended 2, but given that at some point we're going to add a full position controller to this logic, 1 probably makes more sense at second thought.

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    void update_move() {
+        has_active_move = queue.try_read_isr(&buffered_move);
+        if (has_active_move &&
+            buffered_move.stop_condition == MoveStopCondition::limit_switch) {
+            position_tracker = 0x7FFFFFFFFFFFFFFF;
+        }
+        // (TODO: lc) We should check the direction (and set respectively)
```
sfoster1, line 66: let's remove some old comments

gripper/firmware/stm32g4xx_it.c
```diff
@@ -142,7 +142,7 @@ void FDCAN1_IT0_IRQHandler(void) {
  * @brief This function handles TIM1 update interrupt and TIM16 global
  * interrupt.
```
sfoster1, line 143: should change the comment here

sfoster1 (changes requested):
I think we can fix up the interrupt handling logic with a minimum of effort and it'll end up being a lot better. We just have to trigger the logic interrupt off the timer core's overflow rather than the PWM pulse generated.

sfoster1 (changes requested):
This overall looks great. The only thing I want to be sure of is that we either
- Make the grip motor PWM timer base frequency that drives the grip motor interrupt the same as the stepper motor interrupt frequency so that the ticks are the same length, or
- Clearly, obviously, and everywhere note that the length of a brushed dc motor tick and a stepper motor tick are different
Leave it up to you which; both might end up ok, but I'd prefer the first if possible.

PR #343, opened 2022-04-29
feat(motor-control): add tmc2160 motor control

Laura-Danielle:
## Overview

Closes RET-915. Adds in a basic setup for the `tmc2160` driver. For now it follows the `tmc2130` driver very closely. Both files should probably be refactored at some point to reduce the amount of functions in each driver.

## Follow-ups

These tasks will be used in a follow-up PR implementing [Handle multiple motors for pick up tip](https://opentrons.atlassian.net/browse/RET-916)


amitlissack (approved):
Can't speak to the functionality, but the code looks good and clean.

PR #345, opened 2022-04-29
fix(pipettes): add back in configurations for driver clk and spi cs pin

Laura-Danielle:
## Overview

Accidentally deleted the chip select pin on the pipettes project which caused the motors to not be
enabled. Also moved the driver clock stuff around a little bit.


sfoster1 (approved):
excellent

PR #346, opened 2022-04-29
chore(*): update can bit timings

sfoster1:
CANbus bit timings are a complicated thing. They depend on four different parameters. All four of the parameters we were using in the firmware were different from what the host system uses in linux, and were all just written down in a way that made it really non-obvious what they meant.

This makes the non-obviousness different by calculating those values from named parameters with units and with compile-time range and correctness checks.

### DIFF
--- a/head/firmware/can.c
+++ b/head/firmware/can.c
@@ -6,7 +6,12 @@
  * @param handle Pointer to an FDCAN handle
  * @return HAL_OK on success
  */
-HAL_StatusTypeDef MX_FDCAN1_Init(FDCAN_HandleTypeDef* handle) {
+HAL_StatusTypeDef MX_FDCAN1_Init(
+    FDCAN_HandleTypeDef* handle,
+    uint8_t clock_divider,
+    uint8_t segment_1_tqs,
+    uint8_t segment_2_tqs,
+    uint8_t max_sync_jump_width) {
     handle->Instance = FDCAN1;
     handle->Init.ClockDivider = FDCAN_CLOCK_DIV1;
     handle->Init.FrameFormat = FDCAN_FRAME_FD_NO_BRS;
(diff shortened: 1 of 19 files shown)


head/firmware/can.c
```diff
@@ -6,22 +6,27 @@
...
-    handle->Init.NominalPrescaler = 20;
-    handle->Init.NominalSyncJumpWidth = 2;
-    handle->Init.NominalTimeSeg1 = 14;
-    handle->Init.NominalTimeSeg2 = 2;
-    handle->Init.DataPrescaler = 20;
-    handle->Init.DataSyncJumpWidth = 1;
-    handle->Init.DataTimeSeg1 = 14;
-    handle->Init.DataTimeSeg2 = 1;
+    handle->Init.NominalPrescaler = clock_divider;
+    handle->Init.NominalSyncJumpWidth = max_sync_jump_width;
```
Laura-Danielle, line 23: Curious if using the max value here is correct?

head/firmware/main.cpp
```diff
@@ -249,6 +250,21 @@ auto timer_for_notifier = freertos_timer::FreeRTOSTimer(
     }),
     100);
 
+// Unfortunately, these numbers need to be literals or defines
+// to get the compile-time checks to work so we can't actually
+// correctly rely on the hal to get these numbers - they need
+// to be checked against current configuration. However, they are
```
Laura-Danielle, line 256: Should this comment be in every file that creates a bit timing?

Laura-Danielle (approved):
Overall looks great. Love the `bit_timings.hpp` templates. Just had two questions.

head/firmware/can.c
sfoster1: it's just a naming thing. The `Nominal` in the fdcan init structure is contrasted to `Data`, not `max`  - it's for the flexible datarate thing you can do, where you start off at the nominal timings and once you've transferred enough of the arbitration id and header you switch to the Data timings.

head/firmware/main.cpp
sfoster1: I mean maybe but also it's so much

sfoster1:
Superceded by #348

PR #347, opened 2022-05-03
feat(pipettes): add additional basic timer for the gear motors

Laura-Danielle:
## Overview
Closes RET-915 and is related to RET-916. First small step in enabled 3 motors on the 96 channel.

## Notes

I chose to use the other basic timer for now. We might want to re-evaluate the types of timers we use more closely depending on each timers individual application.


sfoster1 (approved):
⚙️

PR #348, opened 2022-05-03
chore(can): generate can timings

sfoster1:
CANbus bit timings are a complicated thing. They depend on four different parameters. All four of the parameters we were using in the firmware were different from what the host system uses in linux, and were all just written down in a way that made it really non-obvious what they meant.

This makes the non-obviousness different by calculating those values from named parameters with units and with compile-time range and correctness checks.

After this PR, the bit timings used by the controllers will be identical to what they were before, but they'll now be generated and tied in with actual canbus canonical parameters. The next step is to change them to more useful values.

This supercedes #346

### DIFF
--- a/pipettes/firmware/main.cpp
+++ b/pipettes/firmware/main.cpp
@@ -5,6 +5,7 @@
 
 // clang-format on
 
+#include "can/core/bit_timings.hpp"
 #include "can/core/ids.hpp"
 #include "can/firmware/hal_can_bus.hpp"
 #include "common/core/app_update.h"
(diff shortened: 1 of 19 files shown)


pipettes/firmware/main.cpp
```diff
@@ -118,6 +119,24 @@ static sensors::hardware::SensorHardware pins_for_sensor_96(gpio::PinConfig{
...
+// - 275.33KHz bitrate
+// - 87.5% sample point
+// Should drive
+// segment 1 = 8 quanta
+// segment 2 = 1 quantum
+
+// For the exact timing values these generate see
+// can/tests/test_bit_timings.cpp
+static constexpr auto can_bit_timings =
+    can::bit_timings::BitTimings<110 * can::bit_timings::MHZ, 455, 275330,
```
Laura-Danielle, line 137: 275.33 * can::bit_timings::KHZ ?

Laura-Danielle (approved):
looks good, just one comment

pipettes/firmware/main.cpp
Laura-Danielle: Resolved by [follow-up](https://github.com/Opentrons/ot3-firmware/pull/349/files), so you can ignore my comment.

PR #349, opened 2022-05-03
chore(can): update timings to 500mhz

sfoster1:
CANbus bit timings need to be the same to a tolerance of 0.5%. Previous bit timings absolutely were not the same across different microcontroller families or between the microcontrollers and linux, for a variety of reasons such as
- They just weren't configured to be: the numbers we put in the registers wouldn't actually generate timings within that spec
- The pipettes are configured with a systemclock of 100MHz, not 110MHz

By clocking the G4 FDCAN off the PLL (aka 170MHz rather than 85MHz) we can achieve a 100ns time quantum duration with a divider of 17, which is also achievable by the L5s (clocked from 100 (!) MHz, divider 10) and linux (20MHz, divider 2). This works for both 250kbaud bitrate and for 500kbaud bitrate. It probably works for 1Mbaud too, but we'll try that later.

Since it works well, this also changes the default baudrate to 500kbits.


sfoster1:
> Looks fine, but wasn't Joe saying 50ns might be too short of a time period yesterday?

Yes, and I don't intend to merge this until we've actually verified the physical functionality - that's a big part of why I split this off

Laura-Danielle (approved):

ahiuchingau (approved):
Code looks good to me, haven't tried testing on the actual gripper board myself

PR #350, opened 2022-05-03
chore(i2c): Writer returns try_write result

amitlissack:
for clients keeping track of transactions, it's nice to know whether the transaction made it to the i2c queue.


Laura-Danielle (approved):
Nice. Would you mind making this change in the `spi/core/writer.hpp` file too? Otherwise I can do it later.

PR #353, opened 2022-05-04
Ret 935 can eeprom

amitlissack:


include/can/core/messages.hpp
```diff
@@ -204,7 +204,7 @@ struct ReadFromEEPromResponse : BaseMessage<MessageId::read_eeprom_response> {
             std::min(data_length,
                      static_cast<eeprom::types::data_length>(limit - iter)),
             iter);
-        return iter - body;
+        return limit - body;
```
Laura-Danielle, line 207: Shouldn't this still be `iter`?

include/eeprom/core/task.hpp
```diff
@@ -85,10 +177,9 @@ class EEPromTask {
     /**
      * Task entry point.
      */
-    template <message_writer_task::TaskClient CanClient>
     [[noreturn]] void operator()(i2c::writer::Writer<QueueImpl> *writer,
-                                 CanClient *can_client) {
-        auto handler = EEPromMessageHandler{*writer, *can_client, get_queue()};
+                                 write_protect::WriteProtectPin *pin) {
+        auto handler = EEPromMessageHandler{*writer, get_queue(), *pin};
```
Laura-Danielle, line 182: I'm very confused as to which queue is receiving the eeprom can messages. Don't we still need the queue client so we can actually receive messages from CAN?

Laura-Danielle (commented):
The diff is being pretty weird here, but I am still confused about not needing the queue client. How does eeprom receive messages from CAN?

include/can/core/messages.hpp
amitlissack: The size of the message must always be the same. Regardless of the number of bytes read, we send the complete buffer.

sfoster1 (approved):
Nice, looks good to me!

include/eeprom/core/task.hpp
Laura-Danielle: got it

PR #356, opened 2022-05-06
adding CAN message for peripheral device status request 

caila-marashaj:


include/can/core/ids.hpp
```diff
@@ -87,6 +87,8 @@ enum class MessageId {
     sensor_diagnostic_response = 0x89,
     bind_sensor_output_request = 0x8a,
     bind_sensor_output_response = 0x8b,
+    peripheral_info_request = 0x8c,
```
amitlissack, line 90: Was this changed by hand or through the code gen?

include/can/core/messages.hpp
```diff
@@ -843,6 +843,35 @@ struct TipActionResponse
...
+
+struct PeripheralInfoResponse
+    : BaseMessage<MessageId::peripheral_info_response> {
+    can_ids::SensorType sensor;
+    uint8_t status;
+
+    template <bit_utils::ByteIterator Output, typename Limit>
+    auto serialize(Output body, Limit limit) const -> uint8_t {
+        auto iter = bit_utils::int_to_bytes(sensor, body, limit);
+        iter = bit_utils::int_to_bytes(status, body, limit);
```
amitlissack, line 867: This looks like a bug. `body` should be `iter`, no?

i2c/simulation/i2c_sim.cpp
```diff
@@ -14,12 +14,22 @@ auto SimI2C::central_transmit(uint8_t *data, uint16_t size,
...
-        auto *iter = data + 1;
-        iter = bit_utils::bytes_to_int(iter, data + size, store_in_register);
-        LOG("Storing %d in register %d", store_in_register, reg);
-        sensor_map[dev_address].REGISTER_MAP[reg] = store_in_register;
         next_register_map[dev_address] = reg;
+        if (size > 1) {
+            auto *iter = data + 1;
+            uint16_t store_in_register = 0;
+            // check if address belongs to pressure sensor
+            if (dev_address == 0x67 << 1) {
```
amitlissack, line 22: The i2c simulator could use some upgrades so you don't have to do stuff like this. We should have simulated i2c devices and route data based on device address. The register map is not sufficient.

I'm not asking you to change anything.

include/can/core/ids.hpp
caila-marashaj: by hand - I branched off of the wrong branch, and I've got to cherry pick that

PR #358, opened 2022-05-10
feat(pipettes): use multiple motor driver tasks

Laura-Danielle:
## Overview

This turned into a refactor I didn't initially intend. I tried doing [SFINAE](https://en.cppreference.com/w/cpp/language/sfinae) to avoid multiple `can_task.cpp` files and multiple types of 'linear queue' clients, but I couldn't get things to work properly -- definitely open to ideas here.

## In this PR
- Separating the task client(s) into multiple types. For now we have:
1. Linear motor tasks (for lead screw driven motors). Single/multi only have these types of motors.
2. Gear motor tasks (for gear driven motors). 96/384 only have these types of motors.
3. peripheral tasks (for SPI/I2C)
4. sensor tasks (for things on the sensor board + eeprom)
- Adding in the usage of tmc2160 as a task. It is only going to be used as the driver for the lead screw motor on the 96 channel in the pipettes project.
- Adding in tasks for the gear motors (this will be fleshed out more in follow-up PRs)

## What's not in this PR?
- Motion Client for the gears
- Implementing a message handler to actually handle picking up/dropping tips.

## Review Requests
If you've got a better idea of how to handle the dispatcher in 1 file, let me know. Also same making the linear motor driver task(s)/queue(s) templatized -- though I'm not sure if that will be more readable or not.

## To-Do

- [x] Test on hardware
- [x] Clean up comments


pipettes/core/can_task_ht.cpp
```diff
@@ -0,0 +1,125 @@
+#include "can/core/freertos_can_dispatch.hpp"
```
amitlissack, line 1: Is it clear to everyone what `ht` and `lt` are?

amitlissack (approved):
Looks good to me.

pipettes/core/can_task_ht.cpp
Laura-Danielle: Probably not. I'll add a comment in both files.

ahiuchingau (approved):
Looks good!

PR #359, opened 2022-05-11
fix(eeprom): eeprom debugging

amitlissack:
Results from debugging eeprom on pipette and gripper.

- pin settings weren't right
- write protect is active high
- read response was needlessly large


sfoster1 (approved):
Looks good!

Laura-Danielle (approved):

PR #360, opened 2022-05-12
fix(pipettes): separate i2c maps for i2c 1 and 3.

amitlissack:
Pipettes simulator was not starting up.

The eeprom and one of the sensors share the same device address.

This separates i2c sensor map for i2c 1 and 3.

@Laura-Danielle and @caila-marashaj , do I have the mappings correct?


Laura-Danielle (approved):
Yep that's good enough for now to mimic the single channel.

PR #361, opened 2022-05-12
fix(gripper): fix motor interrupt handler & switch dir pins

ahiuchingau:
The pins for the brushed motor direction were incorrect. This PR also fixes the logic in the motor interrupt handler so it actually behaves correctly (only monitor the limit switch when the axis is homing).


sfoster1 (approved):
Nice!

sotaoverride (approved):

PR #362, opened 2022-05-13
feat(hardware): Convert encoder pulses to position

Carlos-fernandez:
This PR adds encoder pulses/um and pulses/mm to the linear motion system. This converts the encoder pulses to a readable position value that can be convert from the Opentrons mono repo. 
Additional fixes and add ons:
 -A fix to the overflow value on the encoder
- Change Timers to the encoder head configuration and Timer periods to be the same from the Left and Right Mounts.
- Add clear interrupt register field bit and Direction bit to the motion-control

This branch has been tested on the robot and works just fine.


gantry/tests/test_linear_motion_config.cpp
```diff
@@ -7,8 +7,10 @@ TEST_CASE("Linear motion system using a belt") {
     GIVEN("OT2 X,Y gantry config") {
         struct LinearMotionSystemConfig<BeltConfig> linearConfig {
             .mech_config = BeltConfig{.pulley_diameter = 12.7},
-            .steps_per_rev = 200, .microstep = 32,
+            .steps_per_rev = 200, .microstep = 32, .encoder_ppr = 1000,
         };
         REQUIRE(linearConfig.get_steps_per_mm() == 160.40813f);
+        REQUIRE(linearConfig.get_encoder_pulses_per_mm() ==
+                100.25508f);  // 25.063770566
```
sfoster1, line 14: confused - what's the number in the comment and why is it different?

include/motor-control/core/linear_motion_system.hpp
```diff
@@ -24,11 +24,13 @@ concept MotorMechanicalConfig = requires {
     std::is_same_v<MC, BeltConfig> || std::is_same_v<MC, LeadScrewConfig>;
 };
 
+// template <MotorEncoderConfig ECConfig>
```
sfoster1, line 27: let's remove the commented code

include/motor-control/core/linear_motion_system.hpp
```diff
@@ -24,11 +24,13 @@ concept MotorMechanicalConfig = requires {
...
 template <MotorMechanicalConfig MEConfig>
 struct LinearMotionSystemConfig {
     MEConfig mech_config{};
     float steps_per_rev{};
     float microstep{};
+    float encoder_ppr{};  // pulses_per_revolution
```
sfoster1, line 33: if we're not confident that the name `encoder_ppr` gets it across, we should use a different name so we don't need the comment

include/motor-control/core/linear_motion_system.hpp
```diff
@@ -38,6 +40,14 @@ struct LinearMotionSystemConfig {
         return (mech_config.get_mm_per_rev()) /
                (steps_per_rev * microstep * gear_ratio) * 1000;
     }
+    [[nodiscard]] constexpr auto get_encoder_pulses_per_mm() const -> float {
+        return (encoder_ppr * 4.0 * gear_ratio) /
+               (mech_config.get_mm_per_rev());
+    }
+    [[nodiscard]] constexpr auto get_encoder_um_per_pulse() const -> float {
+        return (mech_config.get_mm_per_rev()) /
+               (encoder_ppr * 4.0 * gear_ratio) * 1000.0;
```
sfoster1, line 49: `encoder_ppr` is occasionally initialized to `0`, making this unsafe - we should have this function return early with 0 if `encoder_ppr == 0`.

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
```diff
@@ -221,9 +221,36 @@ class MotorInterruptHandler {
...
+            } else {
+                enc_position_tracker =
+                    UINT16_MAX + hardware.get_encoder_pulses();
+                enc_overflow_future_flag = true;
+            }
+        } else {
+            enc_position_tracker = hardware.get_encoder_pulses();
+        }
+        return enc_position_tracker;
+        /* This function fixes the overflow issue for each motor.
```
sfoster1, line 239: kind of a nitpick but a comment like this should probably be at the start of the function, not the end

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
```diff
@@ -221,9 +221,36 @@ class MotorInterruptHandler {
...
+        Check whether the UIEF interrupt bit in the Status register gets
+        triggered. If it gets triggered we save the previous state of the
+        triggered flag and clear the status register interrupt bit.
+        */
+    }
 
     void reset_encoder_pulses() { hardware.reset_encoder_pulses(); }
+
+    void clear_encoder_overflow_flag() { hardware.clear_encoder_SR(); }
```
sfoster1, line 248: these one-liner functions are a bit confusing, could we just call the hardware methods directly? Or, if you want to stick with functions like this, you should make sure you use them everywhere.

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
```diff
@@ -221,9 +221,36 @@ class MotorInterruptHandler {
         set_buffered_move(Move{});
     }
 
-    auto get_encoder_pulses() { return hardware.get_encoder_pulses(); }
+    auto get_encoder_pulses() {
+        if (get_encoder_overflow_status() == true) {
+            hardware.clear_encoder_SR();
```
sfoster1, line 226: we should either be always using things like `clear_encoder_overflow_flag` or always be using things like `hardware.clear_encoder_SR`, not both.

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
```diff
@@ -221,9 +221,36 @@ class MotorInterruptHandler {
...
+            if (enc_overflow_future_flag) {
+                enc_position_tracker = hardware.get_encoder_pulses();
+                enc_overflow_future_flag = false;
+            } else {
+                enc_position_tracker =
```
sfoster1, line 231: I don't think this logic is correct - what happens the _second_ time there's an overflow?

include/motor-control/simulation/sim_motor_hardware_iface.hpp
```diff
@@ -27,10 +27,20 @@ class SimMotorHardwareIface : public motor_hardware::StepperMotorHardwareIface {
     void reset_encoder_pulses() final { test_pulses = 0; }
     uint32_t get_encoder_pulses() final { return test_pulses; }
     void sim_set_encoder_pulses(uint32_t pulses) { test_pulses = pulses; }
+    bool clear_encoder_SR() final { return false; }
+    bool get_encoder_SR_flag() final {
+        if (sr_flag) {
```
sfoster1, line 32: kind of a nitpick, but can't we make this function just `return !sr_flag`?

sfoster1 (changes requested):
There's some inline changes I'd like to see, and obviously it needs a format, but there's three bigger problems we should fix before merging:

The overflow handling logic seems like it handles precisely one overflow, and only a positive-direction overflow at that. Whenever there's an overflow, we _set_ the tracker to `UINT16_MAX + pulses`. So the second time you overflow, you'll set it back to `UINT16_MAX + pulses`. And if you overflow negative, you'll set it to `UINT16_MAX + pulses`. We likely need some more elaborate logic here.

Second, the logic here is getting a bit complex - I think we should pull out the encoder tracking stuff into its own class, which will make it easier to solve issue 3, testing.

We definitely need tests for the overflow handling logic.

include/motor-control/core/linear_motion_system.hpp
ahiuchingau, line 33: Small nit: if you specify the default value as 0.0 as in line 34 for gear ratio, then you don't have to write out encoder_ppr=0 for those systems that doesn't actually have an encoder, i.e. gripper z axis

include/motor-control/core/linear_motion_system.hpp
ahiuchingau, line 49: You might want to return a different value for when the `encoder_ppr` value is 0 here

ahiuchingau (commented):
Can't wait to see the encoders in action!

gantry/tests/test_linear_motion_config.cpp
Carlos-fernandez: oh sorry, that was a note, bad habit of mine... That's when i was testing the first logic of the conversion and i was getting that number and i knew it was wrong.. I'll remove these comments...

include/motor-control/core/linear_motion_system.hpp
Carlos-fernandez: this is fine, just forgot to remove this comment.. PPR is standard for encoders, at least that's what I've found online.

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
Carlos-fernandez: Either an overflow or a underflow can happen. When the status register is clear, the flag can get triggered again when an underflow happens. So the previous state of the overflow is saved to track the previous state because we cleared the current status register. On the head, these overflows happen only once, that's if the mounts moves greater  than ~180mm from the home position. Once that is triggered, we can detect an underflow if the motor reverses direction from less ~180mm. This logic wouldn't work if the robot triggers 2 overflows, but on the head can't trigger 2 overflows with one move. There isn't enough distance on the head to trigger 2 overflows.

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
Carlos-fernandez: We can stick with just hardware methods.

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
sfoster1: Yeah, but this is fundamental enough that I'd really like to handle that case, both so it's clearly correct code to read and to handle future physical changes. For instance, what if we put a higher-count encoder on? What if we increase the gear ratio that the encoder is on the other side of.

Carlos-fernandez: I'll look into a different method to handle this.

sfoster1:
Superceded by #406

PR #363, opened 2022-05-13
fix(eeprom): gripper eeprom interaction.

amitlissack:
- fix i2c3 pin definitions for gripper
- fix eeprom wp pin definition
- Move the eeprom write disable to i2c_setup


sfoster1 (approved):
Excellent, looks great

ahiuchingau (approved):

PR #366, opened 2022-05-17
refactor(pipettes): pull motor pin configurations out of main

Laura-Danielle:
## Overview

Related to #RET-916 and #RET-918. Pulling the pin configurations out of main so it will be easier to
get the correct configurations depending on the pipette type and motor.


pipettes/firmware/interfaces.cpp
```diff
@@ -6,9 +6,83 @@
...
 #pragma GCC diagnostic pop
 
-auto interfaces::driver_config_by_axis(PipetteType which)
-    -> tmc2130::configs::TMC2130DriverConfig {
+auto interfaces::driver_config_by_axis(PipetteAxisType which)
+    -> driver_utils::DriverConfigType {
     switch (which) {
+        case PipetteAxisType::left_gear_motor:
+            return tmc2130::configs::TMC2130DriverConfig{
+                .registers = {.gconfig = {.en_pwm_mode = 1},
```
amitlissack, line 14: These configurations look really similar. Is there a way to compose these structs in a way that reduces some of the repetition?

include/motor-control/core/stepper_motor/utils.hpp
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+namespace driver_utils {
+struct DriverConfigType {};
```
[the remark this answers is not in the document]
Laura-Danielle: This was a workaround to have two different config structs be returned from the same function. I'm def open to other ideas.

sfoster1 (changes requested):
We should make it impossible to build a high-throughput pipette configuration and then use the `low_throughput_motor` configuration element. Right now, the only thing preventing that is developer discipline and tests. Let's take more advantage of the type system.

Have the low-throughput and high-throughput motor and driver configurations be different structs. Then,
- If you want to keep `include/core/interfaces.hpp`, have two separate functions that return the appropriate thing, and switch which one you call in `main` based on `PIPETTE_TYPE`. This may be better if you make the functions `consteval`.
- If you want to avoid the explicit switch, then make the `get_config`/`get_driver` functions fully specialized templates on values of the enum type that `PIPETTE_TYPE` is for and dispatch that way; if you do that though, the implementations of `get_config`/`get_driver` have to go in the header and would need to be core-compatible to live in core, and you may need to have two separate headers.

Either way, we definitely need to remove the possibility of a mistake by not having a valid struct that contains elements of both low throughput and high throughput configurations.

include/pipettes/core/interfaces.hpp
```diff
@@ -1,7 +1,89 @@
...
+template <PipetteType P>
+auto driver_config() -> std::enable_if_t<P == PipetteType::SINGLE_CHANNEL,
+                                         LowThroughputPipetteDriverHardware>;
+
+template <PipetteType P>
+auto driver_config() -> std::enable_if_t<P == PipetteType::NINETY_SIX_CHANNEL,
+                                         HighThroughputPipetteDriverHardware>;
+
+template <>
+auto driver_config<PipetteType::SINGLE_CHANNEL>()
```
sfoster1, line 82: don't think we need both the explicit specializations and enable_if

include/motor-control/firmware/stepper_motor/pipette_motor_hardware.hpp
```diff
@@ -0,0 +1,52 @@
+#pragma once
```
sfoster1, line 1: This definitely shouldn't be here - it's specific to pipettes so it should be in `include/pipettes` somewhere.

motor-control/firmware/stepper_motor/pipette_motor_hardware.cpp
```diff
@@ -0,0 +1,43 @@
+#include "motor-control/firmware/stepper_motor/pipette_motor_hardware.hpp"
```
sfoster1, line 1: this file is specific to the pipette and should be in `pipettes/` somewhere

sfoster1 (changes requested):
The structure of this looks good and I really love the type safety but if a file has `pipette` in its name, and it defines classes with `Pipette` in their name, it really needs to live in the `include/pipettes` or `pipettes/` subtree.

sfoster1 (approved):
Awesome!

PR #367, opened 2022-05-17
fix(gripper): direction & v ref calculations

ahiuchingau:
Some small fixes to enable the gripper motor to grip correctly on the OT3: I was wrong about the directions and used the wrong fixed point index for calculating the brushed motor VREF.


Laura-Danielle (approved):
I saw it with me own eyes

PR #371, opened 2022-05-18
feat(gripper,pipettes): read serial number from eeprom

amitlissack:
We read the serial number from the eeprom.

The eeprom on these boards is can only read 8 bytes at a time. Rather than supporting a 12-byte serial number by doing two reads, we'll just use 8 bytes. No need to write throwaway code.

### DIFF
--- a/eeprom/tests/test_serial_number.cpp
+++ b/eeprom/tests/test_serial_number.cpp
@@ -1,6 +1,5 @@
 #include "catch2/catch.hpp"
 #include "eeprom/core/serial_number.hpp"
-//#include "eeprom/core/addresses.hpp"
 #include <vector>
 
 
@@ -14,7 +13,7 @@ struct MockEEPromTaskClient {
 };
 
 
-struct MockListener {
+struct MockListener: serial_number::ReadListener {
     void on_read(const serial_number::SerialNumberType& sn) {this->sn = sn; call_count++;}
     serial_number::SerialNumberType sn{};
     int call_count{0};
--- a/eeprom/tests/test_serial_number.cpp
+++ b/eeprom/tests/test_serial_number.cpp
@@ -77,19 +77,16 @@ SCENARIO("Reading serial number") {
         WHEN("the read completes") {
             auto sn = serial_number::SerialNumberType{8, 7, 6, 5, 4, 3, 2, 1};
 
-            for (auto m = queue_client.messages.cbegin();
-                 m < queue_client.messages.cend(); m++) {
-                auto read_message = std::get<message::ReadEepromMessage>(*m);
-                auto data = types::EepromData{};
-                std::copy_n(sn.cbegin() + read_message.memory_address,
-                            read_message.length, data.begin());
+            auto read_message = std::get<message::ReadEepromMessage>(queue_client.messages[0]);
+            auto data = types::EepromData{};
+            std::copy_n(sn.cbegin() + read_message.memory_address,
+                        read_message.length, data.begin());
 
-                read_message.callback(
-                    {.memory_address = read_message.memory_address,
-                     .length = read_message.length,
-                     .data = data},
-                    read_message.callback_param);
-            }
+            read_message.callback(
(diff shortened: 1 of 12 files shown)


eeprom/tests/test_serial_number.cpp
```diff
@@ -0,0 +1,100 @@
+#include <vector>
+
+#include "catch2/catch.hpp"
+#include "eeprom/core/serial_number.hpp"
+
+using namespace eeprom;
+
+struct MockEEPromTaskClient {
+    void send_eeprom_queue(const task::TaskMessage& m) {
```
Laura-Danielle, line 9: We actually already have a shared mock queue client if you want to use it. It's in `include/common/tests`

eeprom/tests/test_serial_number.cpp
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    GIVEN("A request to read the serial number") {
+        subject.start_read();
+
+        WHEN("the read completes") {
+            auto sn = serial_number::SerialNumberType{8, 7, 6, 5, 4, 3, 2, 1};
+
+            for (auto m = queue_client.messages.cbegin();
```
Laura-Danielle, line 80: Do we need the for loop for this test?

Laura-Danielle (approved):
Some general comments, but overall looks great!

include/eeprom/core/addresses.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include "types.hpp"
+
+namespace eeprom {
+namespace addresses {
+
+constexpr types::data_length serial_number_length = 12;
```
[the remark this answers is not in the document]
sfoster1: Yeah I'm confused about the correspondence of this type and the PR body.

I will also note that we really do want a longer serial number. We don't get to choose this - we use what the factory wants, and what they want is a datecode of YYYYMMDD, the ordinal of when the device was made (i.e. 1 for the first device that day, 2 for the next, etc), and sometimes a letter. 12 is a good compromise, but we may even want it longer at some point, and it definitely won't be shorter.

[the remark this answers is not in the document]
amitlissack: Nah. We'll only read / write 8 of the 12. 

That means that the python can stay as it is.

include/gripper/core/gripper_info.hpp
```diff
@@ -8,50 +8,38 @@
...
-                              const GripperInfo &gripper_info)
-        : writer(writer), response{.model = gripper_info.model} {
-        std::copy_n(
-            gripper_info.serial.cbegin(),
-            std::min(gripper_info.serial.size(), response.serial.size()),
-            response.serial.begin());
-    }
+    explicit GripperInfoMessageHandler(CanClient &writer,
+                                       EEPromClient &eeprom_client)
+        : writer{writer}, serial_number_accessor{eeprom_client, *this} {}
```
sotaoverride, line 35: can we just pass `this` rather than `*this` ?

amitlissack: No. Why do you ask?

include/eeprom/core/addresses.hpp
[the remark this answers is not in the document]
amitlissack: Our current eeprom supports 8 byte reads and writes. Rather than doing multiple writes and reads to support a larger SN, which would be throw away logic, I chose to support an  8 byte SN. 

The production eeprom will support up to 64-byte reads and writes. Large enough for a complete SN.

include/gripper/core/gripper_info.hpp
sotaoverride: ` SerialNumberAccessor(EEPromTaskClient& eeprom_client,
                                  ReadListener& listener)` for some reason got me thinking we could just pass this, rather than a dereference this and pass a copy

amitlissack: `this` is a pointer. We're not passing a copy.

PR #372, opened 2022-05-19
feat(gripper, pipettes): store serial number

amitlissack:
- Implement `SetSerialNumber` message. 
- Handle this message in the Gripper and Pippettes


sfoster1 (approved):
Looks good. Do we want write confirmation?

amitlissack:
> Looks good. Do we want write confirmation?

I don't think so. We can always use the GripperInfoRequest or PipetteInfoRequest messages.

PR #373, opened 2022-05-24
feat(sensors): add autothreshold

sfoster1:
The capacitive sensor on the pipette is extremely sensitive to its immediate environment. We probably want to zero out the ambient capacitance when it is close to, but not yet touching, whatever it's going to sense.

We could do that today by commanding a reading via can from python; getting the reading back; and then commanding a threshold based on that reading via can. We could also send a command that tells the pipette to take a reading and then use that reading as a new threshold, and we'll save some back-and-forth.

Todo: tests, better use of the message including specifying an offset amount. Required for https://github.com/Opentrons/opentrons/pull/10067

### DIFF
--- a/include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
+++ b/include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
@@ -8,6 +8,7 @@
 #include "common/core/logging.h"
 #include "sensors/core/fdc1004.hpp"
 #include "sensors/core/sensor_hardware_interface.hpp"
+#include "sensors/core/utils.hpp"
 
 namespace sensors {
 namespace tasks {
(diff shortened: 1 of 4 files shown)


sotaoverride (approved):

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
```diff
@@ -72,10 +73,15 @@ struct ReadCapacitanceCallback {
         }
         auto capacitance = convert_capacitance(measurement, number_of_reads,
                                                current_offset_pf);
-        auto message = can_messages::ReadFromSensorResponse{
-            .sensor = SensorType::capacitive,
-            .sensor_data = convert_to_fixed_point(capacitance, 15)};
-        can_client.send_can_message(can_ids::NodeId::host, message);
+        if (utils::tag_in_token(m.id.token,
+                                utils::ResponseTag::IS_THRESHOLD_SENSE)) {
+            set_threshold(capacitance + 1);
```
Laura-Danielle, line 78: Why are we adding a 1 here?

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
```diff
@@ -72,10 +73,15 @@ struct ReadCapacitanceCallback {
...
+        } else {
+            auto message = can_messages::ReadFromSensorResponse{
+                .sensor = SensorType::capacitive,
+                .sensor_data = convert_to_fixed_point(capacitance, 15)};
```
Laura-Danielle, line 82: Should this be a `q15_16`? I might've messed up the fixed point location before.

Laura-Danielle (approved):
some small comments, but otherwise looks good

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
sfoster1: That's the threshold. It's the current average, plus a little bit. In a followup PR I definitely want to make that `1` settable from python.

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
sfoster1: It matches up with the python side, so I'd like to keep it for now at least

PR #374, opened 2022-05-25
feat(pipettes): add support for tip action messages

Laura-Danielle:
## Overview

Related to (and closes) RET-916 RET-918.

Ignore the first two commits, and everything after should be fairly separated. This PR enables tip pick up for the 96 channel by separating out the two "gear" motors (aka tip pickup/drop off motors). There may potentially need to be additional work to more properly synchronize the two motors, but that will need to be determined through testing. Also, we still need to figure out how exactly we want to handle (or if we want to handle) `TipAction` requests for the low throughput pipettes.

## Changelog

This PR does a few things:

1. Creates new motion controller that only handles tip action requests and checks tip sense value
2. Make the gear tasks support two different motors
3. Enable the interrupt(s) for both gear motors
4. Separates out the global initialization(s) in `main.cpp` for both simulation and firmware

## Review Requests
How do you feel about the organization of things?

## To Do

- [x] Test on hardware
- Test out the enable motor situation since all three motors share the same enable line
- [x] Test new simulation for 96 channel


include/pipettes/core/motor_configurations.hpp
```diff
@@ -0,0 +1,102 @@
...
+auto motor_configurations()
+    -> std::enable_if_t<P == PipetteType::NINETY_SIX_CHANNEL,
+                        HighThroughputMotorConfigurations>;
+
+template <PipetteType P>
+auto motor_configurations()
+    -> std::enable_if_t<P == PipetteType::THREE_EIGHTY_FOUR_CHANNEL,
+                        HighThroughputMotorConfigurations>;
+
+template <>
```
sfoster1, line 86: This stuff definitely needs some comments about what the SFINAE stuff is for

include/pipettes/firmware/interfaces.hpp
```diff
@@ -0,0 +1,114 @@
+#pragma once
+
+#include "motor-control/core/stepper_motor/motor_interrupt_handler.hpp"
```
sfoster1, line 3: comments about SFINAE here

include/motor-control/core/stepper_motor/motion_controller.hpp
```diff
@@ -110,3 +109,96 @@ class MotionController {
...
+    [[nodiscard]] auto get_mechanical_config() const
+        -> const lms::LinearMotionSystemConfig<MEConfig>& {
+        return linear_motion_sys_config;
+    }
+
+    void move(const can_messages::TipActionRequest& can_msg) {
+        steps_per_tick velocity_steps =
+            fixed_point_multiply(steps_per_mm, can_msg.velocity);
+        auto stop_condition = MoveStopCondition::none;
+        if (check_tip_sense()) {
```
sfoster1, line 153: we definitely cannot do this, at least long term - what if something gets stuck on and pushing again might pop it off? We need to have a way for higher levels with more context to command tip actions that don't necessarily make sense to this level of the code. the tip action message should specify what kind of action it needs to take.

include/pipettes/core/tasks/motion_controller_task.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+    void handle(const can_messages::SetMotionConstraints& m) {
+        LOG("Received set motion constraints: minvel=%d, maxvel=%d, minacc=%d, "
+            "maxacc=%d",
+            m.min_velocity, m.max_velocity, m.min_acceleration,
+            m.max_acceleration);
+        controller.set_motion_constraints(m);
+    }
+
+    void handle(const can_messages::TipActionRequest& m) {
```
sfoster1, line 87: So the gear motors can't home or do arbitrary moves, is that the idea? How do they home?

sfoster1 (changes requested):
Love the reorganization, love the way you've used the type system to keep this safe with the different configurations. Confused about some stuff though:

- What actions, and therefore can messages, can the gear motors handle? Can they home?
- Why should we limit the tip actions based on the current switch? It seems limiting - can we just specify the action in the message?
- There's still some core motor controller stuff with pipette in the name, we should either figure out a way to make it not have pipette in the name, or move it into the pipette tree

include/pipettes/core/tasks/motion_controller_task.hpp
Laura-Danielle: A drop tip will continue moving the motors until the limit switches are triggered.

sfoster1 (approved):
OK, this looks great to me now. Definitely want to fix up the motion controller, but that can be later. Let's just make sure to check this on some low throughput pipettes as well before merging.

PR #375, opened 2022-05-26
feat(sensors): Allow setting an offset to autothreshold

sfoster1:
We shouldn't hardcode the offset we set when we do an autothreshold; we should allow it to be specified from canbus. This adds a new field to the threshold messages that specifies the mode of the threshold.


Laura-Danielle (approved):
lgtm

PR #376, opened 2022-05-31
feat(common): create a custom allocator template that statically allocates memory

Laura-Danielle:
## Overview
We did it!!

Alt Text

This is a first pass at creating a custom allocator template. It seems to generally work with some caveats (i.e. weird stuff happens with vector and deque). Ideally, we'll be able to use this allocator to help reduce repetitive code! For reference, we followed the [Allocator interface](https://en.cppreference.com/w/cpp/named_req/Allocator) to build out this class.

## Review Requests
- Take a look at naming, see if there are other tests (not listed in the to-do list) that might be good to have etc.

## To-Do

- [x] Figure out the exact requirements we aren't meeting for vector and deque (something is missing with the constructor type)
- [x] Add more detailed comments for everything
- [x] Add a test where the element is a class

### DIFF
--- a/include/common/core/allocator.hpp
+++ b/include/common/core/allocator.hpp
@@ -4,42 +4,39 @@
 
 #include "common/core/logging.h"
 
-/*
- * A generic custom allocator class that can be used to statically allocate
- memory
- * for typically dynamic objects (such as maps).
+/**
+ * A generic custom allocator class that can be used to statically
+ * allocate memory for typically dynamic objects (such as maps).
  *
- * To use this class, you must pass in the CustomAllocator class to the template
- of
- * any stl library you would like to use. For example, to use this class with
- map
+ * To use this class, you must pass in the CustomAllocator class
+ * to the template of any stl library you would like to use.
+ *
+ * For example, to use this class with map
  * you would do the following:
  *
- * std::map<std::string, int, std::less<std::string>,
-        CustomAllocator<std::pair<const std::string, int>, max_entries>>
- subject{};
+ * std::map<
+ *      std::string, int, std::less<std::string>,
+ *      CustomAllocator<std::pair<const std::string, int>, max_entries>
+ * > subject{};
  *
  * Caveats:
- * Any containers that reallocate every time (vectors for example) a new element
- is
- * added can only support fully allocated objects. For example, say you have a
- vector:
+ * Any containers that reallocate every time (vectors for example) a
+ * new element is added can only support fully allocated objects.
+ * For example, say you have a vector:
  *
  * max entries = 2;
  * std::vector<int, CustomAllocator<int, max_entries>> subject{2, 3};
  *
- * You can change the elements of this array, but you cannot dynamically re-size
- it.
+ * You can change the elements of this array, but you cannot
+ * dynamically re-size it.
  *
  * If you tried to insert another element like:
  *
  * subject.push_back(4)
  *
  * it will throw a bad alloc which is expected behavior. However, vectors will
- also
- * throw bad allocs if things haven't previously been allocated. For example,
- say you have a vector
- * defined with no elements:
+ * also throw bad allocs if things haven't previously been allocated.
+ * For example, say you have a vector defined with no elements:
  *
  * max entries = 2;
  * std::vector<int, CustomAllocator<int, max_entries>> subject;
@@ -47,10 +44,10 @@
  * subject.push_back(1);
  * subject.push_back(2); // throws bad alloc
  *
- * This will throw a bad alloc because `subject.push_back(2)` tries to allocate
- memory for
- * both integers 1 and 2 which is more space than available.
- */
+ * This will throw a bad alloc because `subject.push_back(2)` tries
+ * to allocate memory for both integers 1 and 2 which
+ * is more space than available.
+ **/
 
 template <typename Element, std::size_t MaxElements>
 class CustomAllocator {


include/common/core/allocator.hpp
```diff
@@ -0,0 +1,84 @@
...
+                      backing.begin()) /
+                     sizeof(AllocatorElement<Element>);
+        LOG("Deallocating %d", index);
+
+        if (index < MaxElements) {
+            backing[index].used = false;
+        }
+    }
+
+    pointer allocate(size_t, const void* = 0) {
```
fsinapi, line 63: Since there's no implementation here for `max_size()`, isn't there a chance that `allocate()` will be invoked to allocate more than one item at a time?

include/common/core/allocator.hpp
```diff
@@ -0,0 +1,84 @@
...
+    using value_type = Element;
+    using size_type = std::size_t;
+    using difference_type = ptrdiff_t;
+    using pointer = Element*;
+    using const_pointer = const Element*;
+    using reference = Element&;
+    using const_reference = const Element&;
+
+    PoolAllocator() {}
+    auto operator=(PoolAllocator&) -> PoolAllocator& = delete;
```
sfoster1, line 44: you sure about this ctor set? You've got the copy ctor defaulted, which means you can happily copy gigantic pools of memory around - which seems wrong - and then the copy-assign deleted, which is a weird combo.

include/common/core/allocator.hpp
```diff
@@ -0,0 +1,84 @@
...
+    auto operator=(PoolAllocator&&) -> PoolAllocator&& = delete;
+    PoolAllocator(const PoolAllocator&) = default;
+    PoolAllocator(PoolAllocator&&) = delete;
+    ~PoolAllocator() = default;
+
+    void deallocate(pointer loc, size_t) {
+        // Find index into backing array using the fact that loc is the first
+        // item in the AllocatorElement struct and so has the same address.
+        auto index = (reinterpret_cast<AllocatorElement<Element>*>(loc) -
```
sfoster1, line 53: You also need to handle unaligned deletes here with something like
```
auto wrapped = reinterpret_cast<AllocatorElement<Element>*>(loc);
if (wrapped < backing.begin()) return;
if ((wrapped - backing.begin()) % sizeof(*wrapped) != 0) return;
```
this shouldn't ever happen, but if something called `deallocate` with a wrongly-aligned address, then you wouldn't necessarily set an arbitrary byte, but you'd set the previous element which might or might not be right.

include/common/core/allocator.hpp
```diff
@@ -0,0 +1,84 @@
...
+        throw std::bad_alloc();
+    }
+
+    template <class U>
+    struct rebind {
+        typedef PoolAllocator<U, MaxElements> other;
+    };
+
+  private:
+    // Use a struct allocator element and then can return the exact address
```
sfoster1, line 80: leftover comment?

include/common/core/allocator.hpp
sfoster1, line 63: not handling the `size_t` makes this not conformant and also means it can't handle array-allocations. This is probably why it doesn't work with `vector`, since that commonly works by proactively allocating array backings and reallocating larger ones when capacity is exceeded. IMO it's reasonable to scan the backing array for the shortest slot that's >= request.

sfoster1 (changes requested):
Couple leftover comments and I think a mistakenly- ommitted `include/common/core/task.hpp`?

Plus I think not handling array-allocations is why `vector` doesn't work.

include/common/core/allocator.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <variant>
+
+#include "common/core/logging.h"
+
+/*
+ * A generic pool allocator class that can be used to statically allocate memory
+ * for typically dynamic objects (such as maps).
+ *
+ * Example Use-Case:
+ *
```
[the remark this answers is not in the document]
Laura-Danielle: Very old comment. Going to revamp the comments before merging.

include/common/core/allocator.hpp
Laura-Danielle: Hm I don't think allocate would be attempting to allocate more than one item, but we should definitely be checking that the size given doesn't exceed the max size of the element. Will probably end up doing something like [seth suggested](https://github.com/Opentrons/ot3-firmware/pull/376#discussion_r886798397).

fsinapi: `vector`, `deque`, and some other container types allocate arrays instead of individual items, which means `allocate` would attempt to allocate more than one item.

I think that the type of your `backing` array will have to change if you want to support block allocation with Seth's suggestion. If you allocate out a block of multiple contiguous elements in the current backing array, you're giving the caller free rein to write over the `AllocatorElement::used` members.

sfoster1: Oh that's a very good point, you're right.

sfoster1 (approved):
Approved once you remove the empty `include/common/core/task.hpp`

include/common/core/allocator.hpp
```diff
@@ -0,0 +1,153 @@
+#include <algorithm>
+#include <array>
+#include <variant>
+
+#include "common/core/logging.h"
+
+/*
+ * A generic custom allocator class that can be used to statically allocate
```
fsinapi, line 8: Might want to fix up this comment, looks like the formatter chewed it up pretty bad

fsinapi (approved):
Changes look good!

PR #377, opened 2022-05-31
refactor(eeprom): CAN messages support 16-bit messaging.

amitlissack:
FW companion to https://github.com/Opentrons/opentrons/pull/10540

New revs have eeprom that supports 16-bit address. This allows sending 16-bit addresses over CAN, but still uses 8-bit addressing. There's a TODO for how to change.


sfoster1 (approved):
Nice! Thanks for the todo

PR #380, opened 2022-05-31
feat(CAN): add InstrumentInfoRequest

ahiuchingau:
Adding InstrumentInfoRequest so both gripper and pipettes would respond when prompted.


Laura-Danielle (approved):
NICE!

PR #384, opened 2022-06-02
Enable data ready fix

caila-marashaj:
This pr adds hardware support for the data ready pins on the pipette main boards, but does not check that pin's value or incorporate it into the pressure sensor reads. That will be done in a follow-up pr.

### DIFF
--- a/pipettes/firmware/hardware_config.c
+++ b/pipettes/firmware/hardware_config.c
@@ -18,12 +18,15 @@ static PipetteHardwarePin get_gpio_ht(PipetteHardwareDevice device) {
             pinout.port = GPIOB;
             pinout.pin = GPIO_PIN_5;
             return pinout;
+        case pipette_hardware_device_data_ready:
+            pinout.port = GPIOC;
+            pinout.pin = GPIO_PIN_15;
+            return pinout;
         default:
             pinout.port = 0;
             pinout.pin = 0;
             return pinout;
     }
-
 }
 
 static uint16_t get_spi_pins_ht(GPIO_TypeDef* for_handle) {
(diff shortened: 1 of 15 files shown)


sfoster1 (approved):
I'm good with this as long as you smoke test it.

pipettes/firmware/hardware_config.h
```diff
@@ -10,7 +10,8 @@ typedef struct {
 typedef enum {
     pipette_hardware_device_LED_drive,
     pipette_hardware_device_sync_in,
-    pipette_hardware_device_sync_out
+    pipette_hardware_device_sync_out,
```
Laura-Danielle, line 13: I think this is not a needed change.

pipettes/firmware/hardware_config.c
```diff
@@ -68,7 +68,6 @@ static PipetteHardwarePin get_gpio_lt(PipetteHardwareDevice device) {
     }
 }
 
-
```
Laura-Danielle, line 71: same here

Laura-Danielle (approved):
Two changes that you can get rid of b/c they're unused, but otherwise looks good to merge.

pipettes/firmware/hardware_config.c
```diff
@@ -61,6 +64,10 @@ static PipetteHardwarePin get_gpio_lt(PipetteHardwareDevice device) {
             pinout.port = GPIOB;
             pinout.pin = GPIO_PIN_4;
             return pinout;
+        case pipette_hardware_device_data_ready:
```
Laura-Danielle, line 67: This is the wrong function. It should be in `get_gpio_lt`

pipettes/firmware/hardware_config.h
Laura-Danielle: Resolving: It is being used, but it wasn't implemented in `hardware_config.c`

Laura-Danielle (changes requested):

Laura-Danielle (approved):

PR #387, opened 2022-06-08
feat(gripper): add capacitive sensor task

ahiuchingau:
Adding capacitive sensor task to the gripper subproject.


sfoster1 (approved):
Awesome, looks good to me if it works. Glad the reusable stuff is paying off!

PR #388, opened 2022-06-14
feat(head): support rev1 hardware

sfoster1:
The next revision of the head board has tmc2160 motor drivers with discrete FETs rather than tmc2130s with integrated FETs. It also has support for electromagnetic brakes on the motors, and slightly different voltage levels for presence detection.

This PR adds support for the new rev without removing the old one by adding a second set of cmake targets to build firmware for the new rev (without qualifying the rev, i.e. just building `head` instead of `head-proto` or `head-rev`, you'll get the proto revision). 

It also refactors things so key files are different between the two revisions - for now, main.cpp, tasks.cpp, and tasks.hpp - and does this by pulling some common factors of headers out into new common headers.

Before this can be merged, we must
- [x] test the new motor drivers
- [ ] verify that proto hardware still works


head/firmware/main_rev1.cpp
```diff
@@ -0,0 +1,298 @@
...
+        .active_setting = GPIO_PIN_RESET}
+};
+
+// TODO clean up the head main file by using interfaces.
+static tmc2160::configs::TMC2160DriverConfig motor_driver_configs_right{
+    .registers =
+        {
+            .gconfig = {.en_pwm_mode = 1},
+            .ihold_irun = {.hold_current = 0x3,
+                           .run_current = 0xe,
```
Laura-Danielle, line 142: We should set these to 31 & 16 respectively.

Laura-Danielle (approved):
One comment about the current. Otherwise looks good to me.

head/firmware/main_rev1.cpp
sfoster1: wait, other way around, right? hold at 16, run at 31

PR #389, opened 2022-06-15
fix(spi): set a timeout and retries on commands to write to the spi queue

Laura-Danielle:
## Overview

The 96 channel has 3 motors all on the same spi bus. We were seeing an issue where the motor drivers weren't getting fully configured most likely because the queue was being overwritten by the three different motor driver tasks.

This PR adds in a timeout and retries to the spi write command in both motor drivers.

## Requirements

- [ ] Test on hardware with logic analyzer

### DIFF
--- a/include/motor-control/core/stepper_motor/tmc2130_driver.hpp
+++ b/include/motor-control/core/stepper_motor/tmc2130_driver.hpp
@@ -62,15 +62,14 @@ class TMC2130 {
         auto response = false;
 
         // setting a 10 ms timeout and 3 repeats.
-        for (int i=3; i>0; i--) {
-            response = _spi_manager.write(
-                converted_addr, command_data, _task_queue, _cs_intf, 10);
+        for (int i = 3; i > 0; i--) {
+            response = _spi_manager.write(converted_addr, command_data,
+                                          _task_queue, _cs_intf, 10);
             if (response) {
                 i = 0;
             }
         }
         return response;
-
     }
 
     /**
--- a/include/motor-control/core/stepper_motor/tmc2130_driver.hpp
+++ b/include/motor-control/core/stepper_motor/tmc2130_driver.hpp
@@ -59,8 +59,18 @@ class TMC2130 {
 
     auto write(Registers addr, uint32_t command_data) -> bool {
         auto converted_addr = static_cast<uint8_t>(addr);
-        return _spi_manager.write(converted_addr, command_data, _task_queue,
-                                  _cs_intf);
+        auto response = false;
+
+        // setting a 10 ms timeout and 3 repeats.
+        for (int i=3; i>0; i--) {
+            response = _spi_manager.write(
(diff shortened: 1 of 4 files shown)


include/motor-control/core/stepper_motor/tmc2130_driver.hpp
```diff
@@ -59,8 +59,17 @@ class TMC2130 {
...
-        return _spi_manager.write(converted_addr, command_data, _task_queue,
-                                  _cs_intf);
+        auto response = false;
+
+        // setting a 10 ms timeout and 3 repeats.
+        for (int i = 3; i > 0; i--) {
+            response = _spi_manager.write(converted_addr, command_data,
+                                          _task_queue, _cs_intf, 10);
+            if (response) {
+                i = 0;
```
sfoster1, line 69: this is a little gross, can we `break` instead

include/motor-control/core/stepper_motor/tmc2160_driver.hpp
```diff
@@ -59,8 +59,16 @@ class TMC2160 {
...
         auto converted_addr = static_cast<uint8_t>(addr);
-        return _spi_manager.write(converted_addr, command_data, _task_queue,
-                                  _cs_intf);
+        auto response = false;
+        // setting a 10 ms timeout and 3 repeats.
+        for (int i = 3; i > 0; i--) {
+            response = _spi_manager.write(converted_addr, command_data,
+                                          _task_queue, _cs_intf, 10);
+            if (response) {
+                i = 0;
```
sfoster1, line 68: ditto here - let's just `break`

pipettes/firmware/motor_hardware.c
```diff
@@ -29,6 +29,8 @@ void HAL_SPI_MspInit(SPI_HandleTypeDef* hspi) {
         GPIO_InitStruct.Pull = GPIO_NOPULL;
         GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
         HAL_GPIO_Init(GPIOC, &GPIO_InitStruct);
+
+        HAL_GPIO_WritePin(GPIOC, GPIO_InitStruct.Pin, GPIO_PIN_SET);
```
sfoster1, line 33: ?

sfoster1 (changes requested):
Let's get those doing a `break`. Let's also add some tests for the retry.

include/motor-control/core/stepper_motor/tmc2130_driver.hpp
Laura-Danielle: sure

pipettes/firmware/motor_hardware.c
Laura-Danielle: The chip select pins are active low. Was still seeing some weird behavior on the logic analyzer until I set all the chip selects to be high.

sfoster1 (approved):
nice!

PR #390, opened 2022-06-16
fix(i2c): restrict the message queue size

Laura-Danielle:
## Overview

In reality, message queues are restricted to a certain size. Some of the tests around polling relied
on the fact that dequeue (the simulation queue backing) does not have a set size. This PR restricts
that size to the default queue size (10) which also happens to be the poll count size.

@sfoster1 let me know if this keeps the intent of the test correctly.

### DIFF
--- a/i2c/tests/test_i2c_poll_task.cpp
+++ b/i2c/tests/test_i2c_poll_task.cpp
@@ -411,7 +411,6 @@ SCENARIO("test the limited-count i2c poller") {
         }
     }
 }
-}
 
 SCENARIO("test the ongoing i2c polling") {
     test_mocks::MockMessageQueue<i2c::writer::TaskMessage> i2c_queue{};


sfoster1 (changes requested):
No, I think this illustrates a bug. 

There are two messages per poll. Let's say the test sets up a 10-count poll. The test code fires the timer 9 times, which sends 18 messages - (10-1)*2 - in a group; it checks the queue size; then it fires the timer one more time and checks that the poll is complete with another 2 messages.

This worked with an infinite deque. When you limited the size of the deque, that made the test fail because there were fewer messages - the timer fired 9 times, it tried to enqueue 18 messages, but let's say 10 made it in. You fixed it so you're only checking for 10 messages, which makes the test pass, but the behavior is still wrong - those 8 messages are still lost. The test doesn't pick it up now, because the final message is sent _after the queue is exhausted again_.

This case is unlikely to happen on hardware because we don't just fire the timer a bunch as fast as we can, it's at a metered rate. The fix is to change the test so instead of
- fire 9 times
- check messages
- fire last time
- check completion

it's 
- for poll_count-1 iterations
   -  fire the timer
   - check that two messages were enqueued
- fire the timer for the last time
- check that the poll is complete

PR #392, opened 2022-06-17
Data ready interrupt

caila-marashaj:
General overview of the functionality here:

- `pressure_sensor_task` gets a request to read from the sensor over CAN
- `pressure_driver` writes a command to the pressure sensor to start measuring, but it does not send data yet
- the data ready pin gets triggered after every measurement
- data ready pin triggers a HAL interrupt function, which calls the callback function in `pressure_driver.hpp`
- callback function sends a command to get the data from the sensor, another command to stop the reading, and also grabs the data from the I2C bus
- the pressure driver then does some quick formatting and sends the result to CAN

Another note: the new functions in `writer.hpp` are there to allow us to get data from the sensor during the interrupt


pipettes/firmware/stm32l5xx_it.c
```diff
@@ -124,6 +124,17 @@ void DebugMon_Handler(void)
 /*  file (startup_stm32l5xx.s).                                               */
 /******************************************************************************/
 
+void EXTI8_IRQHandler(void) {
+    HAL_GPIO_EXTI_IRQHandler(GPIO_PIN_8);
+}
+
+void EXTI9_IRQHandler(void) {
```
Laura-Danielle, line 131: We should probably conditionally initialize these depending on which pipette is loaded in.

pipettes/firmware/utility_gpio.c
```diff
@@ -142,10 +142,12 @@ void data_ready_gpio_init() {
         /*Configure GPIO pin*/
         GPIO_InitTypeDef GPIO_InitStruct = {0};
         GPIO_InitStruct.Pin = hardware_rear.pin;
-        GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
+        GPIO_InitStruct.Mode = GPIO_MODE_IT_FALLING;
         GPIO_InitStruct.Pull = GPIO_PULLUP;
         GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
         HAL_GPIO_Init(hardware.port, &GPIO_InitStruct);
+
+        // TODO: call GPIO_EXTI_IRQ_Handler() and implement HAL_GPIO_EXTI_Callback()
```
Laura-Danielle, line 150: Can we get rid of this comment now?

pipettes/firmware/utility_gpio.c
```diff
@@ -154,10 +156,16 @@ void data_ready_gpio_init() {
...
-    GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
+    GPIO_InitStruct.Mode = GPIO_MODE_IT_FALLING;
     GPIO_InitStruct.Pull = GPIO_PULLUP;
     GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
     HAL_GPIO_Init(hardware.port, &GPIO_InitStruct);
+
+    /* EXTI interrupt init*/
+
+    // TODO: figure out which arg is the priority
+    HAL_NVIC_SetPriority(EXTI15_IRQn, 10, 0);
```
Laura-Danielle, line 167:
```/**
  * @brief  Sets the priority of an interrupt.
  * @param  IRQn: External interrupt number.
  *         This parameter can be an enumerator of IRQn_Type enumeration
  *         (For the complete STM32 Devices IRQ Channels list, please refer to the appropriate CMSIS device file (stm32f2xxxx.h))
  * @param  PreemptPriority: The preemption priority for the IRQn channel.
  *         This parameter can be a value between 0 and 15
  *         A lower priority value indicates a higher priority 
  * @param  SubPriority: the subpriority level for the IRQ channel.
  *         This parameter can be a value between 0 and 15
  *         A lower priority value indicates a higher priority.          
  * @retval None
  */
void HAL_NVIC_SetPriority(IRQn_Type IRQn, uint32_t PreemptPriority, uint32_t SubPriority)
```

Priority information for NVIC. Also you'll need to conditionally enable all three interrupts again, otherwise it won't work.

Out of curiosity, what is the third data ready pin connected to?

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -261,6 +261,7 @@ class MMR92C04 {
     CanClient &can_client;
     OwnQueue &own_queue;
     hardware::SensorHardwareBase &hardware;
+    bool data_ready = false;
```
Laura-Danielle, line 264: I think we don't need this right?

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -116,21 +116,14 @@ class MMR92C04 {
...
     }
 
-    auto get_temperature(bool poll = false, uint16_t sample_rate = 0) -> void {
-        if (poll) {
-            poll_read(mmr920C04::Registers::TEMPERATURE_READ, sample_rate);
-        } else {
-            transact(mmr920C04::Registers::TEMPERATURE_READ);
-        }
+    auto get_temperature() -> void {
+        uint32_t data = 0x0;
```
Laura-Danielle, line 125: Can you add some big TODOs here about implementing proper polling in a follow-up PR?

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -219,6 +212,11 @@ class MMR92C04 {
         can_client.send_can_message(get_host_id(), message);
     }
 
+    auto drdy_interrupt_response() -> void {
```
Laura-Danielle, line 215: I think this needs to be static here since we're calling it from the callback function. @sfoster1 correct me if I'm wrong.

include/sensors/firmware/pressure_sensor_callbacks.h
```diff
@@ -0,0 +1,11 @@
+#include "platform_specific_hal_conf.h"
+#include "sensors/core/tasks/pressure_driver.hpp"
+
```
Laura-Danielle, line 3: Still not great at lifetimes of things. @sfoster1 is it better to have a global variable which the callback gets saved to, or is it fine to have `drdy_interrupt_response` be a static function on the pressure driver class?

Laura-Danielle (commented):
Couple of initial comments.

pipettes/firmware/utility_gpio.c
```diff
@@ -134,18 +135,22 @@ void data_ready_gpio_init() {
         hardware =
             pipette_hardware_get_gpio(
         pipette_type, pipette_hardware_device_data_ready);
-        PipetteHardwarePin hardware_rear =
-        pipette_hardware_get_gpio(
-            pipette_type, pipette_hardware_device_data_ready);
+        PipetteHardwarePin hardware_rear;
+        hardware_rear.port = GPIOA;
```
sfoster1, line 139: I think this should probably still be using `pipette_hardware_get_gpio`, right?

sfoster1 (approved):
Happy with this! But before merging, we should make sure everything possible is conditionally initialized only if we're the right kind of pipette.

include/sensors/firmware/sensor_hardware.hpp
```diff
@@ -11,11 +13,30 @@ class SensorHardware : public SensorHardwareBase {
         : hardware(hardware) {}
     auto set_sync() -> void override { gpio::set(hardware.sync_out); }
     auto reset_sync() -> void override { gpio::reset(hardware.sync_out); }
-    auto check_data_ready() -> bool override {
-        return gpio::is_set(hardware.data_ready);
+
+    auto data_ready() -> void {
```
Laura-Danielle, line 17: If we expect to have different callback functions at some point (either for multiple pressure sensors, or the environment sensor) then we should maybe have these be key value pairs where the key is the gpio pin. Otherwise, we could end up calling the wrong callback function. It works now because we only have 1 callback function here.

Laura-Danielle: If you don't want to address it now, just put a very big TODO here about it.

Laura-Danielle (approved):
Agree with seth's comment and we also need to either refactor the callback list a little bit to make sure we don't accidentally call the wrong callbacks or put a large TODO comment about it.

include/i2c/core/writer.hpp
```diff
@@ -73,6 +74,34 @@ class Writer {
...
+        return do_write_isr(device_address, sizeof(data), max_buffer);
+    }
+    template <typename Data>
+    requires std::is_integral_v<Data>
+    auto write_isr(uint16_t device_address, uint8_t reg, Data data) -> bool {
+        messages::MaxMessageBuffer max_buffer{};
+        auto* iter = max_buffer.begin();
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+        *iter++ = reg;
+        LOG("writer queue mem address = %d", queue);
```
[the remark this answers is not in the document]
caila-marashaj: no that guy slipped through the cracks. will remove

PR #393, opened 2022-06-17
chore(sensors, can): add sensor ids to can messages

ahiuchingau:
Sometimes we have multiples of the same type of sensor present in a subsystem and we need separate sensor IDs to differentiate them.

### DIFF
--- a/gripper/core/tasks.cpp
+++ b/gripper/core/tasks.cpp
@@ -43,9 +43,10 @@ void gripper_tasks::start_tasks(
     i2c3_task_client.set_queue(&i2c3_task.get_queue());
     queues.i2c3_queue = &i2c3_task.get_queue();
 
-    auto& eeprom_task = eeprom_task_builder.start(5, "eeprom", i2c3_task_client,
-                                                  eeprom_hw_iface);
-    tasks.eeprom_task = &eeprom_task;
+    auto& eeprom_task =
+        eeprom_task_builder
+            .start(5, "eeprom", i2c3_task_client, eeprom_hw_iface)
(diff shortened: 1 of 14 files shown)


include/can/core/ids.hpp
```diff
@@ -163,4 +163,11 @@ enum class PipetteTipActionType {
     drop = 0x1,
 };
 
+/**
```
Laura-Danielle, line 166: This needs to be generated from python otherwise it's going to create conflicts later. As long as you have your branch open in the monorepo it should work to use the `update-headers` command. FYI it will use the data string from there, so if you want all this text in the comment, you should update the python `constants.py` file to reflect that.

gripper/core/tasks.cpp
```diff
@@ -41,8 +41,13 @@ static auto i2c2_poll_client =
 static auto i2c3_poll_client =
     i2c::poller::Poller<freertos_message_queue::FreeRTOSMessageQueue>{};
 
-static auto capacitive_sensor_task_builder =
-    freertos_task::TaskStarter<512, sensors::tasks::CapacitiveSensorTask>{};
+static auto capacitive_sensor_task_builder_s0 =
```
sfoster1, line 44: can we give these names that relate to the physical use? like `capacitive_sensor_rear` and `_front` or whatever

sfoster1 (approved):
Looks good, but let's make the task names a little more specific and also test this on pipettes as well as gripper

Laura-Danielle (approved):
lgtm, but one comment about the generated header file.

PR #394, opened 2022-06-21
refactor(motor-control): Cleanup tip action messages from default motion controller

Laura-Danielle:
Some cleanup found while testing the 96 channel


sfoster1 (approved):
Looks good!

PR #395, opened 2022-06-23
feat(gantry): support rev1 boards

sfoster1:
Do similar refactors to what was needed to support head rev1 to bring up the rev1 x and y boards (actually the same board in both places). They, too, have a tmc2160 instead of a tmc2130, so the same changes are needed.

This is tested on the proto boards, but has yet to be tested on the actual new rev hardware.


Laura-Danielle (approved):
Looks good to me. I think it would be nice if we updated the build presets in `CMakePresets.json` to reflect the new target names but not a blocker.

sfoster1:
> Looks good to me. I think it would be nice if we updated the build presets in `CMakePresets.json` to reflect the new target names but not a blocker.

So that all revs were built when you do `cmake --build --preset=gantry`?

PR #396, opened 2022-06-29
chore: store firmware artifacts from actions

sfoster1:
Refactor cmake targets to provide default-named outputs for all projects as well as revision and variant outputs. We can then store those outputs in github actions storage so we can distribute and store controlled outputs rather than dev builds.


Laura-Danielle (approved):
Ignore my previous comment about hex combine. Everything looks good.

sfoster1:
saving the install thing for later i think.

PR #397, opened 2022-07-01
fix(head): wider gripper detection bounds

sfoster1:
They're like the A bounds (though this only matters on the low side).

WIth this pr in place we now properly detect the gripper.


Laura-Danielle (approved):
Nice

PR #400, opened 2022-07-06
refactor(motor-control): setup automatically

sfoster1:
We have this setup request canbus message that commands all the micros
to do a bunch of motor driver configuration that they have built in
anyway. Rather than require this, and therefore run the risk of a
programmer forgetting to do it, let's just make the setuprequest happen
internally without prompting.

## Testing
- [ ] should be tested on all micros along with a followup that enables motors when they begin to move; then some basic repl commands can be run without enabling first.


Laura-Danielle (approved):
lgtm. We should try to merge this pr and python side changes around the same time so folks don't get confused

sfoster1:
> We should try to merge this pr and python side changes around the same time so that the repos are in sync

This can be merged first and be fine - it would just ignore the setup request now

sfoster1:
OK, this works on my test machine - all axes manage to move without any setups!

PR #401, opened 2022-07-06
refactor(motor-control): Enable automatically

sfoster1:
When the motors are about to move, they should get enabled without
requiring a previous enable request. This moves the enable motors
message into something used only for turning on motors and holding, as
in the OT-2, rather than a necessary step before turning motors -
removes the possibility for programmer error.


include/motor-control/core/stepper_motor/motion_controller.hpp
```diff
@@ -172,9 +182,13 @@ class PipetteMotionController {
     void enable_motor() {
         hardware.start_timer_interrupt();
         hardware.activate_motor();
+        enabled = true;
```
Laura-Danielle, line 185: Need to actually enable motor in the TipRequest move handler.

Laura-Danielle (approved):
looks good to me aside from the one comment

PR #402, opened 2022-07-06
refactor(gripper): remove pwm freq via can

sfoster1:
This isn't necessary for gripper workflows.

In addition, we can derive the clock dividers for the pwm driver using
the global from the system file that contains the system clock frequency.

Needs https://github.com/Opentrons/opentrons/pull/10987

### DIFF
--- a/gripper/firmware/motor_timer_hardware.c
+++ b/gripper/firmware/motor_timer_hardware.c
@@ -19,6 +19,16 @@ TIM_OC_InitTypeDef htim3_sConfigOC = {0};
 static motor_interrupt_callback timer_callback = NULL;
 static brushed_motor_interrupt_callback brushed_timer_callback = NULL;
 
+uint32_t clamp(uint32_t val, uint32_t min, uint32_t max) {
+    if (val < min) {
+        return min;
+    }
+    if (val > max) {
+        return max;
+    }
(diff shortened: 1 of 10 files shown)


caila-marashaj (approved):

gripper/firmware/motor_timer_hardware.c
```diff
@@ -10,6 +19,16 @@ TIM_OC_InitTypeDef htim3_sConfigOC = {0};
 static motor_interrupt_callback timer_callback = NULL;
 static brushed_motor_interrupt_callback brushed_timer_callback = NULL;
 
+uint32_t clamp(uint32_t val, uint32_t min, uint32_t max) {
```
DerekMaggio, line 22: By `clamp` you mean set to within the bounds of min and max? Not actually actuate the gripper right? 

Dumb question I know.

DerekMaggio (approved):
So if I understand correctly you are removing the ability to move the motor at different frequencies and are instead just hardcoding it to 2/3s of the set PWM width?

gripper/firmware/motor_timer_hardware.c
sfoster1: yeah correct. numeric clamp

PR #403, opened 2022-07-07
fix(gripper): do not stop motor when moves "end"

sfoster1:
Because moves don't really end in the same way as a stepper - we want to
continue applying the force, we just need to send an ack to indicate
we're in more or less the correct location.

### DIFF
--- a/include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
+++ b/include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
@@ -53,12 +53,10 @@ class BrushedMotorInterruptHandler {
             if (buffered_move.stop_condition ==
                     MoveStopCondition::limit_switch &&
                 limit_switch_triggered()) {
-                hardware.stop_pwm();
                 homing_stopped();
             } else {
                 tick_count++;
                 if (!should_continue()) {
-                    hardware.stop_pwm();
                     finish_current_move(
                         AckMessageId::complete_without_condition);
                 }


include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -53,12 +53,10 @@ class BrushedMotorInterruptHandler {
             if (buffered_move.stop_condition ==
                     MoveStopCondition::limit_switch &&
                 limit_switch_triggered()) {
-                hardware.stop_pwm();
```
Laura-Danielle, line 56: Don't we want to not apply force if we're at home?

sfoster1:
> Larger overarching question. Why do we still want to apply force when movement stops?

The way we define moves in the system is tailored for position control. A move starts when the system is at a fixed position; when the system reaches its destination fixed position, the move ends. Subsequent moves must begin after the prior moves end, because until the prior moves end the subsequent move won't be correct. If you consider the stepper (and other position-control) systems as a state machine, the states are positions and the edges are moves.

The gripper jaw's force control modes are very different, because it's a force control system. If you consider the gripper jaw as a state machine, its states are continuous applications of force and its edges are moving between applied forces.

A gripper jaw move, the edges in its state machine, should terminate when the gripper senses that it is now applying the requested amount of force through its jaw. And when it terminates, what that means is that the gripper's current state is "applying X force", not "holding X position", and in the same way that the stepper holds its X position, the gripper holds its X force.

PR #404, opened 2022-07-11
Refactor(eeprom): refactor the hardware interface and task to support 16bit and 8bit memory addresses

ryanthecoder:
refactor the eeprom hardware interface so that it can be configured to use 16 bit memory addresses. still defaults to the old 8 bit addresses so it doesn't require changes in any of the current code outside of the eeprom library

### DIFF
--- a/eeprom/tests/test_eeprom_task.cpp
+++ b/eeprom/tests/test_eeprom_task.cpp
@@ -8,7 +8,8 @@
 #include "i2c/tests/mock_response_queue.hpp"
 
 class MockHardwareIface : public eeprom::hardware_iface::EEPromHardwareIface {
-  using eeprom::hardware_iface::EEPromHardwareIface::EEPromHardwareIface;
+    using eeprom::hardware_iface::EEPromHardwareIface::EEPromHardwareIface;
+
   public:
     void set_write_protect(bool enabled) { set_calls.push_back(enabled); }
     std::vector<bool> set_calls{};
@@ -24,6 +25,11 @@ SCENARIO("Sending messages to Eeprom task") {
     auto eeprom = eeprom::task::EEPromMessageHandler{writer, response_queue,
                                                      hardware_iface};
 
+    auto hardware_iface_16 =
+        MockHardwareIface(eeprom::hardware_iface::EEPROM_ADDR_16_BIT);
+
+    auto eeprom_16 = eeprom::task::EEPromMessageHandler{writer, response_queue,
+                                                       hardware_iface_16};
     GIVEN("A write message") {
         auto data = eeprom::types::EepromData{1, 2, 3, 4};
         eeprom::types::address address = 12;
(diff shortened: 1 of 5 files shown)


include/eeprom/core/hardware_iface.hpp
```diff
@@ -3,12 +3,20 @@
 namespace eeprom {
 namespace hardware_iface {
 
+static const size_t EEPROM_ADDR_8_BIT = sizeof(uint8_t);
```
sfoster1, line 6: can we make these special elements of an `enum class` or something so they can be typed as like `EnumAddressType`

sfoster1 (approved):
Nice, good tests particularly. I'd love it if those addresses could be enum classes, and we should make sure to change the PR title (and thus the squash merge result) to something nicer before pressing the button, but looks good to me!

eeprom/tests/test_eeprom_task.cpp
```diff
@@ -7,7 +7,10 @@
 #include "i2c/core/writer.hpp"
 #include "i2c/tests/mock_response_queue.hpp"
 
-struct MockHardwareIface : public eeprom::hardware_iface::EEPromHardwareIface {
+class MockHardwareIface : public eeprom::hardware_iface::EEPromHardwareIface {
+    using eeprom::hardware_iface::EEPromHardwareIface::EEPromHardwareIface;
```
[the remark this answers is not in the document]
ryanthecoder: It needs to be a class so it can inherit the base class's constructor with the 'using eeprom...' line below.

include/eeprom/core/hardware_iface.hpp
```diff
@@ -3,12 +3,28 @@
 namespace eeprom {
 namespace hardware_iface {
 
+enum EEPromAddressType {
```
sfoster1, line 6: nit: if we make these `enum class` we don't have to have c-style NAMESPACE_PREFIX_ELEMENT member names

include/eeprom/core/hardware_iface.hpp
```diff
@@ -3,12 +3,28 @@
...
+ *
+ * Consecutive write requests will want to have write protection disabled. The
+ * enable should happen when all the write transactions have completed.
+ *
  */
 class EEPromHardwareIface {
   public:
     EEPromHardwareIface() = default;
+    EEPromHardwareIface(const size_t addr_bytes)
+        : eeprom_addr_bytes(addr_bytes) {
```
sfoster1, line 23: i'd love to see an init-list that explicitly specifies all the values based on the argument, ideally by using a `switch` statement so the compiler yells at us if we add enum elements and don't handle it

sfoster1 (approved):
Looks overall good, though i'd love to see some of the detailed changes.

include/eeprom/core/hardware_iface.hpp
ryanthecoder: made the other changes and I think I did what you meant here. I also cleaned up a bunch of extra namespace things we didn't need.

include/eeprom/core/hardware_iface.hpp
```diff
@@ -3,12 +3,28 @@
...
  * Interface to eeprom. Must be implemented in FW and Simulation
+ *
+ * Consecutive write requests will want to have write protection disabled. The
+ * enable should happen when all the write transactions have completed.
+ *
  */
 class EEPromHardwareIface {
   public:
     EEPromHardwareIface() = default;
+    EEPromHardwareIface(const size_t addr_bytes)
```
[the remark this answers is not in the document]
Laura-Danielle: Couldn't we just make this into a template class dependent on a concept?

sfoster1 (approved):
Awesome, love all the tests and the way the interface is shaping up.

PR #406, opened 2022-07-12
feat(motor-control,head,gripper): encoder bringup

sfoster1:
Supercedes #362 

Bring up the quadrature encoder handling core code and integrate it into the head (tested) and the gripper (not as well tested, and requires a fast-follow anyway).

This is a large PR that should be read commit-by-commit, including the messages which is where there's more detail about implementation.


include/motor-control/core/utils.hpp
```diff
@@ -17,6 +17,10 @@ auto fixed_point_multiply(sq31_31 a, sq0_31 b) -> sq0_31;
 
 auto fixed_point_multiply(sq31_31 a, uint32_t b) -> uint32_t;
 
+struct radix_offset_0 {};
```
Laura-Danielle, line 20: Is this a workaround of some sort?

include/can/core/messages.hpp
```diff
@@ -350,6 +352,20 @@ struct MoveCompleted : BaseMessage<MessageId::move_completed> {
     auto operator==(const MoveCompleted& other) const -> bool = default;
 };
 
+struct EncoderPositionResponse
```
Laura-Danielle, line 355: When would we want encoder position data outside of a completed move?

include/motor-control/core/linear_motion_system.hpp
```diff
@@ -29,6 +29,7 @@ struct LinearMotionSystemConfig {
     MEConfig mech_config{};
     float steps_per_rev{};
     float microstep{};
+    float encoder_ppr{0.0};
```
Laura-Danielle, line 32: could we make this `encoder_pulse_per_rev` to make it more in line with the other variable names here?

motor-control/firmware/motor_control_hardware.c
```diff
@@ -86,7 +86,12 @@ uint32_t motor_hardware_encoder_pulse_count(void* enc_htim) {
 }
 
 void motor_hardware_reset_encoder_count(void* enc_htim) {
-    if (enc_htim != NULL) {
-        __HAL_TIM_SET_COUNTER((TIM_HandleTypeDef*)enc_htim, 0);
+    /* Offset the zero position by 10 pulses to eliminate any noise on the system.
+     */
+    bool overflow_flag = __HAL_TIM_GET_FLAG((TIM_HandleTypeDef*)enc_htim, TIM_FLAG_UPDATE);
+    while(overflow_flag){
```
Laura-Danielle, line 92: Can you explain what's going on here a little more. When an overflow happens, the flag will be 1 and then we immediately update again? Is this handling an edge case of some sort?

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
```diff
@@ -227,16 +224,17 @@ class MotorInterruptHandler {
...
 
     [[nodiscard]] static auto overflow(q31_31 current, q31_31 future) -> bool {
         /*
          * Check whether the position has overflowed. Return true if this has
          * happened, and false otherwise.
          *
-         * (TODO lc): If an overflow has happened, we should probably notify the
-         * server somehow that the position needed to be capped.
-         * Note: could not get the built-ins to work as expected.
+         * (TODO lc): If an overflEncoderHandlerow has happened, we should
```
Laura-Danielle, line 235: typo in comment?

motor-control/firmware/motor_control_hardware.c
sfoster1: This is the reset that gets called after home; this code resets the overflow flag in case it happened to be set at that time. The loop I guess makes sure it's cleared. I don't love this, and I think I'll remove it.

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
sfoster1: a find and replace gone awry i assume

include/motor-control/core/utils.hpp
sfoster1: Yeah, unfortunately c++ overloaded function dispatch considers `typedef`s to be the same. So `fixed_point_multiply(sq31_31 a, sq0_31 b)` is the same as `fixed_point_multiply(sq31_31 a, int32_t b)`. This dummy struct changes the signature enough that you can specify which function you want, since they are in fact quite different.

Laura-Danielle: Do you mind putting a small comment as to why this is needed?

PR #407, opened 2022-07-12
refactor(pipettes): support g4 rev1 board pipettes

Laura-Danielle:
## Overview

Related to #RET-848. For now, we want to support both types of board types. The default for the `pipettes` target is still going to be the proto board (L5s) until we start getting in the rev1 boards for the pipettes. I decided not to upload `rev1` artifacts to s3 until it can be tested on physical hardware. 

This PR does _not_ change pin configurations or handle any other changes around timers. That will happen in a follow-up PR.

## Review Requests

Ok with this setup? Any issues with how the build presets were handled? Other things?


bootloader/firmware/stm32G4/node_id_stm32g4xx.c
```diff
@@ -1,4 +1,107 @@
+#include <stdbool.h>
+#include <stdint.h>
 #include "bootloader/core/node_id.h"
+#include "stm32g4xx_hal.h"
+
+#if defined(node_id_pipette_dynamic)
```
sfoster1, line 6: when wouldn't it be dynamic?

sfoster1 (approved):
Nice!

PR #408, opened 2022-07-13
Pressure Sensor Simulator Fix

caila-marashaj:
In main currently, we have two files that simulate the SensorHardware class and do almost the exact same thing(`include/sensors/simulation/hardware.hpp` and `include/sensors/tests/mock_hardware.hpp`), so I consolidated those into `mock_hardware.hpp` and moved that into the simulation directory.

Also, got the pressure sensor simulator sending response messages over CAN by adding an override of the I2C simulator's `handle_write`inside the `mmr920C04` simulator class.


sfoster1 (approved):
Ah, makes sense. Nice!

PR #409, opened 2022-07-14
refactor(pipettes): update the G4 pin values for low throughput pipettes

Laura-Danielle:
## Overview

Related to #RET-848. For now, we only have the pin configurations available for single/eight channel pipettes. This PR only addresses pin changes, and no functionality change around timers or EXTI callbacks that may need to change.


sfoster1 (approved):
nice!

PR #411, opened 2022-07-14
fix(gantry): fix rev1 direction and handle current better

sfoster1:
Two fixes for the revision 1 gantry boards:
- y needs its direction swapped
- the tmc2160 motor drivers were only paying attention to the hold current which usually is not high enough for motion (that's the point). for now, just care about the run; in the future, we'll probably want fancier behavior.


include/motor-control/core/tasks/tmc2160_motor_driver_task.hpp
```diff
@@ -86,10 +86,7 @@ class MotorDriverMessageHandler {
             "run_current=%d",
             m.hold_current, m.run_current);
 
-        if (m.hold_current != 0U) {
-            driver.get_register_map().glob_scale.global_scaler =
-                driver.convert_to_tmc2160_current_value(m.hold_current);
-        } else if (m.run_current != 0U) {
+        if (m.run_current != 0U) {
```
Laura-Danielle, line 89: oops

Laura-Danielle (approved):
nice

PR #412, opened 2022-07-18
Pressure Sensor Polling Support

caila-marashaj:
This branch implements some support for polling on the firmware side:

- calls driver functions when the `pressure_sensor_task` receives a `BaselineSensorRequest` or `BindSensorOutputRequest`
- sets the sync pin active when the pressure driver receives a measurement above a threshold value
- makes sending a `RESET` command during the pressure driver's interrupt conditional on the pressure driver performing a single read, so polling doesn't get cut short after the first measurement

### DIFF
--- a/include/sensors/core/tasks/pressure_driver.hpp
+++ b/include/sensors/core/tasks/pressure_driver.hpp
@@ -128,10 +128,7 @@ class MMR920C04 {
         }
     }
 
-    auto get_pressure(mmr920C04::SensorMode mode =
-                          mmr920C04::SensorMode::SINGLE_READ) -> bool {
-        //        _output_sync_bind = sync_bind;
-        _sensor_mode = mode;
+    auto get_pressure() -> bool {
         return set_measure_mode(mmr920C04::Registers::MEASURE_MODE_4);
     }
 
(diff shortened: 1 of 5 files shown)


include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -233,10 +236,15 @@ class MMR920C04 {
...
+                             static_cast<uint8_t>(mmr920C04::Registers::RESET),
+                             data);
+        }
+    }
 
-        writer.write_isr(mmr920C04::ADDRESS,
-                         static_cast<uint8_t>(mmr920C04::Registers::RESET),
-                         data);
+    auto set_sync_bind(SensorOutputBinding binding) -> void {
+        _output_sync_bind = binding;
```
sfoster1, line 247: we should also call `hardware.reset_sync()` in this function, because
- if this binding contains `::sync`, then we should make sure we start with it reset
- if we were sync-bound, and had raised sync, and this is `::none`, we won't have another opportunity to reset it

sfoster1 (changes requested):
Couple inline changes, but also
- does this handle writing back data to can if the "echo" flag is set?
- tests

caila-marashaj:
> Nice! LGTM if it works. This lets us stream data right?



> Nice! LGTM if it works. This lets us stream data right?

yes! It will send the data back over CAN

PR #414, opened 2022-07-20
fix(pipettes): switch I2C buses for sensors

Laura-Danielle:
## Overview
For the G4 board, the primary sensors live on I2C3 bus now and the eeprom is on the I2C2 bus.

## Changelog
- Add a separate sensor task file for the G4
- Modify the CMakeLists.txt files to load the sensor file based on revision

## TODO

- [x] Test on the G4 board


pipettes/core/CMakeLists.txt
```diff
@@ -1,41 +1,47 @@
...
+function(target_pipettes_core_common TARGET REVISION)
     target_sources(${TARGET} PUBLIC
             ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/configs.cpp
             ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/peripheral_tasks.cpp
             ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/central_tasks.cpp
             ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/linear_motor_tasks.cpp
-            ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/gear_motor_tasks.cpp
-            ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/sensor_tasks.cpp)
+            ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/gear_motor_tasks.cpp)
+    if (${REVISION} STREQUAL "rev1")
```
sfoster1, line 8: we might make further revisions; proto would be the odd one out. maybe invert this so it's `if (${REVISION} STREQUAL "proto") ... else` and that way if we make a `rev2` we won't have to change it

sfoster1 (approved):
lgtm! maybe invert the logic in common but this should work

PR #418, opened 2022-07-22
feat: OT3 State Manager

DerekMaggio:
# Overview

Creating an OT3 State Manager in Python which maintains the current state of the OT3, and can pass states between peripherals.

[Requirements Doc](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3322773722/OT3+Emulation+State+Manager#SystemStateDataModel-%5BJIRA-Ticket%5D)


DerekMaggio:
@sfoster1 I am still working on setting up Github Actions and getting formatting and linting to pass. 

In the meantime, can you take a look at how I implemented this and let me know what you think?

ot3_state_manager/ot3_state_manager/hardware.py
```diff
@@ -0,0 +1,104 @@
...
+
+from ot3_state_manager.ot3_state_manager.measurable_states import Force, Position
+from ot3_state_manager.ot3_state_manager.ot3_axis import OT3Axis
+from ot3_state_manager.ot3_state_manager.pipette_model import PipetteModel
+
+
+class UpdatablePositions(Enum):
+    """Represents positions that can be changed on the OT3"""
+
+    GANTRY_X = "Gantry X"
```
sfoster1, line 17: why do we need this instead of just using the axis?

ot3_state_manager/ot3_state_manager/hardware.py
```diff
@@ -0,0 +1,104 @@
...
+            axis = pos_dict[pos]
+        except KeyError:
+            raise Exception(f"Hardware {pos.name} does not exist")
+        return axis
+
+
+class Hardware(ABC):
+    """ABC for all OT3 hardware."""
+
+    hardware_name: str
```
sfoster1, line 50: this probably should be an enum element; we shouldn't use strings for programmatic comparison

ot3_state_manager/ot3_state_manager/measurable_states.py
```diff
@@ -0,0 +1,23 @@
...
+    axis: OT3Axis
+    current_position: float = 0.0
+    commanded_position: float = 0.0
+    encoder_position: float = 0.0
+
+
+@dataclass
+class Force:
+    """Represents force state."""
+
```
sfoster1, line 20: could reasonably have an axis in here too

ot3_state_manager/ot3_state_manager/ot3_state_manager.py
```diff
@@ -0,0 +1,262 @@
...
+            )
+
+        self.head = Head()
+        self.gantry_x = GantryX()
+        self.gantry_y = GantryY()
+        self.left_pipette = left_pipette
+        self.right_pipette = right_pipette
+        self.gripper = gripper
+
+    def _pos_dict(self) -> Dict[str, Optional[Position]]:
```
sfoster1, line 101: on the topic of strings - here for instance these really should not be strings. we should live in axis land as much as possible and convert only on boundaries, if there.

ot3_state_manager/ot3_state_manager/ot3_state_manager.py
```diff
@@ -0,0 +1,262 @@
...
+    @property
+    def is_high_throughput_pipette_attached(self) -> bool:
+        """Returns True if high throughput pipette is attached, False if not."""
+        return (
+            self.right_pipette is not None
+            and self.right_pipette.model == PipetteModel.MULTI_96_1000
+        )
+
+    @staticmethod
+    def _update_pos(
```
sfoster1, line 219: the position update seems a little complex - `_update_pos` and therefore `update_position` can do a couple different things depending how it's called.

I think it's a better idea to have one way to update position that requires everything. That's easier to typecheck and verify the correct use of - just by removing the `Optional`, we can remove the need for `NoPositionProvidedError`. Also, the decision about what to update is ultimately not going to rest here, so let's keep the logic of what to update and what not to update above.

ot3_state_manager/ot3_state_manager/pipette_model.py
```diff
@@ -0,0 +1,15 @@
+"""Enum containing valid pipette models."""
+from enum import Enum
+
+
+class PipetteModel(Enum):
+    """Enum containing valid pipette models for OT3."""
+
+    # TODO: I know these aren't the correct pipette models
+    # Need to update when there is more clarity on what they are
+    # Maybe pull from shared data in monorepo?
```
sfoster1, line 10: pulling them from shared data is almost certainly the right thing, but yeah this is definitely not something we need for this pr.

sfoster1 (changes requested):
Style things:
- I personally don't love the incredibly specific and prescriptive exception types and am generally more in favor of consistent exception usage - if this isn't an allowable value, use `ValueError("x must be less than 10")`, not `ValuesLessThanTenNotAcceptedInThisFunctionError`, but we don't have a lot of them yet so whatever
- Let's get rid of some of these strings; generally better to use enum elements. If we really really need to keep them strings, they should be annotated as `Literal` or `Final`.
Structural/package management things:
- Can we try using poetry and a pyproject.toml? If we're only running this locally or pushing to the robot (i.e. as opposed to adding it to the build) then it should be fine, and even if we add it to the build hopefully by then we'll have updated yocto
- I dont think `ot3_state_manager/__init__.py` should exist, right? just `ot3_state_manager/ot3_state_manager/__init__.py`

Overall, looks great! Just some little things I'd like to see changed.

ot3_state_manager/.flake8
```diff
@@ -0,0 +1,28 @@
+[flake8]
+
+# set line-length for black support
+# https://github.com/psf/black/blob/master/docs/compatible_configs.md
```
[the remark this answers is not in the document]
DerekMaggio: https://github.com/Opentrons/ot3-firmware/pull/418/commits/bcf8088c643d236dde9bbdf404caff9db0e0b88f

ot3_state_manager/.flake8
```diff
@@ -0,0 +1,28 @@
...
+max-line-length = 88
+
+# max cyclomatic complexity
+max-complexity = 9
+
+extend-ignore =
+    # ignore E203 because black might reformat it
+    E203,
```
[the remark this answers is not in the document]
DerekMaggio: https://github.com/Opentrons/ot3-firmware/pull/418/commits/bcf8088c643d236dde9bbdf404caff9db0e0b88f

ot3_state_manager/.flake8
```diff
@@ -0,0 +1,28 @@
...
+    # do not require type annotations for self nor cls
+    ANN101,
+    ANN102,
+    # who wants an __init__ method docstring? You are creating the object
```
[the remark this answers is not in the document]
DerekMaggio: https://github.com/Opentrons/ot3-firmware/pull/418/commits/bcf8088c643d236dde9bbdf404caff9db0e0b88f

ot3_state_manager/setup.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    Build an absolute path from *parts* and and return the contents of the
+    resulting file.  Assume UTF-8 encoding.
+    """
+    with codecs.open(os.path.join(HERE, *parts), "rb", "utf-8") as f:
+        return f.read()
+
+
+setup(
+    python_requires=">=3.10",
```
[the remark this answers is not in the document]
DerekMaggio: > * I personally don't love the incredibly specific and prescriptive exception types and am generally more in favor of consistent exception usage - if this isn't an allowable value, use `ValueError("x must be less than 10")`, not `ValuesLessThanTenNotAcceptedInThisFunctionError`, but we don't have a lot of them yet so whatever

https://github.com/Opentrons/ot3-firmware/pull/418/commits/e4234f299ffc2b147daa1eea5dc1b055c4243d13

DerekMaggio:
> * I personally don't love the incredibly specific and prescriptive exception types and am generally more in favor of consistent exception usage - if this isn't an allowable value, use `ValueError("x must be less than 10")`, not `ValuesLessThanTenNotAcceptedInThisFunctionError`, but we don't have a lot of them yet so whatever

https://github.com/Opentrons/ot3-firmware/pull/418/commits/e4234f299ffc2b147daa1eea5dc1b055c4243d13

DerekMaggio:
> * I dont think `ot3_state_manager/__init__.py` should exist, right? just `ot3_state_manager/ot3_state_manager/__init__.py`

https://github.com/Opentrons/ot3-firmware/pull/418/commits/bcf8088c643d236dde9bbdf404caff9db0e0b88f

Laura-Danielle (changes requested):
I _really_ don't think this should be in `ot3-firmware` if we plan to rely on many things from the monorepo. At the minimum, we'll need all the plumbing from `hardware/` around sending/receiving raw can messages. Is the ultimate plan that we would be separately issue commands from the hardware controller and then use this to keep track of state?

1. We should probably not have to copy all of the enum types over (such as `OT3Axis`) because it becomes a pain to keep parity between the monorepo + ot3-firmware.
2. Unless we're fully planning to only check the raw values of all the CAN output messages (which is again a pain, and I don't see the benefit), we will want some of the helpful conversions used in the `hardware/` folder as well.

sfoster1:
also IMO it'd be nice to have `make` used only here and nowhere else and instead contrive a way to get this stuff into cmake. things that are makefile targets here could be `custom_command`s in cmake.

sfoster1:
> > * Can we try using poetry and a pyproject.toml? If we're only running this locally or pushing to the robot (i.e. as opposed to adding it to the build) then it should be fine, and even if we add it to the build hopefully by then we'll have updated yocto
> 
> @sfoster1 why should I use poetry over pipenv? To me, it makes sense to use pipenv since we are using it in our other repos.

Because it's slow and it has hard to diagnose locking failures when you're trying to update things. There's also a lot of progress being made on poetry plugins that will come in really handy for more monorepo focused structures, and it lets you use pyproject.toml which can concentrate all the dotfiles into one.

sfoster1:
> I _really_ don't think this should be in `ot3-firmware` if we plan to rely on many things from the monorepo. At the minimum, we'll need all the plumbing from `hardware/` around sending/receiving raw can messages. Is the ultimate plan that we would be separately issue commands from the hardware controller and then use this to keep track of state?
> 
>     1. We should probably not have to copy all of the enum types over (such as `OT3Axis`) because it becomes a pain to keep parity between the monorepo + ot3-firmware.
> 
>     2. Unless we're fully planning to only check the raw values of all the CAN output messages (which is again a pain, and I don't see the benefit), we will want some of the helpful conversions used in the `hardware/` folder as well.

I agree with the mechanical issues with having this elsewhere, but the functionality of it is so bound up with specifically the firmware that it's hard to see it working well in the monorepo either.

I wonder if the thing that can solve the problem here is explicitly requiring the monorepo to be present somewhere and depending on it directly. We can get the path configured in cmake if we use it and explicitly depend on it from poetry or pipenv or whatever, and then use stuff like the axis definitions.

Does that make things better?

Laura-Danielle:
> 

I think I am fine with that. I am still curious about the purpose of the state manager in general. Do we expect this to behave like a redux layer where the hardware controller issues commands, and the state manager monitors the can bus?

ot3_state_manager/ot3_state_manager/hardware.py
DerekMaggio: 100b27c5fe94daf9508ee98fcb3e0ecc46012cec

ot3_state_manager/ot3_state_manager/hardware.py
DerekMaggio: Made all child classes use an enum element. This value has to be a string since the underlying enum value is a string

ot3_state_manager/ot3_state_manager/ot3_state_manager.py
DerekMaggio: 100b27c5fe94daf9508ee98fcb3e0ecc46012cec

sfoster1:
> > 
> 
> I think I am fine with that. I am still curious about the purpose of the state manager in general. Do we expect this to behave like a redux layer where the hardware controller issues commands, and the state manager monitors the can bus?

No - the idea here is that the robot server talks to simulators on can, and the state manager is on the other side, coordinating the responses of the simulators based on whole-system state concerns. some new interface will connect the simulators and the state manager, and the robot server won't know about the state manager at all

DerekMaggio:
> > > * Can we try using poetry and a pyproject.toml? If we're only running this locally or pushing to the robot (i.e. as opposed to adding it to the build) then it should be fine, and even if we add it to the build hopefully by then we'll have updated yocto
> > 
> > 
> > @sfoster1 why should I use poetry over pipenv? To me, it makes sense to use pipenv since we are using it in our other repos.
> 
> Because it's slow and it has hard to diagnose locking failures when you're trying to update things. There's also a lot of progress being made on poetry plugins that will come in really handy for more monorepo focused structures, and it lets you use pyproject.toml which can concentrate all the dotfiles into one.

https://github.com/Opentrons/ot3-firmware/pull/418/commits/abfbd1187b7316c78df4a3f939d8ae9ed854b19d

ot3_state_manager/ot3_state_manager/hardware.py
DerekMaggio: Removed 22396f7a6c90521777af37f8084b954144201857

ot3_state_manager/ot3_state_manager/hardware.py
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        try:
+            axis = pos_dict[pos]
+        except KeyError:
+            raise Exception(f"Hardware {pos.name} does not exist")
+        return axis
+
+
+class Hardware(ABC):
```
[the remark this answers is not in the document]
DerekMaggio: 22396f7a6c90521777af37f8084b954144201857

ot3_state_manager/setup.py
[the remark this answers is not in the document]
DerekMaggio: Changed to 3.8 https://github.com/Opentrons/ot3-firmware/pull/418/commits/abfbd1187b7316c78df4a3f939d8ae9ed854b19d

ot3_state_manager/ot3_state_manager/measurable_states.py
DerekMaggio: c588d69673f561cba5773dff8239cf1593298f69
Is this correct @sfoster1?

ot3_state_manager/ot3_state_manager/ot3_state_manager.py
DerekMaggio: eb45b5ae0f393ab82db1b12df7ba201b46b515a0
How's this?

DerekMaggio:
> Nit: Curious why we need to have the readable name be `Gantry X` etc. I feel like it's pretty straightforward to have it just be `X` axis

It is just so I can have a user-friendly name to surface instead of using `__class__.__name__`

ot3_state_manager/ot3_state_manager/ot3_state_manager.py
```diff
@@ -0,0 +1,209 @@
...
+        left_pipette_pos = (
+            self.left_pipette.position if self.left_pipette is not None else None
+        )
+        right_pipette_pos = (
+            self.right_pipette.position if self.right_pipette is not None else None
+        )
+        gripper_pos = self.gripper.position if self.gripper is not None else None
+
+        return {
+            Head.gripper_mount_position.axis: self.head.gripper_mount_position,
```
Laura-Danielle, line 112: A little confused as to what this gripper mount position is. I think when @sfoster1 and I discussed awhile back, we decided on not characterizing the gripper mount as an axis thing because it doesn't actually move. Is this meant to be the Z axis of the gripper itself?

If so, would it be possible to call this `gripper_z_position`?

Laura-Danielle (approved):
One little nit comment, but otherwise looks like a good first pass to me.

ot3_state_manager/ot3_state_manager/ot3_state_manager.py
DerekMaggio: Addressed in: 
https://github.com/Opentrons/ot3-firmware/pull/418/commits/df794f19371acc636beebcc2b974988975faf588
https://github.com/Opentrons/ot3-firmware/pull/418/commits/ed2860dd90e8aa7f14d009ec625d81147a1c404d

ot3_state_manager/Makefile
```diff
@@ -0,0 +1,39 @@
+python := poetry run python
+pytest := poetry run py.test
+SHX := npx shx
```
sfoster1, line 3: definitely do not use npx here, we can use normal shell

ot3_state_manager/ot3_state_manager/measurable_states.py
sfoster1: yeah

ot3_state_manager/pyproject.toml
```diff
@@ -0,0 +1,25 @@
+[tool.poetry]
+name = "ot3_state_manager"
+version = "0.1.0"
+description = ""
+authors = ["Derek Maggio <derek.maggio@opentrons.com>"]
```
sfoster1, line 5: we tend to make our authors `Opentrons <engineering@opentrons.com>` i believe

ot3_state_manager/pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
```
sfoster1, line 1: https://docs.pytest.org/en/6.2.x/customize.html#pyproject-toml we can fold this setting into pyproject.toml

ot3_state_manager/setup.cfg
```diff
@@ -0,0 +1,8 @@
+[metadata]
```
sfoster1, line 1: see if we can fold these into pyproject toml too

ot3_state_manager/setup.py
```diff
@@ -0,0 +1,72 @@
+# Inspired by:
```
sfoster1, line 1: setup.py should be replaceable with pyproject toml configuration: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

sfoster1 (changes requested):
The code looks great, but there's some little tuneups in the makefile and we should be able to fold what's left of setup.py into the toml file.

ot3_state_manager/Makefile
DerekMaggio: 2b32951e169f0fc8559a78f4f1188103dfad43f2

ot3_state_manager/pytest.ini
DerekMaggio: https://github.com/Opentrons/ot3-firmware/pull/418/commits/7d71df09720e545a05b7d9d7ef1b08717593ec2f

ot3_state_manager/pyproject.toml
DerekMaggio: https://github.com/Opentrons/ot3-firmware/pull/418/commits/7d71df09720e545a05b7d9d7ef1b08717593ec2f

DerekMaggio:
@sfoster1 I got pytest, mypy, and isort folded into the pyproject.toml.

I am really struggling with getting setup.py and setup.cfg to work. 

Can we either abandon this section or can I get some help?

DerekMaggio:
@sfoster1 I was able to fold setup.py into pyproject.toml
I was not able to fold in setup.cfg

Laura-Danielle (approved):
Yep this still looks good to me. It seems like you might need additional plugins right now to add the stuff in `setup.cfg` to the `toml` file. See [this repo](https://github.com/carlosperate/awesome-pyproject) and [this article](https://ianhopkinson.org.uk/2022/02/understanding-setup-py-setup-cfg-and-pyproject-toml-in-python/).

sfoster1 (approved):
Awesome! Love to see pyproject toml getting used finally

PR #419, opened 2022-07-22
fix(pipettes): spi chip select gpio mode should be push-pull

Laura-Danielle:
The motor was not engaging on the G4 boards because the chip select pin was using the wrong gpio
mode.


sfoster1 (approved):
Nice

PR #420, opened 2022-07-25
feat(gripper): encoder position feedback in Ack message

ahiuchingau:
Gripper now reports encoder position in um when a jaw-move finishes. Tested on the robot and verified the reported values are correct.

The next step is to use the ability to read the encoder position to measure the speed of the motor. We can then monitor the speed and use that info to determine whether or not the jaw has finished gripping a labware (instead of hardcoding a duration for the move to complete).


gripper/core/can_tasks.cpp
```diff
@@ -96,7 +96,14 @@ static auto dispatcher_g = can::dispatch::Dispatcher(
 
 /** Dispatcher to the various handlers */
 static auto main_dispatcher = can::dispatch::Dispatcher(
-    can::dispatch::StandardArbIdTest{.node_id = can::ids::NodeId::gripper},
+    [](uint32_t arbitration_id) -> bool {
```
Laura-Danielle, line 99: Do you need to rebase? Thought you already made a pr with this change.

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -85,8 +85,8 @@ class BrushedMotorInterruptHandler {
     }
 
     void homing_stopped() {
-        finish_current_move(AckMessageId::stopped_by_condition);
         hardware.reset_encoder_pulses();
+        finish_current_move(AckMessageId::stopped_by_condition);
```
Laura-Danielle, line 89: shouldn't this be before resetting the encoder pulses?

Laura-Danielle (approved):
Two comments, but generally lgtm

sfoster1 (approved):
Sick

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
ahiuchingau: We want to reset first because if this is the first time the motor homes after boot up, the encoder position could be an extremely large negative value.

PR #421, opened 2022-07-27
feat(pipettes): add support for hdc3020 environment sensor

Laura-Danielle:
## Overview
We are switching to a different environment sensor due to supply chain constraints. The new hdc3020 sensor is quite different than the original one, and I also took the liberty of refactoring the code to look more like the pressure sensor.

A typical reading from this sensor would look like `<humidity><humidity_crc><temperature><temperature_crc>`, regardless of whether you use the single read setting or the auto measurement setting.

## Changelog
There are some big changes in this PR such as:
- Changing the environment sensor to have it's own `sensor type`
- Adding an `environment_driver` to handle underlying functionality

## ToDos
A few things that haven't been figured out quite yet are:
- When should we change the power consumption mode?
- What to do for continuous polling

I imagine we will be working through what it means to periodically poll both this sensor and the pressure sensor so I'm pushing that work off to a future PR.


include/sensors/simulation/hdc3020.hpp
```diff
@@ -0,0 +1,43 @@
...
+using namespace i2c::hardware;
+
+class HDC3020 : public I2CRegisterMap<uint8_t, uint64_t> {
+  public:
+    HDC3020()
+        : I2CRegisterMap(
+              hdc3020::ADDRESS,
+              {{static_cast<uint8_t>(
+                    hdc3020::Registers::TRIGGER_ON_DEMAND_MODE),
+                0x209631509031},
```
sfoster1, line 18: can we build these from real values with a helper function or something

sfoster1 (approved):
Looks great, would love to see that test be made a little more legible by building the uint64s in a helper function or something

PR #423, opened 2022-07-28
chore: Add monorepo as dependency

DerekMaggio:
# Overview

Add `api` and `opentrons_hardware` projects from [monorepo](https://github.com/Opentrons/opentrons) as dependencies to OT3 State Manager

# Changelog

- Add `api` and `hardware` project to pyproject.toml. It expects the monorepo to be a sibling of `ot3-firmware`
- Remove local OT3Axis class and instead refer to class from monorepo
- Update Github Action by adding monorepo

# Review Requests

@sfoster1, I made the requirement that the monorepo be a sibiling to the `ot3-firmware` directory. I did this to match the expectation of where the monorepo is when generating the CAN Message header file. Is this the right move?

# Risk

None, this is not being used anywhere yet.


sfoster1 (approved):
Yeah this is awesome though we should get something about it in a readme

DerekMaggio:
> Yeah this is awesome though we should get something about it in a readme

That is my next PR :)

PR #424, opened 2022-07-28
docs: OT-3 State Manager README Init

DerekMaggio:
Added a README
Add README linting to make targets and CI


ahiuchingau (approved):
Coolio

PR #425, opened 2022-07-28
feat(gripper): track brushed motor encoder speed and use it to complete brushed motor move

ahiuchingau:
Configures TIM4 to be a tracker for the encoder pulse speed. 

TIM4 counts at 1 MHz and by capturing the counter value between two consecutive encoder pulse edges (4 edges make up one full encoder pulse), we can estimate the speed of the encoder/brushed motor. When the timer overflows (timer period is set to 3 ms), we can assume that the encoder is pulsing at most 8.33 Hz (1 MHz / (30000 x 4 count)). This is slow enough that we can assume the encoder has pretty much stopped. 

We can then use this info to report back to the host that the gripper jaw has stopped moving, instead of relying on an arbitrary duration.

Here's some result I got from testing:
The compare/capture (CCR1) register reading of the TIM4 was 0x2FF, which is 767 us. 

Screen Shot 2022-07-28 at 2 37 58 PM

### DIFF
--- a/gripper/firmware/motor_encoder_hardware.c
+++ b/gripper/firmware/motor_encoder_hardware.c
@@ -42,12 +42,12 @@ void Encoder_GPIO_Init(void) {
     GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
     HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
 
-    //    GPIO_InitStruct.Pin = GPIO_PIN_5;
-    //    GPIO_InitStruct.Mode = GPIO_MODE_AF_PP;
-    //    GPIO_InitStruct.Pull = GPIO_NOPULL;
-    //    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_HIGH;
-    //    GPIO_InitStruct.Alternate = GPIO_AF1_TIM2;
-    //    HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
+    GPIO_InitStruct.Pin = GPIO_PIN_5;
+    GPIO_InitStruct.Mode = GPIO_MODE_AF_PP;
+    GPIO_InitStruct.Pull = GPIO_NOPULL;
+    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_HIGH;
+    GPIO_InitStruct.Alternate = GPIO_AF1_TIM2;
+    HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
 }
 
 void TIM2_EncoderG_Init(void) {
--- a/gripper/firmware/motor_encoder_hardware.c
+++ b/gripper/firmware/motor_encoder_hardware.c
@@ -3,7 +3,20 @@
 #include "common/firmware/errors.h"
 #include "stm32g4xx_hal.h"
 
+#define GRIPPER_ENCODER_SPEED_TIMER_FREQ (1000000UL)
+
 TIM_HandleTypeDef htim2;
+TIM_HandleTypeDef htim4;
+
+uint32_t round_closest(uint32_t dividend, uint32_t divisor) {
+    return (dividend + (divisor / 2)) / divisor;
+}
+
+uint32_t calc_prescaler(uint32_t timer_clk_freq, uint32_t counter_clk_freq) {
+    return timer_clk_freq >= counter_clk_freq
+               ? round_closest(timer_clk_freq, counter_clk_freq) - 1U
+               : 0U;
+}
 
 void Encoder_GPIO_Init(void) {
     /* Peripheral clock enable */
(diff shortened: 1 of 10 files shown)


include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -115,9 +116,9 @@ class BrushedMotorInterruptHandler {
     }
 
     bool has_active_move = false;
+    bool is_idle = true;
```
sfoster1, line 119: probably wants to be `atomic_bool` since this is coming from another isr

sfoster1 (changes requested):
tests for the motor interrupt handler! but this looks awesome

gripper/firmware/motor_encoder_hardware.c
```diff
@@ -70,6 +83,67 @@ void TIM2_EncoderG_Init(void) {
...
+ * which the encoder is rotating, hence, the speed at which the gripper jaw is
+ * travelling.
+ *
+ * TIM4 receives the encoder clock output signal from TIM2 (tim_trgo). It counts
+ * at 1 MHz and captures the current count whenever an encoder pulse is
+ * received, then resets its counter. The velocity would thus roughly be
+ * (1 MHz / (count * 4)). Counting both rising and falling edges from channel A
+ * + B.
+ *
+ * The timer overflows 3 ms after an edge is received. This means any encoder
```
Laura-Danielle, line 98: How did you decide on 3ms?

ahiuchingau: Just made it up tbh! Probs should align with the testing team to make sure this behaves okay, depending on what we're gripping.

gripper/firmware/motor_encoder_hardware.c
```diff
@@ -70,6 +83,73 @@ void TIM2_EncoderG_Init(void) {
...
```
sfoster1, line 98: can we add an estimate of what this means in terms of linear velocity for the current motion config

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -55,28 +57,30 @@ class BrushedMotorInterruptHandler {
                 limit_switch_triggered()) {
                 homing_stopped();
             } else {
-                tick_count++;
-                if (!should_continue()) {
+                if (is_idle) {
```
sfoster1, line 60: We may need a holdoff here, a minimum time before we start checking `is_idle` so we give the motor time to accelerate and actually start moving.

sfoster1 (changes requested):
I think we should avoid altering the `is_idle` state in the interrupt itself, keep it just being an always-current thing that monitors the physical state of the system. Then on top of that we can layer some holdoff time that enforces a minimum of say a millisecond or two before we accept `is_idle`.

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -55,28 +57,30 @@ class BrushedMotorInterruptHandler {
...
-        /*
-         * TODO: Check encoder position as the condition to stop motor instead
-         * of duration.
+    void enc_speed_timer_overflows() {
+        /* When the encoder speed timer overflows, this means the encoder is
+         * moving at a frequency lower than 8 Hz, we can assume the gripper jaw
+         * has stopped moving and the active move is complete.
          */
-        return tick_count < buffered_move.duration;
+        if (has_active_move) {
```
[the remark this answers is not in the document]
ahiuchingau: True, I'll remove this

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -87,6 +91,7 @@ class BrushedMotorInterruptHandler {
     void homing_stopped() {
         hardware.reset_encoder_pulses();
         finish_current_move(AckMessageId::stopped_by_condition);
+        is_idle = true;
```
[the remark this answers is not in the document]
ahiuchingau: Probably not, I was afraid that it would trip the timer interrupt to report a finish move without meeting the condition (if for whatever reason the encoder speed timer overflows right when homing_stopped happens), but i can test this to verify

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -55,28 +57,30 @@ class BrushedMotorInterruptHandler {
...
+            is_idle = true;
+        }
     }
 
     void update_and_start_move() {
         has_active_move = queue.try_read_isr(&buffered_move);
         if (buffered_move.duty_cycle != 0U) {
             driver_hardware.update_pwm_settings(buffered_move.duty_cycle);
         }
+        is_idle = false;
```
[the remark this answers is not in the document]
ahiuchingau: Currently nothing is setting this value to false, maybe I can make the input capture callback to set it so it doesn't get handled by the motor interrupt handler.

gripper/firmware/motor_encoder_hardware.c
```diff
@@ -22,12 +35,19 @@ void Encoder_GPIO_Init(void) {
...
     GPIO_InitStruct.Pin = GPIO_PIN_5;
-    GPIO_InitStruct.Mode = GPIO_MODE_AF_PP;
+    GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
     GPIO_InitStruct.Pull = GPIO_NOPULL;
-    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_HIGH;
-    GPIO_InitStruct.Alternate = GPIO_AF1_TIM2;
+    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
     HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
+
+    //    GPIO_InitStruct.Pin = GPIO_PIN_5;
```
sfoster1, line 45: remove commented out code

sfoster1 (approved):
This looks great! Just remove the commented out code before merging please

PR #426, opened 2022-07-29
feat(eeprom):  implement eeprom filesystem

ryanthecoder:
This PR is paired with this PR https://github.com/Opentrons/opentrons/pull/11260

### DIFF
--- a/include/eeprom/core/accessor.hpp
+++ b/include/eeprom/core/accessor.hpp
@@ -1,8 +1,6 @@
 #pragma once
 
 #include <array>
-#include <concepts>
-#include <cstdint>
 #include <vector>
 
 #include "addresses.hpp"
--- a/include/eeprom/core/accessor.hpp
+++ b/include/eeprom/core/accessor.hpp
@@ -47,15 +47,24 @@ class ReadListener {
     virtual void read_complete() = 0;
 };
 
+/**
+ * Base class for all of the other accessors
+ * This class should not be directly used.
+ *
+ * @tparam EEPromTaskClient client of eeprom task
+ * @tparam data_begin Adddress from addresses.hpp where this data section begins
+ **/
 template <task::TaskClient EEPromTaskClient, types::address data_begin>
 class EEPromAccessor {
-  public:
+  protected:
     explicit EEPromAccessor(EEPromTaskClient& eeprom_client,
                             ReadListener& listener, AccessorBuffer buff)
         : type_data(buff),
           eeprom_client{eeprom_client},
           read_listener(listener),
           begin{data_begin} {}
+
+  public:
     EEPromAccessor(const EEPromAccessor&) = delete;
     EEPromAccessor(EEPromAccessor&&) = delete;
     auto operator=(EEPromAccessor&&) -> EEPromAccessor& = delete;
--- a/include/eeprom/core/accessor.hpp
+++ b/include/eeprom/core/accessor.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+
 #include <array>
 #include <vector>
 
@@ -13,7 +14,6 @@ namespace accessor {
 /**
  * Interface of an object that listens for reads
  */
-template <typename T>
 class ReadListener {
   public:
     ReadListener() = default;
@@ -23,7 +23,7 @@ class ReadListener {
     auto operator=(ReadListener&&) noexcept -> ReadListener& = default;
 
     virtual ~ReadListener() = default;
-    virtual void on_read(const T&) = 0;
+    virtual void read_complete() = 0;
 };
 
 template <task::TaskClient EEPromTaskClient, typename T,
@@ -55,7 +55,7 @@ class EEPromAccessor {
   protected:
     T type_data = T(default_size);
     EEPromTaskClient& eeprom_client;
-    ReadListener<T>& read_listener;
+    ReadListener& read_listener;
     types::address begin;
 
     /**
@@ -127,8 +127,7 @@ class EEPromAccessor {
         std::copy_n(msg.data.cbegin(), msg.length, buffer_ptr);
         bytes_recieved += msg.length;
         if (bytes_recieved == bytes_to_read) {
-            read_listener.on_read(
-                T(type_data.begin(), type_data.begin() + bytes_recieved));
+            read_listener.read_complete();
         }
     }
 
--- a/include/eeprom/core/accessor.hpp
+++ b/include/eeprom/core/accessor.hpp
@@ -1,0 +1,156 @@
+#pragma once
+
+#include <array>
+#include <concepts>
+#include <cstdint>
+#include <vector>
+
+#include "addresses.hpp"
+#include "task.hpp"
+#include "types.hpp"
+
+namespace eeprom {
+namespace accessor {
+
+/**
+ * Interface of an object that listens for reads
+ */
+template <typename T>
+class ReadListener {
+  public:
+    ReadListener() = default;
+    ReadListener(const ReadListener&) = default;
+    auto operator=(const ReadListener&) -> ReadListener& = default;
+    ReadListener(ReadListener&&) noexcept = default;
+    auto operator=(ReadListener&&) noexcept -> ReadListener& = default;
+
+    virtual ~ReadListener() = default;
+    virtual void on_read(const T&) = 0;
+};
+
+template <task::TaskClient EEPromTaskClient, typename T,
+          types::address data_begin, types::address default_size>
+class EEPromAccessor {
+  public:
+    explicit EEPromAccessor(EEPromTaskClient& eeprom_client,
+                            ReadListener<T>& listener)
+        : eeprom_client{eeprom_client},
+          read_listener{listener},
+          begin{data_begin} {}
+    EEPromAccessor(const EEPromAccessor&) = delete;
+    EEPromAccessor(EEPromAccessor&&) = delete;
+    auto operator=(EEPromAccessor&&) -> EEPromAccessor& = delete;
+    auto operator=(const EEPromAccessor&) -> EEPromAccessor& = delete;
+    virtual ~EEPromAccessor() = default;
+    /**
+     * Begin a read of the data
+     */
+    auto start_read() -> void { start_read_at_offset(0, default_size); }
+
+    /**
+     * Write data to eeprom
+     */
+    auto write(const T& type_value) -> void {
+        write_at_offset(type_value, 0, default_size);
+    }
+
+  protected:
+    T type_data = T(default_size);
+    EEPromTaskClient& eeprom_client;
+    ReadListener<T>& read_listener;
+    types::address begin;
+
+    /**
+     * Write data to eeprom at a specified offset
+     */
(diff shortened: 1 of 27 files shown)


include/eeprom/core/accessor.hpp
```diff
@@ -0,0 +1,154 @@
...
+    ReadListener(const ReadListener&) = default;
+    auto operator=(const ReadListener&) -> ReadListener& = default;
+    ReadListener(ReadListener&&) noexcept = default;
+    auto operator=(ReadListener&&) noexcept -> ReadListener& = default;
+
+    virtual ~ReadListener() = default;
+    virtual void on_read(const T&) = 0;
+};
+
+template <task::TaskClient EEPromTaskClient, typename T,
```
sfoster1, line 29: what's `T` here? can we restrict it more tightly with a concept or a comment or something

include/eeprom/core/accessor.hpp
```diff
@@ -0,0 +1,154 @@
...
+  protected:
+    T type_data = T(default_size);
+    EEPromTaskClient& eeprom_client;
+    ReadListener<T>& read_listener;
+    types::address begin;
+
+    /**
+     * Write data to eeprom at a specified offset
+     */
+    auto write_at_offset(const std::vector<uint8_t>& data,
```
sfoster1, line 64: factoring this into taking an iterator pair or a range means this would not need a direct reference to `vector` anymore

include/eeprom/core/data_rev.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include "accessor.hpp"
+#include "addresses.hpp"
+#include "common/core/bit_utils.hpp"
+#include "task.hpp"
+
+namespace eeprom {
+namespace data_revision {
+
+using DataRevisionType = std::vector<uint8_t>;
```
sfoster1, line 11: so we definitely can't use vectors that call `malloc` since we want to avoid that in a microcontroller context. we do have a memory pool allocator implementation, but those work best for, well, memory pool contexts - situations where we have a bunch of objects of the same size and repeatedly want to fill them. 

i think that the needs of the interface can be better fulfilled by using iterator pairs or standard ranges. callers can then create arrays of the right size for the specific data they're trying to retrieve or write since they know what the data is, while the interface itself uses the iterpairs/ranges to preserve generality.

include/eeprom/core/dev_data.hpp
```diff
@@ -0,0 +1,388 @@
...
+                                   addresses::lookup_table_tail_begin,
+                                   addresses::lookup_table_tail_length>::
+              EEPromAccessor(eeprom_client, listener) {}
+
+  public:
+    auto increase_data_tail(const DataTailType& data_added) -> void {
+        types::data_length amount_to_read;
+        auto read_addr = addresses::lookup_table_tail_begin;
+        auto bytes_remain = addresses::lookup_table_tail_length;
+        const auto cb_param = new increase_data_cb_param{this, data_added};
```
sfoster1, line 53: similar to vectors, we definitely should not be using `new` - for a variety of reasons, really. for predictability, in a microcontroller context we want to avoid malloc; and in general from a robustness perspective, we want to use smart pointers. here i think we're best served by packing these structs in a variant and putting them in the message by value.

include/eeprom/core/dev_data.hpp
```diff
@@ -0,0 +1,388 @@
...
+                .memory_address = read_addr,
+                .length = amount_to_read,
+                .callback = increase_tail_callback,
+                .callback_param = cb_param});
+    }
+
+    auto increase_data_tail(const types::data_length data_added) -> void {
+        DataTailType typed_length =
+            DataTailType(addresses::lookup_table_tail_length);
+        std::ignore = bit_utils::int_to_bytes(
```
sfoster1, line 66: i did not know about std::ignore, that's cool!

include/eeprom/core/messages.hpp
```diff
@@ -29,5 +29,18 @@ struct ReadEepromMessage {
 /** The write to eeprom message */
 using WriteEepromMessage = EepromMessage;
 
+struct ConfigResponseMessage {
+    eeprom::hardware_iface::EEPromChipType chip;
+    types::address addr_bytes;
+    types::data_length mem_size;
+};
+
+using ConifgRequestCallback = void (*)(const ConfigResponseMessage&, void*);
```
sfoster1, line 38:
```suggestion
using ConfigRequestCallback = void (*)(const ConfigResponseMessage&, void*);
```
also maybe a `std::function`?

include/eeprom/core/messages.hpp
```diff
@@ -29,5 +29,18 @@ struct ReadEepromMessage {
...
+
+struct ConfigRequestMessage {
+    ConifgRequestCallback callback;
```
sfoster1, line 41:
```suggestion
    ConfigRequestCallback callback;
```

sfoster1 (changes requested):
Overall it looks good but we need to change the memory management and data structure strategy.

We cannot use `malloc` in microcontroller contexts. There's a couple uses of `new` we need to remove, but the larger issue is `vector`. We have a memory pool allocator implementation that we could use to back it up, but I'm not sure how useful that will be since these are mostly one-off uses. I think you've used `vector` to make this tolerant of filesystem elements of varying size, but given that I think there's a better way to do this: iterator pairs in the interface and callers with the context to know the appropriate size.

- The interface to the file system takes iterator pairs or `std::range`s - the iterator pairs thing is a little more verbose but doesn't need (quite) as many templates
- Callers use `arrays` or similar given that they know what they're requesting

This should provide the level of generality we want in something like a filesystem.

ryanthecoder:
@sfoster1 changed the implementation to the way you suggested.

include/eeprom/core/accessor.hpp
```diff
@@ -0,0 +1,183 @@
+#pragma once
+
+#include <array>
+#include <cassert>
+#include <vector>
```
sfoster1, line 5: don't need this anymore right

include/eeprom/core/accessor.hpp
```diff
@@ -0,0 +1,183 @@
...
+#include "types.hpp"
+
+namespace eeprom {
+namespace accessor {
+
+class AccessorBuffer {
+  public:
+    AccessorBuffer() = default;
+    template <typename B_ITER, typename L_ITER>
+    requires bit_utils::ByteIterator<B_ITER> &&
```
sfoster1, line 19: kind of a nit but we should require https://en.cppreference.com/w/cpp/iterator/contiguous_iterator if we're going to alias this with a pair of raw pointers

include/eeprom/core/accessor.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+#include "addresses.hpp"
+#include "common/core/bit_utils.hpp"
+#include "task.hpp"
+#include "types.hpp"
+
+namespace eeprom {
+namespace accessor {
+
+class AccessorBuffer {
```
sfoster1, line 15: i'm kind of concerned with lifetimes here. `AccessorBuffer` only holds pointers to the data you pass in; is it possible that an `AccessorBuffer` instance can outlive the data it points to? can we ensure that doesn't happen somehow?

include/eeprom/core/messages.hpp
```diff
@@ -15,6 +15,9 @@ struct EepromMessage {
 };
 
 using ReadResponseCallback = void (*)(const EepromMessage&, void*);
+// would love to use this instead if gcc wasn't broken on my system. revist when
```
sfoster1, line 18: can you drop a link to the gcc bug here

include/eeprom/core/messages.hpp
```diff
@@ -29,5 +32,21 @@ struct ReadEepromMessage {
...
 
+struct ConfigResponseMessage {
+    eeprom::hardware_iface::EEPromChipType chip;
+    types::address addr_bytes;
+    types::data_length mem_size;
+};
+
+using ConfigRequestCallback = void (*)(const ConfigResponseMessage&, void*);
+// would love to use this instead if gcc wasn't broken on my system. revist when
+// gcc > 12.1.1 using ConfigRequestCallback = std::function<void(const
```
sfoster1, line 43: ditto bug

sfoster1 (approved):
A couple nits but overall looks great!

include/eeprom/core/accessor.hpp
ryanthecoder: I looked and it does not seem that there is a good way to detect that a pointer points to valid memory this unless you attempt to catch SYSSEGV which is discouraged because its supposed to be an unrecoverable error.

include/eeprom/core/messages.hpp
ryanthecoder: 👍

include/eeprom/core/messages.hpp
ryanthecoder: 👍

Laura-Danielle (commented):
Have some random comments/questions. In general, it might be nice to have a centralized place of all the different types of things that can be stored in the file system (though that definitely shouldn't happen in this PR). It doesn't have to be implemented exactly like the sensor drivers or motor control drivers (see `tmc2130.hpp` + `tmc2130_driver.hpp` for a reference), but maybe having each data type defined in structs like those files might be cool.

Do we plan to store each data type at specific addresses, or are we reading/writing things to the eeprom as the different tasks receive commands?

include/eeprom/core/dev_data.hpp
```diff
@@ -0,0 +1,384 @@
...
+                .callback = table_action_callback,
+                .callback_param = this});
+        }
+    }
+    void get_data(uint16_t key, uint16_t len) { get_data(key, len, 0); }
+    void get_data(uint16_t key) { get_data(key, 0, 0); }
+
+    template <std::size_t SIZE>
+    void create_data_part(uint16_t key, uint16_t len,
+                          std::array<uint8_t, SIZE>& data) {
```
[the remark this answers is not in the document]
ryanthecoder: So the data being written to the eeprom will always need to be passed in as byte arrays since this is the way the underlying eeprom reads/writes data. Implementing clients will need to serialize/parse anything that they want to read/write.
This accessor aims to be as generic as possible. 

Also the data being passed in here is not of DataBufferType because the data being written can come from any type of byte data the client wants to send and it doesn't necessarily need to match the size of the DataBufferType used as the backing storage.

include/eeprom/core/dev_data.hpp
```diff
@@ -0,0 +1,384 @@
...
+                        .callback = table_action_callback,
+                        .callback_param = this});
+            }
+        } else {
+            LOG("ERROR, attempting to create data part before driver "
+                "initalized");
+        }
+    }
+
+    void create_data_part(uint16_t key, uint16_t len) {
```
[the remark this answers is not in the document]
ryanthecoder: figured it may be helpful in initial flashing or setup, for example to set up and reserve usage data fields before any usage occurs.

sfoster1 (approved):
All right! Looks great, nice work

PR #427, opened 2022-08-01
feat(motor-control): add stealth chop config for TMC2130

ahiuchingau:
Enables passive motor braking using StealthChop. 

Passive braking is enabled by setting the standstill current IHOLD to zero and choosing the desired option using the FREEWHEEL setting (%10 - Coil shorted using LS drivers).

### DIFF
--- a/include/motor-control/core/stepper_motor/tmc2130.hpp
+++ b/include/motor-control/core/stepper_motor/tmc2130.hpp
@@ -456,8 +456,7 @@ struct __attribute__((packed, __may_alias__)) StealthChop {
 
     uint32_t pwm_ampl : 8 = 0;
     uint32_t pwm_grad : 8 = 0;
-    uint32_t pwm_freq0 : 1 = 0;
-    uint32_t pwm_freq1 : 1 = 0;
+    uint32_t pwm_freq : 2 = 0;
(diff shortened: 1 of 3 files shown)


include/motor-control/core/stepper_motor/tmc2130.hpp
```diff
@@ -446,6 +446,31 @@ struct __attribute__((packed, __may_alias__)) DriveStatus {
...
+ * This register sets the control current for voltage PWM mode stealth chop.
+ */
+struct __attribute__((packed, __may_alias__)) StealthChop {
+    static constexpr Registers address = Registers::PWMCONF;
+    static constexpr bool writable = true;
+    static constexpr uint32_t value_mask = (1 << 22) - 1;
+
+    uint32_t pwm_ampl : 8 = 0;
+    uint32_t pwm_grad : 8 = 0;
+    uint32_t pwm_freq0 : 1 = 0;
```
fsinapi, line 459: `pwm_freq` and `freewheel` could each be consolidated into 2-bit bitfields, which is how the rest of the registers are defined - I think it makes it a little clearer to set/read registers that way.

sfoster1 (approved):
That's a cool feature! Agree with @fsinapi about the the bitfield thing

PR #429, opened 2022-08-02
feat: OT-3 state manager communication (RET-1104)

DerekMaggio:
# Overview

Added asyncio socket communication layer to OT-3 State Manager
See requirements doc [here](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3322773722/OT3+Emulation+State+Manager#Phase-2%3A-Connect-to-Emulation-(In-Progress))

# Changelog

- Rename `OT3StateManager` to `OT3State` and have it be the only thing that handles update to the overall state view of OT-3
- Create `OT3StateManager` which is responsible for communicating over socket and parsing received messages into update commands for the `OT3State` class to make.
- Add SyncPin` class which represents the OT-3 Sync Pin
- Refactor all classes in `hardware.py` to normal Python classes instead of dataclasses

# Review Requests

- What do you think about the format of the messages? 
- In `hardware.py` I had to refactor all the dataclasses to normal python classes because default values are stored as Class Variables for dataclasses. This led to state being shared across multiple instances of a hardware class. Any suggestions for how to remedy this. See [stdlib docs](https://docs.python.org/3/library/dataclasses.html#mutable-default-values)
- The `_handle_message` method inside of OT3StateManager is way too complex and confusing. Any suggestions for parsing this in a cleaner way? 
- Any suggestions for reporting errors back across the socket communication? Currently it feels very hacky and flakey.

### DIFF
--- a/ot3_state_manager/ot3_state_manager/measurable_states.py
+++ b/ot3_state_manager/ot3_state_manager/measurable_states.py
@@ -9,9 +9,9 @@ class Position:
     """Represents position state."""
 
     axis: OT3Axis
-    current_position: float = 0.0
-    commanded_position: float = 0.0
-    encoder_position: float = 0.0
+    current_position: int = 0
+    commanded_position: int = 0
+    encoder_position: int = 0
 
 
 @dataclass
@@ -19,6 +19,6 @@ class Force:
     """Represents force state."""
 
     axis: OT3Axis
-    current_force: float = 0.0
-    commanded_force: float = 0.0
-    encoder_force: float = 0.0
+    current_force: int = 0
+    commanded_force: int = 0
+    encoder_force: int = 0
--- a/ot3_state_manager/ot3_state_manager/ot3_state.py
+++ b/ot3_state_manager/ot3_state_manager/ot3_state.py
@@ -34,8 +34,8 @@ class OT3State:
     @classmethod
     def build(
         cls,
-        left_pipette_model: Optional[PipetteModel],
-        right_pipette_model: Optional[PipetteModel],
+        left_pipette_model_name: Optional[str],
+        right_pipette_model_name: Optional[str],
         use_gripper: bool,
     ) -> OT3State:
         """Helper function to build an OT3State object.
@@ -51,13 +51,13 @@ class OT3State:
         slot.
         """
         left_pipette = (
-            LeftPipette(model=left_pipette_model)
-            if left_pipette_model is not None
+            LeftPipette(PipetteModel.model_name_to_enum(left_pipette_model_name))
+            if left_pipette_model_name is not None
             else None
         )
         right_pipette = (
-            RightPipette(model=right_pipette_model)
-            if right_pipette_model is not None
+            RightPipette(PipetteModel.model_name_to_enum(right_pipette_model_name))
+            if right_pipette_model_name is not None
             else None
         )
         gripper = Gripper() if use_gripper else None
--- a/ot3_state_manager/ot3_state_manager/ot3_state.py
+++ b/ot3_state_manager/ot3_state_manager/ot3_state.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import logging
 from typing import Dict, List, Optional
 
-from ot3_state_manager.util import Direction, Message
 from opentrons.hardware_control.types import OT3Axis
 
 from ot3_state_manager.hardware import (
@@ -19,9 +18,10 @@ from ot3_state_manager.hardware import (
 )
 from ot3_state_manager.measurable_states import Position
 from ot3_state_manager.pipette_model import PipetteModel
-
+from ot3_state_manager.util import Direction
 
 log = logging.getLogger(__name__)
+
 
 class OT3State:
     """Class to maintain state of OT3.
@@ -63,9 +63,7 @@ class OT3State:
         gripper = Gripper() if use_gripper else None
 
         return cls(
-            left_pipette=left_pipette,
-            right_pipette=right_pipette,
-            gripper=gripper
+            left_pipette=left_pipette, right_pipette=right_pipette, gripper=gripper
         )
 
     def __init__(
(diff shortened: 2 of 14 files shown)


ot3_state_manager/ot3_state_manager/ot3_state_manager.py
```diff
@@ -1,213 +1,118 @@
...
-        return self.right_pipette is not None
-
-    @property
-    def is_high_throughput_pipette_attached(self) -> bool:
-        """Returns True if high throughput pipette is attached, False if not."""
-        return (
-            self.right_pipette is not None
-            and self.right_pipette.model == PipetteModel.MULTI_96_1000
+    async def _handle_message(
+        self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter
```
sfoster1, line 38: I think we don't want to use tcp and streams here but rather udp and datagrams: https://docs.python.org/3/library/asyncio-protocol.html#udp-echo-client

sfoster1: we're sending discrete messages after all, and we'd like to be able to loadshed if we need to, so udp is a much better fit

ot3_state_manager/pyproject.toml
```diff
@@ -23,13 +24,15 @@ isort = "5.10.1"
 mdformat-gfm = "^0.3.5"
 mdformat_frontmatter = "^0.4.1"
 mdformat_footnote = "^0.1.1"
+pydantic = "1.8.2"
```
sfoster1, line 27: what are we using pydantic for? it doesn't seem to be used anywhere

sfoster1 (changes requested):
Overall looking pretty good but we should really use udp streams for a message oriented protocol

ot3_state_manager/pyproject.toml
DerekMaggio: Thank you. I was going to use it for the message format but changed my mind. I will remove it

ot3_state_manager/ot3_state_manager/util.py
```diff
@@ -0,0 +1,108 @@
...
+                if sync_pin_state == "HIGH":
+                    return cls(state="HIGH")
+                elif sync_pin_state == "LOW":
+                    return cls(state="LOW")
+
+        return None
+
+
+@dataclass
+class MoveMessage(Message):
```
Laura-Danielle, line 64: technically the sync pin would also have its state change as a result of a move -- should we incorporate that into a move message?

ot3_state_manager/ot3_state_manager/ot3_state.py
```diff
@@ -0,0 +1,239 @@
...
+
+        if pos_to_update is None:
+            raise ValueError(
+                f'Axis "{axis_to_update.name}" is not available. '
+                f"Cannot update it's position."
+            )
+        pos_to_update.current_position = current_position
+        pos_to_update.encoder_position = encoder_position
+
+    def pulse(self, axis: OT3Axis, direction: Direction) -> None:
```
Laura-Danielle, line 213: Question for both @sfoster1 and @DerekMaggio. I know we want this state manager to sit somewhere below the hardware controller. Do we want it and/or need the simulator to send status commands over CAN, or should we make this state manager a bit more low level and have it be a task in the simulator which can provide state to whoever is interested.

I know this would be very different that what we're currently doing, but a lot of this feels like duplication of logic to me so far.

Laura-Danielle (commented):
Have a few comments -- and general architecture question.

ot3_state_manager/ot3_state_manager/measurable_states.py
```diff
@@ -9,16 +9,16 @@ class Position:
...
 
 @dataclass
 class Force:
     """Represents force state."""
 
     axis: OT3Axis
-    current_force: float = 0.0
-    commanded_force: float = 0.0
-    encoder_force: float = 0.0
+    current_force: int = 0
```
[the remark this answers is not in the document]
DerekMaggio: What is the benefit of using `IntFieldBase` over just a normal int?

ot3_state_manager/ot3_state_manager/ot3_state_manager.py
DerekMaggio: https://github.com/Opentrons/ot3-firmware/pull/429/commits/ef93b42b7f6b7e14ff0d6f997de22f4e894695ae

Here is my first pass at implementing this.

ot3_state_manager/ot3_state_manager/util.py
DerekMaggio: I can do that. How does it work?

ot3_state_manager/ot3_state_manager/ot3_state.py
sfoster1: IMO the simulator should not send stuff over the emulated canbus network because that will make it share resources with the system under test. It should communicate via some side channel.

sfoster1: On all its interfaces - the simulators shouldn't use CAN to speak with the system manager, the system manager shouldn't use CAN to report back to anything else. It needs to be something unused by the rest of the system.

sfoster1 (approved):
This overall looks good but IMO let's get ahead of the game here and use a binary message format. Dealing with strings in c++ is awful, parsing strings is always fraught with pain especially if you're dealing with numerics, i hate magic strings so much, etc etc.

We can merge this as is for the structure but I would really prefer to move forward in the long term with binary messages - just writing `int`s raw to the streams.

DerekMaggio:
> This overall looks good but IMO let's get ahead of the game here and use a binary message format. Dealing with strings in c++ is awful, parsing strings is always fraught with pain especially if you're dealing with numerics, i hate magic strings so much, etc etc.
> 
> We can merge this as is for the structure but I would really prefer to move forward in the long term with binary messages - just writing `int`s raw to the streams.

Are you talking like some kind of enum, like as follows? 

```python
class OT3Axis(enum.Enum):
    X_INCREMENT = 1000
    Y_INCREMENT = 1001
    Z_L_INCREMENT = 1002
    Z_R_INCREMENT = 1003
    Z_G_INCREMENT = 1004
    P_L_INCREMENT = 1005
    P_R_INCREMENT = 1006
    Q_INCREMENT = 1007
    G_INCREMENT = 1008

    X_DECREMENT = 2000
    Y_DECREMENT = 2001
    Z_L_DECREMENT = 2002
    Z_R_DECREMENT = 2003
    Z_G_DECREMENT = 2004
    P_L_DECREMENT = 2005
    P_R_DECREMENT = 2006
    Q_DECREMENT = 2007
    G_DECREMENT = 2008
```

Then just send the binary serialized equivalent of their numeric values?

PR #430, opened 2022-08-05
refactor(ot3_state_manager): Rework messages

DerekMaggio:
# Overview

Each message is now mapped to an int value instead using arbitrary string parsing

[Related Docs](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3322773722/OT3+Emulation+State+Manager#Phase-2%3A-Connect-to-Emulation-(In-Progress))

# Changelog

- Remove `to_message` method from all `Message` classes
- Changed format of messages. All message are now 4 bytes long and have the following format: `<message-id><message-content>` where the length of `message-id` is a single byte and the length of `message-content` is 3 bytes.
- Replace `to_message` method and `parse_message` function with `Message` ABC.
  - Each concrete concrete class of Message is required to implement a `build` and `to_bytes` message
- The `_parse_message` method looks at the `message-id` byte and based off of the value passes `message-content` to the `build` method of the appropriate concrete implementation of `Message`
- Remove hashing of message value as response. Instead just return bytes value.
- Update tests accordingly

# Review Requests

None

# Risk

None

### DIFF
--- a/ot3_state_manager/ot3_state_manager/messages.py
+++ b/ot3_state_manager/ot3_state_manager/messages.py
@@ -32,19 +32,6 @@ class Message(ABC):
     """
 
     @staticmethod
-    def parse(message_bytes: bytes) -> Message:
-        """Parse sequence of bytes into a Message object."""
-        if not len(message_bytes) == MESSAGE_BYTE_LENGTH:
-            raise ValueError(
-                f"Message length must be {MESSAGE_BYTE_LENGTH} bytes. "
-                f"Your message was {len(message_bytes)} bytes."
-            )
-        message_id, message_content = struct.unpack(STRUCT_FORMAT_STRING, message_bytes)
-        return MessageID.from_id(message_id).builder_func(
-            message_content=message_content
-        )
-
-    @staticmethod
     @abstractmethod
     def build_message(message_content: bytes) -> Message:
         """Parse message_content into a Message object."""
@@ -140,11 +127,22 @@ class MessageID(Enum):
             raise ValueError(f"Could not find MessageID with message_id: {enum_id}.")
 
 
+def _parse_message(message_bytes: bytes) -> Message:
+    """Parse sequence of bytes into a Message object."""
+    if not len(message_bytes) == MESSAGE_BYTE_LENGTH:
+        raise ValueError(
+            f"Message length must be {MESSAGE_BYTE_LENGTH} bytes. "
+            f"Your message was {len(message_bytes)} bytes."
+        )
+    message_id, message_content = struct.unpack(STRUCT_FORMAT_STRING, message_bytes)
+    return MessageID.from_id(message_id).builder_func(message_content=message_content)
+
+
 def handle_message(data: bytes, ot3_state: OT3State) -> bytes:
-    """Parse incoming message, react to it accordingly, and respond."""
+    """Function to handle incoming message, react to it accordingly, and respond."""
     error_response: Optional[str] = None
     try:
-        message = Message.parse(data)
+        message = _parse_message(data)
     except ValueError as err:
         error_response = f"{err.args[0]}"
     except:  # noqa: E722
--- a/ot3_state_manager/ot3_state_manager/messages.py
+++ b/ot3_state_manager/ot3_state_manager/messages.py
@@ -8,8 +8,8 @@ from __future__ import annotations
 import struct
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from enum import Enum
-from typing import Any, Callable, Optional, Type
+from enum import Enum, unique
+from typing import Any, Optional, Type
 
 from opentrons.hardware_control.types import OT3Axis
 
@@ -40,10 +40,9 @@ class Message(ABC):
                 f"Your message was {len(message_bytes)} bytes."
             )
         message_id, message_content = struct.unpack(STRUCT_FORMAT_STRING, message_bytes)
-        # mypy thinks that I am instantiating a MoveMessageHardware object here.
-        # I am not, I am performing a lookup as defined in the enum docs.
-        # https://docs.python.org/3/library/enum.html#programmatic-access-to-enumeration-members-and-their-attributes
-        return MessageID(message_id).builder_func(message_content=message_content)  # type: ignore[call-arg]
+        return MessageID.from_id(message_id).builder_func(
+            message_content=message_content
+        )
 
     @staticmethod
     @abstractmethod
@@ -78,10 +77,7 @@ class MoveMessage(Message):
     def build_message(message_content: bytes) -> MoveMessage:
         """Convert message_content into a MoveMessage object."""
         hw_id, direction_val = struct.unpack(">HB", message_content)
-        # mypy thinks that I am instantiating a MoveMessageHardware object here.
-        # I am not, I am performing a lookup as defined in the enum docs.
-        # https://docs.python.org/3/library/enum.html#programmatic-access-to-enumeration-members-and-their-attributes
-        axis = MoveMessageHardware(hw_id).axis  # type: ignore[call-arg]
+        axis = MoveMessageHardware.from_id(hw_id).axis
         direction = Direction.from_int(direction_val)
         return MoveMessage(axis=axis, direction=direction)
 
@@ -122,20 +118,9 @@ class SyncPinMessage(Message):
         return struct.pack(">BHB", MessageID.SYNC_PIN.message_id, 0, self.state)
 
 
+@unique
 class MessageID(Enum):
     """Enum class defining the relationship between the message_id byte and the corresponding Message object."""
-
-    def __new__(cls, message_id: int, builder_func: Callable) -> MessageID:
-        """Create a new MessageID object.
-
-        Lookup value will be by message_id. So MessageID(<message_id>) can be used to
-        get a move message based off of the message_id.
-        """
-        obj = object.__new__(cls)
-        obj._value_ = message_id
-        obj.message_id = message_id
-        obj.builder_func = builder_func
-        return obj
 
     def __init__(self, message_id: int, message_class: Type[Message]) -> None:
         """Create MessageID object."""
@@ -144,6 +129,15 @@ class MessageID(Enum):
 
     MOVE = 0x00, MoveMessage
     SYNC_PIN = 0x01, SyncPinMessage
+
+    @classmethod
+    def from_id(cls, enum_id: int) -> MessageID:
+        """Get MessageID object by message_id."""
+        for val in cls:
+            if val.message_id == enum_id:
+                return val
+        else:
+            raise ValueError(f"Could not find MessageID with message_id: {enum_id}.")
 
 
 def handle_message(data: bytes, ot3_state: OT3State) -> bytes:
--- a/ot3_state_manager/ot3_state_manager/messages.py
+++ b/ot3_state_manager/ot3_state_manager/messages.py
@@ -22,9 +22,6 @@ MESSAGE_BYTE_LENGTH = MESSAGE_ID_BYTE_LENGTH + MESSAGE_CONTENT_BYTE_LENGTH
 STRUCT_FORMAT_STRING = f"B{MESSAGE_CONTENT_BYTE_LENGTH}s"
 
 
-# mypy doesn't like that I am instantiating a dataclass based on an abstract class (ABC)
-# instead of a concrete class.
-@dataclass  # type: ignore[misc]
 class Message(ABC):
     """Parent level class for all Message objects.
 
--- a/ot3_state_manager/ot3_state_manager/messages.py
+++ b/ot3_state_manager/ot3_state_manager/messages.py
@@ -6,42 +6,60 @@ Also, location for all message handling functions.
 from __future__ import annotations
 
 import struct
+from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Any, Dict, Literal, Optional
+from enum import Enum
+from typing import Any, Callable, Optional
 
 from opentrons.hardware_control.types import OT3Axis
 
 from ot3_state_manager.ot3_state import OT3State
-from ot3_state_manager.util import Direction
+from ot3_state_manager.util import Direction, MoveMessageHardware, SyncPinState
+
+MESSAGE_BYTE_LENGTH = 4
+MESSAGE_ID_BYTE_LENGTH = 1
+MESSAGE_CONTENT_BYTE_LENGTH = 3
+STRUCT_FORMAT_STRING = f"B{MESSAGE_CONTENT_BYTE_LENGTH}s"
 
 
-@dataclass
-class Message:
-    """Generic Message Type."""
+# mypy doesn't like that I am instantiating a dataclass based on an abstract class (ABC)
+# instead of a concrete class.
+@dataclass  # type: ignore[misc]
+class Message(ABC):
+    """Parent level class for all Message objects.
 
+    Is abstract base class that requires both a build and to_bytes method to be defined.
+    """
+
+    @staticmethod
+    def parse(message_bytes: bytes) -> Message:
+        """Parse sequence of bytes into a Message object."""
+        if not len(message_bytes) == MESSAGE_BYTE_LENGTH:
+            raise ValueError(
+                f"Message length must be {MESSAGE_BYTE_LENGTH} bytes. "
+                f"Your message was {len(message_bytes)} bytes."
+            )
+        message_id, message_content = struct.unpack(STRUCT_FORMAT_STRING, message_bytes)
+        # mypy thinks that I am instantiating a MoveMessageHardware object here.
+        # I am not, I am performing a lookup as defined in the enum docs.
+        # https://docs.python.org/3/library/enum.html#programmatic-access-to-enumeration-members-and-their-attributes
+        return MessageID(message_id).builder_func(message_content=message_content)  # type: ignore[call-arg]
+
+    @staticmethod
+    @abstractmethod
+    def build(message_content: bytes) -> Message:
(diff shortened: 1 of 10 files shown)


ot3_state_manager/ot3_state_manager/messages.py
```diff
@@ -52,55 +45,80 @@ class MoveMessage(Message):
...
+
+    def __hash__(self) -> int:
+        """Return hash for instance."""
+        return hash(str(self))
+
+
+class MessageLookup:
+    """Lookup class for mapping a message to a byte value."""
+
+    MESSAGE_LOOKUP_DICT: Dict[bytes, Message] = {
```
sfoster1, line 62: so this isn't quite what we'd be looking for. The key here is `b"0x0000"`, which is the sequence of bytes representing the UTF-8 codepoint sequence "0x0000", which is the bytes (note, no quotes) 0x307830303030 , 6 bytes total - one for each character.

What we want is the hex sequence 0x0000.

This can be represented in python with the sequence `b"\x00\x00"` using hex escapes (see [https://docs.python.org/3/reference/lexical_analysis.html](the docs) in the "Escape Sequences" table which for some reason doesn't have a direct link anchor), and 0x0001 could be `b"\x00\x01"`, etc; but this gets a little clunky, and instead i'd recommend using the `[struct](https://docs.python.org/3/library/struct.html#struct.pack)` module, whose job is to pack data into and parse data out of byte strings.

Typically you wouldn't want to put byte strings in lookups like this directly; rather, you'd take the incoming string and do `struct.unpack('<H', inputbytes)`, which would tell struct to interpret the inputbytes as a two-byte short integer, and turn it into a python integer, which could then be a struct key.

ot3_state_manager/ot3_state_manager/messages.py
```diff
@@ -52,55 +45,80 @@ class MoveMessage(Message):
...
-    sync_pin_message = SyncPinMessage.to_message(string)
-    move_message = MoveMessage.to_message(string)
+        return cls.REVERSE_LOOKUP_DICT[message]
 
-    if sync_pin_message is not None:
-        return sync_pin_message
+    @classmethod
+    def bytes_to_message(cls, byte_string: bytes) -> Message:
+        """Looks up and returns message for byte value."""
+        if byte_string not in cls.MESSAGE_LOOKUP_DICT:
```
sfoster1, line 100: so as with what we discussed above, we'd make this method look like
```
message_key = struct.unpack('<H', byte_string)
return cls.MESSAGE_LOOKUP_DICT[message_key]
```

sfoster1 (changes requested):
We have to change the way we handle the byte strings here, but this is definitely on the right track

DerekMaggio:
@sfoster1 is the format string `<H` correct? Do we actually want it to be Big Endian? 

```python
int_val = 0xFF00
print(struct.pack(">H", int_val))
print(struct.pack("<H", int_val))
```

```
>>> b'\xff\x00'
>>> b'\x00\xff'
```

With the way we have been talking about the messages, we want the message to be formatted like the first print statement (Little Endian)

ot3_state_manager/ot3_state_manager/messages.py
```diff
@@ -52,55 +46,94 @@ class MoveMessage(Message):
...
+    def __hash__(self) -> int:
+        """Return hash for instance."""
+        return hash(str(self))
+
+
+class MessageLookup:
+    """Lookup class for mapping a message to a byte value."""
+
+    MESSAGE_LOOKUP_DICT: Dict[int, Message] = {
+        0x0000: MoveMessage(OT3Axis.X, Direction.NEGATIVE),
```
sfoster1, line 64: rather than having the entirety of a binary message be a key in this dict, let's separate this into layers:
- Define the message binary format as having a `message-id` part (one or two bytes) that identifies the _type_ of the message, and a `message-content` part (the rest of the sequence) that contains the _content_ of the message
- Have a dict whose keys are the `message-id` part, and the values are dataclasses with a `build_from_bytes` static method that take a `bytes` instance; not dataclass instances, but the class types themselves
- Then the deserialization process is:
```
message = read_bytes(...)
message_id = struct.unpack('<H', message[:2])  # unpack the two bytes of message id
message_class = MESSAGE_LOOKUP_DICT[message_id]  # find the class
message = message_class.build_from_bytes(message[2:])
````
the dict can be typed in one of two ways:
- You can build an inheritance hierarchy for messages
- You can call it a typed dict with integer literals keys mapping to different class types. this might end up being a nightmare, fair warning

File: ot3_state_manager/ot3_state_manager/messages.py [hunk 4]
```diff
@@ -52,55 +46,94 @@ class MoveMessage(Message):
...
-    """Parse string into a Message Type"""
-    sync_pin_message = SyncPinMessage.to_message(string)
-    move_message = MoveMessage.to_message(string)
-
-    if sync_pin_message is not None:
-        return sync_pin_message
+    @classmethod
+    def message_to_bytes(cls, message: Message) -> bytes:
+        """Looks up and returns byte value for message."""
+        if message not in cls.REVERSE_LOOKUP_DICT:
```
sfoster1, line 95: do we need something like this? if it doesn't exist, you get a `KeyError` later right?

ot3_state_manager/ot3_state_manager/messages.py
```diff
@@ -52,55 +46,94 @@ class MoveMessage(Message):
...
 
-    raise ValueError(f'Not able to parse message: "{string}"')
+    @classmethod
+    def bytes_to_message(cls, byte_string: bytes) -> Message:
+        """Looks up and returns message for byte value."""
+        try:
+            val_tuple = struct.unpack(cls.FORMAT_STRING, byte_string)
+        except struct.error:
+            raise ValueError(
+                f'Passed message, "{byte_string!r}" is not a valid hexadecimal byte '
```
sfoster1, line 107: bit of a nitpick, but the input value is _not_ a `hexadecimal byte string` - it is just a byte string. what "validity" means in this context is that `byte_string` had fewer than 2 elements, and something like that is what the error should say.

ot3_state_manager/ot3_state_manager/messages.py
```diff
@@ -52,55 +46,94 @@ class MoveMessage(Message):
...
+                f"string."
+            )
+
+        assert len(val_tuple) == 1
```
sfoster1, line 111: similarly, the only way this can _not_ have precisely one element is if you got the `struct.error` exception that the string was too short. we can just say `val_tuple, = struct.unpack(...)` with that `,` turning it into a tuple destructure

ot3_state_manager/ot3_state_manager/messages.py
```diff
@@ -52,55 +46,94 @@ class MoveMessage(Message):
...
+        int_val = val_tuple[0]
+        if int_val not in cls.MESSAGE_LOOKUP_DICT:
```
sfoster1, line 113: all this does is transform a `KeyError` into a `ValueError` - I don't really see the purpose

sfoster1 (changes requested):
I think we're almost there but there's a couple subtle changes we should make.

ot3_state_manager/ot3_state_manager/messages.py
DerekMaggio: Removed in 2cfb104355ce884c880cf4f82bae2112146bce4c

ot3_state_manager/ot3_state_manager/messages.py
DerekMaggio: Removed in 2cfb104355ce884c880cf4f82bae2112146bce4c

ot3_state_manager/ot3_state_manager/messages.py
DerekMaggio: Removed in 2cfb104355ce884c880cf4f82bae2112146bce4c

ot3_state_manager/ot3_state_manager/messages.py
DerekMaggio: Removed in https://github.com/Opentrons/ot3-firmware/commit/2cfb104355ce884c880cf4f82bae2112146bce4c

DerekMaggio:
@sfoster1 I feel like I should pull the `parse` method out of the `Message` class and make it a pure function. 
I don't think that the concrete implementations of the `Message` class should have access to the `parse` function since the `parse` function has a larger scope than the concrete class should have. 

What do you think?

ot3_state_manager/ot3_state_manager/messages.py
```diff
@@ -5,102 +5,156 @@
...
-from ot3_state_manager.util import Direction, get_md5_hash
+from ot3_state_manager.util import Direction, MoveMessageHardware, SyncPinState
+
+MESSAGE_BYTE_LENGTH = 4
+MESSAGE_ID_BYTE_LENGTH = 1
+MESSAGE_CONTENT_BYTE_LENGTH = 3
+STRUCT_FORMAT_STRING = f"B{MESSAGE_CONTENT_BYTE_LENGTH}s"
+
+
+# mypy doesn't like that I am instantiating a dataclass based on an abstract class (ABC)
```
sfoster1, line 25: why not? is it a bug in mypy, or is there a good reason for it?

ot3_state_manager/ot3_state_manager/messages.py
```diff
@@ -5,102 +5,156 @@
...
+
+    def __hash__(self) -> int:
+        """Return hash for instance."""
+        return hash(str(self))
+
+    @staticmethod
+    def build(message_content: bytes) -> MoveMessage:
+        """Convert message_content into a MoveMessage object."""
+        hw_id, direction_val = struct.unpack(">HB", message_content)
+        # mypy thinks that I am instantiating a MoveMessageHardware object here.
```
sfoster1, line 81: But those are the same thing. what's the error?

ot3_state_manager/ot3_state_manager/messages.py
```diff
@@ -5,102 +5,156 @@
...
+        return hash(str(self))
 
-            try:
-                direction = Direction.from_symbol(split_string[0])
-            except ValueError:
-                direction = None
+    @staticmethod
+    def build(message_content: bytes) -> SyncPinMessage:
+        """Convert message_content into a SyncPinMessage object."""
+        # Have to use from_bytes method because struct package won't accept a 3 byte val
```
sfoster1, line 111: that seems a bit annoyingly inconsistent - perhaps we should just make it four bytes

ot3_state_manager/ot3_state_manager/messages.py
```diff
@@ -5,102 +5,156 @@
...
-        return None
 
+class MessageID(Enum):
+    """Enum class defining the relationship between the message_id byte and the corresponding Message object."""
 
-def parse_message(string: str) -> Union[SyncPinMessage, MoveMessage]:
-    """Parse string into a Message Type"""
-    sync_pin_message = SyncPinMessage.to_message(string)
-    move_message = MoveMessage.to_message(string)
+    def __new__(cls, message_id: int, builder_func: Callable) -> MessageID:
```
sfoster1, line 125: we probably want a mypy protocol or some other richer type to describe callable; that's likely the source of the type confusion above.

ot3_state_manager/ot3_state_manager/util.py
```diff
@@ -3,28 +3,92 @@
...
 
-    POSITIVE = "+"
-    NEGATIVE = "-"
+    POSITIVE = 1
+    NEGATIVE = 0
+
+    @classmethod
+    def from_int(cls, direction_val: int) -> Direction:
+        """Parse integer into direction."""
+        if direction_val == Direction.NEGATIVE:
```
sfoster1, line 20: can this be something more like 
```
if direction_val not in (Direction.NEGATIVE.value, Direction.POSITIVE.value):
    raise ValueError(...)
return Direction(direction_val)
```

ot3_state_manager/ot3_state_manager/util.py
```diff
@@ -3,28 +3,92 @@
...
     @classmethod
-    def from_symbol(cls, symbol: str) -> "Direction":
-        """Parse plus or minus symbol into direction."""
-        if symbol == "+":
-            return cls.POSITIVE
-        elif symbol == "-":
-            return cls.NEGATIVE
+    def from_int(cls, state_val: int) -> SyncPinState:
+        """Parse integer into sync pin state."""
+        if state_val == 0:
```
sfoster1, line 41: same as above with direction

ot3_state_manager/ot3_state_manager/util.py
```diff
@@ -3,28 +3,92 @@
...
 
-def get_md5_hash(data: bytes) -> bytes:
-    """Convert bytes into a md5 hash."""
-    return hashlib.md5(data).hexdigest().encode()
+
+@unique
+class MoveMessageHardware(Enum):
+    """Enum representing a mapping of an integer hardware id to an OT3Axis object."""
+
+    def __new__(cls, hw_id: int, axis: OT3Axis) -> MoveMessageHardware:
```
sfoster1, line 58: can we do this in a way that's easier to typecheck? perhaps as two objects, or with helper functions?

sfoster1 (approved):
This looks much much better structurally but there's some things to fix up.

sfoster1 (changes requested):
Sorry - last review was supposed to be this status

ot3_state_manager/ot3_state_manager/messages.py
DerekMaggio: It's based off of this [issue](https://github.com/python/mypy/issues/5374). Some people think it is a bug some don't. 

Based off of [this comment](https://github.com/python/mypy/issues/5374#issuecomment-568335302) I can turn the dataclass into a mixin and add it to a regular ABC. But it seemed messy.

ot3_state_manager/ot3_state_manager/messages.py
DerekMaggio: If I make message_content 4 bytes then the hw_id portion of the MoveMessage becomes 3 bytes which leads me to the same issue. 
Instead of making it 4 bytes, I am going to prepend a `\x00` to the `message_content` of the `build` method of `SyncPinMessage`. This will make it 4 bytes, making it parseable by the struct module, without changing the integer value of it

DerekMaggio: e33ed939761988879d0989afccd96b647c613005

ot3_state_manager/ot3_state_manager/util.py
DerekMaggio: add3d48d0400a12a4af54bf5312c0140c5418fb1

ot3_state_manager/ot3_state_manager/util.py
DerekMaggio: add3d48d0400a12a4af54bf5312c0140c5418fb1

DerekMaggio: add3d48d0400a12a4af54bf5312c0140c5418fb1

ot3_state_manager/ot3_state_manager/messages.py
DerekMaggio: Instead of doing a mypy protocol, I just passed the whole Message object, which requires a `build_message` method. And in the __init__ set the builder_func to the `build_message` method.

ot3_state_manager/ot3_state_manager/messages.py
DerekMaggio: The error is that the `axis` parameter is not defined.

ot3_state_manager/ot3_state_manager/util.py
DerekMaggio: Removed `__new__` method and added helper functions to lookup by id.
26c23bb154ed560b05060954bc2d0f4c682ec017

ot3_state_manager/ot3_state_manager/messages.py
DerekMaggio: Removed `__new__` method and added helper functions to lookup by id. Making this a non-issue
26c23bb154ed560b05060954bc2d0f4c682ec017

ot3_state_manager/ot3_state_manager/messages.py
DerekMaggio: I ended up going a totally different route. 
`Message` does not need to be a dataclass. It doesn't have any data in it. 

Just making it an AbstractBaseClass that the concrete Message classes inherit from. 

5e9d5bd084a9179e33239417c6fa40880c955a57

ot3_state_manager/ot3_state_manager/util.py
```diff
@@ -3,28 +3,81 @@
...
+    # hw_id, axis
+    X_AXIS = 0x0000, OT3Axis.X
+    Y_AXIS = 0x0001, OT3Axis.Y
+    Z_L_AXIS = 0x0002, OT3Axis.Z_L
+    Z_R_AXIS = 0x0003, OT3Axis.Z_R
+    Z_G_AXIS = 0x0004, OT3Axis.Z_G
+    P_L_AXIS = 0x0005, OT3Axis.P_L
+    P_R_AXIS = 0x0006, OT3Axis.P_R
+    G_AXIS = 0x0007, OT3Axis.G
+    Q_AXIS = 0x0008, OT3Axis.Q
```
caila-marashaj, line 63: What does Q axis represent here?

DerekMaggio: To my understanding, the Q axis is the high-throughput pipette axis.
https://github.com/Opentrons/opentrons/blob/edge/api/src/opentrons/hardware_control/types.py#L128

sfoster1 (approved):
Looks great! Definitely the right way to do it, very solid

PR #434, opened 2022-08-09
fix(head): motor driver & encoder config

ahiuchingau:
Makes the following changes to the Rev1 board:
1. Reverts the ihold irun value to match the proto board to solve motor movement issues
2. Flips the encoder timer (correct config: left -> tim3, right -> tim2)
3. Adds encoder config so we can actually get encoder position from Ack messages


Laura-Danielle (approved):
lgtm!

sfoster1:
What does the ihold/irun do, given that these are tmc2160s and we wanted to control them through the global scaler while keeping those registers at fixed values?

PR #435, opened 2022-08-10
feat: Add header file generation for OT3StateManager Messages

DerekMaggio:
# Overview

Adds header file generation for OT3StateManager messages

# Changelog

- Pulled general functions from generate_header.py and put the in header_generation_utils.py
- Created generate_state_manager_headers.py
- Added command to Makefile to generate headers and put them in include/common/simulation
- Added generate_state_manager_headers.py script to linting and formatting

# Review Requests

None

# Risk

Low, as long as this doesn't break the CAN header generation. I looked at the generated file and it looks the same to me.


sfoster1 (changes requested):
You're right about the py.typed thing; the way to fix it is to put that file in an actual python package, which i think is a pretty good idea anyway.

I think the biggest thign I'd change here is that the header should be just-in-time generated and not checked in. That way you don't have to frequently check in changes in two places. You can do this with a `cmake_custom_command` pretty easily (see the way the `version` is generated).

Laura-Danielle:
Do we need the `py.typed` file?

DerekMaggio:
> Do we need the `py.typed` file?

Yes. I get the following when I run `make lint` without it

```
poetry run python -m mypy ot3_state_manager tests/ generate_state_manager_headers.py
generate_state_manager_headers.py:8: error: Skipping analyzing "header_generation_utils.header_generation_utils": module is installed, but missing library stubs or py.typed marker  [import]
generate_state_manager_headers.py:8: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 15 source files)
```

Laura-Danielle:
> > Do we need the `py.typed` file?
> 
> Yes. I get the following when I run `make lint` without it
> 
> ```
> poetry run python -m mypy ot3_state_manager tests/ generate_state_manager_headers.py
> generate_state_manager_headers.py:8: error: Skipping analyzing "header_generation_utils.header_generation_utils": module is installed, but missing library stubs or py.typed marker  [import]
> generate_state_manager_headers.py:8: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
> Found 1 error in 1 file (checked 15 source files)
> ```

Shouldn't you be able to add a `toml` file to satisfy seth's request: `the way to fix it is to put that file in an actual python package, which i think is a pretty good idea anyway.`

DerekMaggio:
> > > Do we need the `py.typed` file?
> > 
> > 
> > Yes. I get the following when I run `make lint` without it
> > ```
> > poetry run python -m mypy ot3_state_manager tests/ generate_state_manager_headers.py
> > generate_state_manager_headers.py:8: error: Skipping analyzing "header_generation_utils.header_generation_utils": module is installed, but missing library stubs or py.typed marker  [import]
> > generate_state_manager_headers.py:8: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
> > Found 1 error in 1 file (checked 15 source files)
> > ```
> 
> Shouldn't you be able to add a `toml` file to satisfy seth's request: `the way to fix it is to put that file in an actual python package, which i think is a pretty good idea anyway.`

I think I fixed it in this PR

CMakeLists.txt
```diff
@@ -141,3 +141,8 @@ add_custom_target(update-headers
         VERBATIM
         COMMAND ${SHELL_FOR_PYTHON} -c "python ./generate_header.py ${CMAKE_SOURCE_DIR}/include/can/core/ids.hpp --source-dir=${OPENTRONS_HARDWARE_PATH} --language=c++"
         COMMAND ${SHELL_FOR_PYTHON} -c "python ./generate_header.py ${CMAKE_SOURCE_DIR}/include/bootloader/core/ids.h --source-dir=${OPENTRONS_HARDWARE_PATH} --language=c")
+
+add_custom_target(update-ot3-state-manager-headers
```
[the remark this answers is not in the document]
sfoster1: finally, once you've done this (and this may not be a this-pr thing) it becomes a lot easier to bring the poetry stuff up into cmake rather than living in its own little makefile world. once you have a subdirectory like this, you can do the poetry install stuff in [https://cmake.org/cmake/help/latest/command/execute_process.html](execute_process) so it runs automatically.

Then, you can tell cmake that it must regenerate the project if the pyproject.toml` or `poetry.lock` files change by adding those files to [CMAKE_CONFIGURE_DEPENDS](https://cmake.org/cmake/help/latest/prop_dir/CMAKE_CONFIGURE_DEPENDS.html): `set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS pyproject.toml poetry.lock)`

At that point, you should see that when running `cmake --preset=simulators`, cmake runs poetry install; and that if you edit either file, running `cmake --build --preset=simulators` will rerun project generation, so you'll get any changed dependencies.

ot3_state_manager/pyproject.toml
```diff
@@ -28,6 +30,7 @@ mdformat_footnote = "^0.1.1"
 # Expects the monorepo (opentrons) to be a sibling of this repo
 opentrons-hardware = { path = "../../opentrons/hardware" }
 opentrons = { path = "../../opentrons/api" }
+header_generation_utils = { path = "../header_generation_utils"}
```
sfoster1, line 33: this dep, and probably `opentrons` and `opentrons-hardware`, want to be [`editable = True`](https://python-poetry.org/docs/cli/#add) if possible

CMakeLists.txt
```diff
@@ -141,3 +141,8 @@ add_custom_target(update-headers
         VERBATIM
         COMMAND ${SHELL_FOR_PYTHON} -c "python ./generate_header.py ${CMAKE_SOURCE_DIR}/include/can/core/ids.hpp --source-dir=${OPENTRONS_HARDWARE_PATH} --language=c++"
```
sfoster1, line 142: note for the future: we should make this a nice poetry script now that it's in a language environment

sfoster1 (changes requested):
The code looks more or less good (though we should change some names) but there's some stuff we should fix up for the poetry interaction changes.

ot3_state_manager/generate_state_manager_headers.py
```diff
@@ -0,0 +1,66 @@
...
+from header_generation_utils.utils import generate_file_comment, write_enum_cpp
+
+from ot3_state_manager.messages import MessageID
+from ot3_state_manager.util import Direction, MoveMessageHardware, SyncPinState
+
+
+def get_message_id_enum() -> Type[Enum]:
+    """Creates a MessageID enum."""
+    message_ids = [
+        ("_".join([message.name, "message_id"]).lower(), message.message_id)
```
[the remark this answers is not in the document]
DerekMaggio: Are you meaning turn it into something like 
```
enum class MessageID {
    move = 0x0,
    sync_pin = 0x1,
    get_axis_location = 0x2,
    get_sync_pin_state = 0x3,
};
```
instead of 
```
enum class MessageID {
    move_message_id = 0x0,
    sync_pin_message_id = 0x1,
    get_axis_location_message_id = 0x2,
    get_sync_pin_state_message_id = 0x3,
};
```
because the `message_id` part is redundant?

ot3_state_manager/pyproject.toml
DerekMaggio: https://github.com/Opentrons/ot3-firmware/pull/435/commits/2effe2974c6e39769347d3cf149a189484f6d776

FYI, it is `develop=true`

CMakeLists.txt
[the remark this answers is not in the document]
DerekMaggio: https://github.com/Opentrons/ot3-firmware/pull/435/commits/083d1708763c5ac30b95a646983690a371635705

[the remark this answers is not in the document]
DerekMaggio: > this isn't quite the right way to do this in cmake, for a couple reasons:
> 
> * The output gets emitted to the source directory. this requires a gitignore and is generally not great practice; you really want only stuff you check in to live in the source directory. That way, you can `rm` the build directory and be confident everything did get wiped out. I'd recommend doing the following:
>   
>   1. Put this command, and others related to `ot3_state_manager`, in `ot3_state_manager/CMakeLists.txt` (and `add_subdirectory` it, ideally guarded by `if(NOT ${CMAKE_CROSSCOMPILING})`
>   2. In that `CMakeLists.txt`, generate this header file to `${CMAKE_CURRENT_BINARY_DIR}`
>   3. use `add_library(ot3_state_manager INTERFACE)` and then `target_include_directories(ot3_state_manager INTERFACE ${CMAKE_CURRENT_BINARY_DIR}` to create a special header-only library that only exists to propagate this include
>   4. Add `target_link_libraries(${SIMEXE} ot3_state_manager)`, replacing `${SIMEXE}` with the name of the simulator executable, to each simulator executable in its cmakelists
> 
> By doing this, you get the following benefits:
> 
> 1. The simulator executables have a known-to-cmake explicit dependency on the generated header, which means if you update the header they should get rebuilt
> 2. you can get rid of the gitignore because the header is in the build dir

@sfoster1 I made the changes in https://github.com/Opentrons/ot3-firmware/pull/435/commits/083d1708763c5ac30b95a646983690a371635705 but I am having trouble testing if the header file is actually being created. 

I ran `cmake --preset=host-gcc10 .` then `cmake --build --preset=simulators` and I can't find the header file anywhere

DerekMaggio:
[OT3StateManager Messages Header File Generation Pull Request](https://github.com/Opentrons/ot3-firmware/pull/435) adds the dependency of the monorepo directory being a sibling to the `ot3-firmware` directory.  This is because I am using the `OT3Axis` enum from the monorepo to generate axis designations in the header files. I will also soon be pulling Pipette Models from the monorepo as well.

I have addressed this dependency in almost all Github Actions in [this commit](https://github.com/Opentrons/ot3-firmware/pull/435/commits/483d02fde2663cfac4a22de141c54e1f62fe3bd1).

Where the issue lies is the `run_emulation.yaml` Github Action which confirms that emulation can be spun up with the current state of `ot3-firmware`.

Originally, all emulation containers required only 1 source repo (`opentrons`, `ot3-firmware`, or `opentrons-modules`) to function. Now, in order to be built, all hardware simulators from `ot3-firmware` require that the monorepo be a sibiling to `ot3-firmware`.

Because of this change `run_emulation.yaml` is now failing. In order to get it working, `opentrons-emulation` requires some relatively large changes.

ot3_state_manager/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+	ot3-state-manager-format
+	# Have to wrap this isort command in a bash -c for some reason because otherwise it fails on the -m for no reason
+	# other than cmake is dumb sometimes.
+	bash -c "${Poetry_EXECUTABLE} run python -m isort ot3_state_manager tests generate_state_manager_headers.py"
+	${Poetry_EXECUTABLE} run python -m black ot3_state_manager tests generate_state_manager_headers.py
+	${Poetry_EXECUTABLE} run python -m mdformat README.md
+	WORKING_DIRECTORY ../../ot3_state_manager
+)
+
+add_dependencies(ot3-state-manager-test ot3-state-manager-setup)
```
sfoster1, line 75: these dependencies mean that the setup task will be run whenever you run any of the other ones - it never gets saved as "already did that". it's like make in that way. as with the makefiles in the monorepo, we shouldn't explicitly have a `setup` phony target (which is what `add_custom_target` creates) as dependency to other dtargetrs.

ot3_state_manager/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+
+add_custom_target(
+	ot3-state-manager-setup
+	COMMAND ${Poetry_EXECUTABLE} env use 3.8
+	COMMAND ${Poetry_EXECUTABLE} update
+	COMMAND ${Poetry_EXECUTABLE} install
+	WORKING_DIRECTORY ../../ot3_state_manager
+)
+
+add_custom_target(
```
sfoster1, line 29: if we can get the venv to live in a binary directory then cmake will happily remove it for us

ot3_state_manager/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+	)
+add_custom_target(state-manager-headers DEPENDS ${GENERATED_MESSAGE_HEADER})
+# By depending on this target, you'll be able to include the generated file
+# with no path qualifiers (i.e. #include "ot3_state_manager_messages.hpp")
+add_library(ot3_state_manager INTERFACE)
+add_dependencies(ot3_state_manager state-manager-headers)
+target_include_directories(ot3_state_manager INTERFACE ${CMAKE_CURRENT_BINARY_DIR})
+
+add_custom_target(
+	ot3-state-manager-setup
```
sfoster1, line 22: if you make this run `COMMAND ${Poetry_EXECUTABLE} config virtualenvs.path = ${CMAKE_BINARY_DIR}/virtualenvs --local` then we'll get the virtualenv in the binaries dir, and we won't need to teardown (you can remove that dir as you always do if you want)

ot3_state_manager/generate_state_manager_headers.py
```diff
@@ -0,0 +1,62 @@
...
+from enum import Enum
+from typing import Type
+
+from header_generation_utils.utils import generate_file_comment, write_enum_cpp
+
+from ot3_state_manager.messages import MessageID
+from ot3_state_manager.util import Direction, MoveMessageHardware, SyncPinState
+
+
+def get_message_id_enum() -> Type[Enum]:
```
sfoster1, line 14: why do we need to do this? isn't `MessageId` already an enum?

ot3_state_manager/ot3_state_manager/messages.py
```diff
@@ -102,6 +117,54 @@ def to_bytes(self) -> bytes:
         return struct.pack(">BHB", MessageID.SYNC_PIN.message_id, 0, self.state)
 
 
+@dataclass
+class GetAxisLocationMessage(Message):
+    """Message to get current location of axis."""
+
+    axis: OT3Axis
+
+    def __eq__(self, other: Any) -> bool:
```
sfoster1, line 126: do we really need this? dataclasses should generate this for free. same with hash.

ot3_state_manager/ot3_state_manager/messages.py
```diff
@@ -137,24 +202,29 @@ def _parse_message(message_bytes: bytes) -> Message:
...
+        response.is_error = True
+        response.content = f"{err.args[0]}"
     except:  # noqa: E722
-        error_response = "Unhandled Exception"
+        response.is_error = True
+        response.content = "Unhandled Exception"
     else:
         if isinstance(message, MoveMessage):
             ot3_state.pulse(message.axis, message.direction)
+            response.content = data.decode()
```
sfoster1, line 217: maybe a different pr, but i'd recommend pushing this behavior down into the classes somehow - give them a `handle` function so we don't need a big if/elif block that we can't really statically verify

sfoster1 (changes requested):
Comments inline, but also
- Rather than adding a `FindPoetry.cmake` here, we should remove both these files and just update the subtree that holds `cmake-utils`.
- feels redundant to have everything called `ot3_*`, the repo already says ot3. will this python project be used outside this repo?

ot3_state_manager/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+	COMMAND ${Poetry_EXECUTABLE} env use 3.8
+	COMMAND ${Poetry_EXECUTABLE} update
+	COMMAND ${Poetry_EXECUTABLE} install
```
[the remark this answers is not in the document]
DerekMaggio: `poetry install` doesn't have `--sync` or `--dev` options

ot3_state_manager/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+# By depending on this target, you'll be able to include the generated file
+# with no path qualifiers (i.e. #include "ot3_state_manager_messages.hpp")
+add_library(ot3_state_manager INTERFACE)
+add_dependencies(ot3_state_manager state-manager-headers)
+target_include_directories(ot3_state_manager INTERFACE ${CMAKE_CURRENT_BINARY_DIR})
+
+add_custom_target(
+	ot3-state-manager-setup
+	COMMAND ${Poetry_EXECUTABLE} env use 3.8
+	COMMAND ${Poetry_EXECUTABLE} update
```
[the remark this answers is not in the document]
DerekMaggio: Modified `state-manager-setup` to not alter the lockfile.
Added `state-manager-update` command that will modify the lockfile
9a4527f272a20c797d599a86b420e4d6ad2ada05

ot3_state_manager/generate_state_manager_headers.py
DerekMaggio: See comments added in following commit
21aaf6cbb6418fb6650ce0ab1fd8fbd38e470935

ot3_state_manager/ot3_state_manager/messages.py
DerekMaggio: e540a74f5db3251c4db29cbc66ae381c83d2bcba

ot3_state_manager/CMakeLists.txt
DerekMaggio: 675b0c4ad94de42806749397190d29a69a3d2d29

ot3_state_manager/CMakeLists.txt
DerekMaggio: 675b0c4ad94de42806749397190d29a69a3d2d29

ot3_state_manager/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+	WORKING_DIRECTORY ../../ot3_state_manager
+)
+
+add_custom_target(
+	ot3-state-manager-build
+	COMMAND ${Poetry_EXECUTABLE} build
+	WORKING_DIRECTORY ../../ot3_state_manager
+)
+
+add_custom_target(
```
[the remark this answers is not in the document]
DerekMaggio: @sfoster1 
I am struggling to get the clean command to work. 

I added 

```cmake
file(GLOB_RECURSE EGGS RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} "*.egg-info")
file(GLOB_RECURSE PYC RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} "*.pyc")
set_property(DIRECTORY APPEND PROPERTY ADDITIONAL_MAKE_CLEAN_FILES
	${EGGS}
	${PYC}
    "build"
    "dist"
    ".coverage"
    "coverage.xml"
    "__pycache__"
    "*.pyc"
    ".mypy_cache"
)
```
to the state_manager/CmakeLists.txt

I am running `cmake --build --preset state-manager --target clean`, and it seems to execute but my files are not removed.

ot3_state_manager/CMakeLists.txt
DerekMaggio: I am not understanding how the `setup` target is phony since it is defined above in the file.

ot3_state_manager/ot3_state_manager/messages.py
DerekMaggio: 7586e06b0aa8bfedfb2b8d9946e7750010978aaa

ot3_state_manager/CMakeLists.txt
DerekMaggio: c803840c2aa3a2a59cf199437f4b969c247896ae

sfoster1: So how do you square this circle, and make sure that
- Before you run anything in poetry, `poetry install` has been run
- But you don't run `poetry install` unnecessarily?
Well, there's two ways. One way is the build-time way, and the other way is the configure-time way.

The build-time way is more complex, but means you could run `poetry install` without doing a full project reconfigure. What you'd do is actually use _two_ commands: [`add_custom_command`](https://cmake.org/cmake/help/latest/command/add_custom_command.html) and `add_custom_target`.

First, you'd use `add_custom_command` to tell cmake about the project virtualenv that poetry creates:
```
# this must be the name of a _file_; pick something that will always be in the virtualenv like its activation script.
# we know where the virtualenv will be because we're using the local config to set it. it's probably a good idea
# to put this file path in a variable since we'll need it later
set(virtualenv_file path/to/some/file/in/virtualenv/like/bin/activate})

add_custom_command(
    # by using a custom command with an explicit output, cmake can look at the output file to decide if the command needs
    # to run
    OUTPUT ${virtualenv_file}
    # by telling cmake this command depends on pyproject.toml and poetry.lock, we can get cmake to rerun it if
    # they change
    DEPENDS pyproject.toml poetry.lock
    # the same commands you're using in add_custom_target go here
    COMMAND poetry install
)
```
The file that you use to signify "this virtualenv is built" is now something you put as a dependency in future `add_custom_target`s (this can kind of mess up parallel runs but that's fine for this use case), so
```
add_custom_target(
   ot3-state-manager-test
   COMMANDS poetry py.test
   DEPENDS ${virtualenv_file}
)
```

The configure-time way is actually much easier and fits better into cmake's view of the world. In cmake's view of the world, "configuration" is what you do to make your project ready to run _just the commands that actually make code_. In that view, `poetry install` is properly part of configuration, because `poetry run py.test` and `poetry run pylint` and `poetry build` are the commands that actually build code. By making the install happen during configuration time, you can be completely certain that it has been run by the time you run `cmake --build --target ot3-state-manager-test`. What you'd do is... call `poetry install` using [execute_process](https://cmake.org/cmake/help/latest/command/execute_process.html). That's it! This will run when you run `cmake` (note: _not_ `cmake --build`). Rerunning `cmake` will rerun it. That's all. The downside is that to re-install, you have to run `cmake` again, and this doesn't have dependency detection so it will run whenever you run `cmake`. I'd try the build time way first.

ot3_state_manager/CMakeLists.txt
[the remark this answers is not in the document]
DerekMaggio: 387f50dbedcfd9317b0349321dbf900cf1956948

Run it with `cmake --build --preset state-manager && cmake --build ./build-host/ --target clean`

CMakePresets.json
```diff
@@ -267,6 +267,22 @@
         "pipettes-384-simulator",
         "gripper-simulator"
       ]
+    },
+    {
```
sfoster1, line 271: Oh wait this doesn't really make sense to me - this is fundamentally part of simulation right? should be in the `simulators` preset, since this builds headers that those need

sfoster1 (approved):
Looks good to me modulo the inline comments

.github/workflows/run_emulation.yaml
```diff
@@ -16,22 +18,22 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: Opentrons/opentrons-emulation
-          ref: release-v2.3.1
+          ref: ot3-firmware-header-creation-script-compatability-RQA-219
```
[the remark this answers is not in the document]
DerekMaggio: Will update to `release-v2.3.3` after https://github.com/Opentrons/opentrons-emulation/pull/194 is merged

CMakePresets.json
DerekMaggio: I didn't even think of this. 
This makes setup so much easier if I add the `setup` target to this.

Thank you!!!

CMakeLists.txt
```diff
@@ -69,6 +69,16 @@ add_subdirectory(eeprom)
 add_subdirectory(bootloader)
 list(REMOVE_DUPLICATES LINT_TARGETS)
 
+message(CHECK_START "Checking if cross-compiling.")
```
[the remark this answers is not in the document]
DerekMaggio: https://github.com/Opentrons/ot3-firmware/pull/435/commits/4f88b6869b995373c875d0fefd4cb4243c3ae8c9

CMakePresets.json
DerekMaggio: https://github.com/Opentrons/ot3-firmware/pull/435/commits/61f1b08b05a0d92c4a4a02a25275965f56a529d2

PR #437, opened 2022-08-16
feat(pipette): Read pipette information from serial

sfoster1:
On the OT-2, we store the literal serial number of a pipette on its EEPROM. When we detect whether it's connected, we read that serial number and parse it out in python.

We could do that here, but we have access to a lot more smarts on the pipette now, so instead let's parse the serial at flash time and store a binary-packed value for the name and model. The datecode part is still stored literally. This is unfortunately right now all based on python-side changes in https://github.com/Opentrons/opentrons/pull/11349

When we're sending the data back, we can therefore pull two ints and a bunch of ascii out of the eeprom and send it on back to the host.

To make it a lot easier to make integration tests for this, we can use a file as the backup for the eeprom. We'll just write() and read() the exact same binary data we'd already send the eeprom. This ends up being a lot of code because, well, c++, but it's not that many actual operations.

When we're using a file backing store, we'd really like to be able to say where it is. That makes it a lot easier to use data from a previous operation, after all. We have this pattern of using environment variables for this, but that's annoying. Instead, let's use `boost::program_options` ([(not that) friendly intro](https://theboostcpplibraries.com/boost.program_options), [truly awful main docs](https://www.boost.org/doc/libs/1_71_0/doc/html/program_options.html)) which will let us parse the command line flags and environment variables in parallel, and add flags for the can and new eeprom stuff.

You can try this out! `run_simulators.sh` has been set up to provide a consistent filename for the backing store, so if you run the socket simulator, you can send a serial write; get the pipette info and see that it's good; stop the simulator and start it back up again, and get the same value. Pretty cool!

# TODO
- [x] tests for the gross filesystem stuff

Closes RET-1169

### DIFF
--- a/can/simlib/transport.cpp
+++ b/can/simlib/transport.cpp
@@ -59,7 +59,8 @@ auto can::sim::transport::create(
     const boost::program_options::variables_map& options)
     -> std::shared_ptr<can::sim::transport::BusTransportBase> {
 #ifdef USE_SOCKETCAN
-    auto channel = options["can-channel"].as<std::string>() auto transport =
+    auto channel = options["can-channel"].as<std::string>();
+    auto transport =
         std::make_shared<can::sim::transport::socketcan::SocketCanTransport<
             freertos_synchronization::FreeRTOSCriticalSection>>(channel);
 #else
--- a/can/simlib/transport.cpp
+++ b/can/simlib/transport.cpp
@@ -1,5 +1,8 @@
 #include "can/simlib/transport.hpp"
 
+#include <string>
+
+#include "boost/program_options.hpp"
 #include "common/core/freertos_synchronization.hpp"
 #ifdef USE_SOCKETCAN
 #include "can/simlib/socketcan_transport.hpp"
@@ -7,37 +10,65 @@
 #include "can/simlib/socket_transport.hpp"
 #endif
 
+namespace po = boost::program_options;
+
+auto can::sim::transport::add_options(po::options_description& cmdline_desc,
+                                      po::options_description& env_desc)
+    -> std::function<std::string(std::string)> {
+#ifdef USE_SOCKETCAN
+    cmdline_desc.add_options()("can-channel,c",
+                               po::value<std::string>()->default_value("vcan0"),
+                               "can channel identifier. May be specified in an "
+                               "environment variable called CAN_CHANNEL.");
+    env_desc.add_options()("CAN_CHANNEL",
+                           po::value<std::string>()->default_value("vcan0"));
+    return [](std::string input_val) -> std::string {
+        if (input_val == "CAN_CHANNEL") {
+            return "can-channel";
+        }
+        return "";
+    }
+#else
+    cmdline_desc.add_options()(
+        "server-host,s", po::value<std::string>()->default_value("localhost"),
+        "can server to connect to. May be specified in an environment variable "
+        "called CAN_SERVER_HOST.")(
+        "port,p", po::value<uint16_t>()->default_value(9898),
+        "port for the can server. May be specified in an environment variable "
+        "called CAN_PORT.");
+    env_desc.add_options()("CAN_SERVER_HOST",
+                           po::value<std::string>()->default_value("localhost"),
+                           "can server to connect to")(
+        "CAN_PORT", po::value<uint16_t>()->default_value(9898));
+    return [](std::string input_val) -> std::string {
+        if (input_val == "CAN_SERVER_HOST") {
+            return "server-host";
+        } else if (input_val == "CAN_PORT") {
+            return "port";
+        }
+        return "";
+    };
+#endif
+}
+
 /**
  * Create simulating bus transport
  * @return pointer to bus transport
  */
-auto can::sim::transport::create()
+auto can::sim::transport::create(
+    const boost::program_options::variables_map& options)
     -> std::shared_ptr<can::sim::transport::BusTransportBase> {
 #ifdef USE_SOCKETCAN
-    auto constexpr ChannelEnvironmentVariableName = "CAN_CHANNEL";
-    auto constexpr DefaultChannel = "vcan0";
-
-    const char* env_channel_val = std::getenv(ChannelEnvironmentVariableName);
-    auto channel = env_channel_val ? env_channel_val : DefaultChannel;
-    auto transport =
+    auto channel = options["can-channel"].as<std::string>() auto transport =
         std::make_shared<can::sim::transport::socketcan::SocketCanTransport<
             freertos_synchronization::FreeRTOSCriticalSection>>(channel);
 #else
-    auto constexpr ServerHostEnvironmentVariableName = "CAN_SERVER_HOST";
-    auto constexpr DefaultServerHost = "localhost";
-    auto constexpr PortEnvironmentVariableName = "CAN_PORT";
-    auto constexpr DefaultPort = 9898;
-
-    const char* env_server_host_val =
-        std::getenv(ServerHostEnvironmentVariableName);
-    auto host = env_server_host_val ? env_server_host_val : DefaultServerHost;
-    const char* env_port_val = std::getenv(PortEnvironmentVariableName);
-    auto port =
-        env_port_val ? std::strtoul(env_port_val, nullptr, 10) : DefaultPort;
+    auto server = options["server-host"].as<std::string>();
+    auto port = options["port"].as<std::uint16_t>();
 
     auto transport =
         std::make_shared<can::sim::transport::socket::SocketTransport<
-            freertos_synchronization::FreeRTOSCriticalSection>>(host, port);
+            freertos_synchronization::FreeRTOSCriticalSection>>(server, port);
 
 #endif
     return transport;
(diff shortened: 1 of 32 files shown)


gripper/simulator/main.cpp
```diff
@@ -18,33 +25,65 @@ void signal_handler(int signum) {
...
 
     LOG_INIT("GRIPPER", []() -> const char* {
         return pcTaskGetName(xTaskGetCurrentTaskHandle());
     });
-
+    auto options = handle_options(argc, argv);
+    auto sim_eeprom = std::make_shared<eeprom::simulator::EEProm>(options);
+    auto i2c_device_map = i2c::hardware::SimI2C::DeviceMap{
+        {sim_eeprom->get_address(), *sim_eeprom}};
+    auto i2c3 = std::make_shared<i2c::hardware::SimI2C>(i2c_device_map);
```
Laura-Danielle, line 77: what is the benefit of making these shared pointers here?

include/can/simlib/socketcan_transport.hpp
```diff
@@ -10,6 +10,7 @@
 #include <unistd.h>
 
 #include <cstring>
```
Laura-Danielle, line 12: do you still need to include the cstring lib?

include/pipettes/core/pipette_info.hpp
```diff
@@ -15,29 +15,22 @@ namespace pipette_info {
...
-// A full Serial number looks like P1KSV20201907243 and contains the name, model
-// and individual data code [name]P1K [model]SV [data_code] 20201907243
+// These defines are u    sed to help parse pipette serials, which are stored in
+// a structured binary +     ascii format:
+// +-----+--------+------+
+// |0x00 |name    |binary|
+// |0x01 |        |      |
+// +-----+--------+------+
+// |0x02 |model   |binary|and individual data code [name]P1K [model]SV
+// [data_code] 20201907243 |0x03 |        |      |
```
Laura-Danielle, line 25: The table formatting is a little hard to read here.

pipettes/simulator/main.cpp
```diff
@@ -131,16 +119,74 @@ auto initialize_motor_tasks(
...
     LOG_INIT(PipetteTypeString[PIPETTE_TYPE], []() -> const char* {
         return pcTaskGetName(xTaskGetCurrentTaskHandle());
     });
-
-    central_tasks::start_tasks(can_bus_1, node_from_env(std::getenv("MOUNT")));
-    peripheral_tasks::start_tasks(i2c3_comms, i2c1_comms, spi_comms);
-    initialize_motor_tasks(node_from_env(std::getenv("MOUNT")),
-                           motor_config.driver_configs, gear_motion_control);
+    auto options = handle_options(argc, argv);
+    auto hdcsensor = std::make_shared<hdc3020_simulator::HDC3020>();
```
Laura-Danielle, line 167: Same here -- don't understand the benefit of these being shared pointers

Laura-Danielle (approved):
Had a few questions/comments. In general looks good to me.

include/eeprom/simulation/eeprom.hpp
```diff
@@ -67,10 +79,154 @@ class EEProm : public I2CDeviceBase,
...
+            path.make_preferred();
+            LOG("Backing up eeprom with %s", path.c_str());
+            return std::fopen(path.c_str(), "a+b");
+        }
+        static auto get_prepped_backing_file(std::string pathstr)
+            -> std::FILE* {
+            auto file = ((pathstr == TEMPFILE_KEY) ? get_temp_file()
+                                                   : get_backing_file(pathstr));
+            if (!file) {
+                fprintf(stderr, "Could not open backing file %s: %d %s\n",
```
fsinapi, line 117: This is the only file that writes to `stderr` as far as I can tell, would it make more sense to consolidate all message in the same log output?

can/simlib/transport.cpp
```diff
@@ -1,44 +1,75 @@
...
+    const boost::program_options::variables_map& options)
     -> std::shared_ptr<can::sim::transport::BusTransportBase> {
 #ifdef USE_SOCKETCAN
-    auto constexpr ChannelEnvironmentVariableName = "CAN_CHANNEL";
-    auto constexpr DefaultChannel = "vcan0";
-
-    const char* env_channel_val = std::getenv(ChannelEnvironmentVariableName);
-    auto channel = env_channel_val ? env_channel_val : DefaultChannel;
-    auto transport =
+    auto channel = options["can-channel"].as<std::string>() auto transport =
```
fsinapi, line 62: Looks like a missing semicolon here?

fsinapi (approved):
LGTM, I really like this approach for the environment variables.

can/simlib/transport.cpp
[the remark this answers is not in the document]
sfoster1: yeah wait what

gantry/simulator/interfaces.cpp
```diff
@@ -76,10 +80,34 @@ static motor_handler::MotorInterruptHandler motor_interrupt(
...
+    }
+    po::store(po::parse_environment(envdesc, can_arg_xform), options);
+    po::notify(options);
+}
+
+std::shared_ptr<can::bus::CanBus> canbus;
+
+auto interfaces::get_can_bus() -> can::bus::CanBus& {
+    canbus.reset(
+        new can::sim::bus::SimCANBus(can::sim::transport::create(options)));
```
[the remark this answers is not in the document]
sfoster1: the fundamental problem is that the interfaces design really isn't suitable for this. we need a different interface for simulators that lets us pass created objects around rather than relying them being global.

if we keep this design, which we do in this pr, we should initialize our canbus in `initialize` and use it here. that means it needs a global home. that global home in this case is a `shared_ptr`.

Given that it's a shared ptr, making it valid means that we need to `new` up an object and hand control over - so now the shared ptr is in charge of deleting it.

we could also use an `optional<SimCANBus>` i think, but we have a bunch of shared ptrs everywhere so what's one more

gripper/simulator/main.cpp
sfoster1: Before this PR, all these objects could be statically created as globals. Now, some of them depend on command line options, so they need to be created in `main`. We could create them on the stack but that's usually bad practice when you're writing a general purpose program, so let's put them on the heap. Then, if we're putting it on the heap, we should really have it in a smart pointer. So, shared ptrs

include/pipettes/core/pipette_info.hpp
sfoster1: oh this got screwed up i think by auto formatting, dang

include/can/simlib/socketcan_transport.hpp
sfoster1: yeah the socket setup uses `strcpy`

gantry/simulator/interfaces.cpp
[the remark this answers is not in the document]
Laura-Danielle: OK now I understand why we've got smart pointers everywhere now. Makes sense!

can/simlib/transport.cpp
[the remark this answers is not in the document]
sfoster1: ok i'm going to fix this but it's not really something i can compile test because i don't have socketcan

caila-marashaj (approved):

include/eeprom/simulation/eeprom.hpp
sfoster1: No, I don't think so - this really doesn't want to be captured because it's not part of the normal logging. This would indicate something going wrong with the simulator itself.

Laura-Danielle (approved):
changes lgtm

include/eeprom/simulation/eeprom.hpp
```diff
@@ -67,10 +79,164 @@ class EEProm : public I2CDeviceBase,
...
+                            "EEPROM read of %ld bytes from %d on file %p read "
+                            "only %ld hit error: %d %s\n",
+                            size, address, backing, read_count, errno,
+                            strerror(errno));
+                    std::abort();
+                }
+            }
+        }
+
+        auto write(const uint8_t* writebuf, uint16_t address, size_t size)
```
ryanthecoder, line 206: Small thing, right now we don't attempt to catch when writing over a page boundry on the older chips which causes it to wrap to the same page, not the next page.
in include/eeprom/core/task.hpp we have:
if (hw_iface.get_eeprom_chip_type() ==
                hardware_iface::EEPromChipType::MICROCHIP_24AA02T &&
            ((m.memory_address % 8) + m.length) > 8) {
            LOG("Warning: write request will overrun page");
        }
        
If you want the simulator to be exact to that hardware when the hardware is the MICROCHIP_24AA02T (which is the current default) you'd have to catch and reproduce that behavior. Or we can edit task.hpp to actually handle these writes across pages. This shouldn't happen if you're using one of my accessor classes but if someone was trying to directly control the eeprom from Can messages or something it could happen

sfoster1: Oh interesting. That definitely might happen here (though I'm not that worried about the older chips since we're gonna stop using them) but i think we'd want to put that in the main sim eeprom class, not the backing storage.

PR #438, opened 2022-08-18
refactor(pipettes): add delay on boot to detect the correct id

Laura-Danielle:
## Overview

The right mount has a much higher goal tool id voltage than the left mount. When the pipette is hot-plugged into the robot, it takes close to 400ms for the tool id line to settle into the expected voltage.

To get around this, we should delay when we actually sample the tool id line on the pipette to determine what mount it's connected to.

### DIFF
--- a/pipettes/firmware/main.cpp
+++ b/pipettes/firmware/main.cpp
@@ -169,7 +169,7 @@ auto main() -> int {
     adc_init();
     initialize_enc(PIPETTE_TYPE);
 
-    delay_start(400);
+    delay_start(500);
 
     auto id = pipette_mounts::detect_id();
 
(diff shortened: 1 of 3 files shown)


pipettes/firmware/main.cpp
```diff
@@ -33,10 +33,9 @@
 #pragma GCC diagnostic push
 // NOLINTNEXTLINE(clang-diagnostic-unknown-warning-option)
 #pragma GCC diagnostic ignored "-Wvolatile"
-//#include "motor_encoder_hardware.h"
 #include "motor_hardware.h"
-//#include "motor_timer_hardware.h"
 #include "pipettes/firmware/i2c_setup.h"
+#include "stm32g4xx_hal.h"
```
sfoster1, line 38: shouldn't include this - we don't need it now that we moved `delay_start` to c

sfoster1 (approved):
🍣 (but let's remove that include)

ahiuchingau (approved):
🙌

PR #439, opened 2022-08-18
refactor(head): change the instrument values reported by the ADC

Laura-Danielle:
## Overview
Awhile ago, we no longer wanted to try and determine the instrument type based on the ADC because it
was too variable. Now, we are cutting down what values are valid and also changing the ranges
provided in [this confluence doc](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/940834928/OT-3+Tool+ID+Implementation).

### DIFF
--- a/common/core/tool_detection.cpp
+++ b/common/core/tool_detection.cpp
@@ -9,7 +9,7 @@ constexpr static auto pipette_chan_z_bounds =
     ToolCheckBounds{.upper = 580, .lower = 540};
 
 constexpr static auto pipette_chan_a_bounds =
-    ToolCheckBounds{.upper = 2950, .lower = 3050};
+    ToolCheckBounds{.upper = 3050, .lower = 2900};
 
 constexpr static auto nothing_connected_z_bounds =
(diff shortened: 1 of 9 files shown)


common/core/tool_detection.cpp
```diff
@@ -5,70 +5,62 @@
 
 using namespace tool_detection;
 
-constexpr static auto pipette_384_chan_z_bounds =
-    ToolCheckBounds{.upper = 1883, .lower = 1850};
+constexpr static auto pipette_chan_z_bounds =
+    ToolCheckBounds{.upper = 580, .lower = 540};
```
sfoster1, line 9: might want to rename this to like `pipette_z_bounds`

sfoster1 (changes requested):
Let's widen these ranges quite a bit. The hardware changes mean we have the space, and we should take advantage of it to cover small amounts of variability. From my reading of the spec, we could do something like 0-0.2V for nothing, 0.201-1.5V for Z, 1.501-3V for A, and 3V-(rail) for cable failure.

sfoster1 (approved):
Nice

PR #440, opened 2022-08-19
fix(pipettes): correct adc channel usage for evt

sfoster1:
EVT pipettes use pin B0/adc channel 15 for their tool detect lines.

Closes RLIQ-96


Laura-Danielle (approved):
lgtm

PR #441, opened 2022-08-19
fix(pipettes): rev1: bring up M24128 eeprom

sfoster1:
The rev1 pipettes use an M24128 eeprom. We added code to handle this ahead of time, but it had a few issues that this PR fixes.
- rev1 pipettes use M24128 eeprom, not the older kind. Explicitly set them to do so.
- M24128 eeproms have a time period after they start a write during which they won't respond on I2C. We can fix this by retrying I2C messages when we get NAKs
   - An I2C NAK indicates that a device won't respond to a message. This can
     be for basically any reason, including that it doesn't recognize it, but
     one reason is that the device isn't ready yet. For instance, the M24128
     eeprom will NAK all incoming messages during its write delay. That means
     that the NAK can be a good way to get low-latency estimations of when a
     device is ready to transmit. By altering the i2c hardware interface code so that it retries a
     transaction if it failed because it was NAK'd, we can use that
     retryability to handle the write delay without an explicit sleep.
- Use 10-byte buffers instead of 9 because an M24128 address is 2 bytes, so we were cutting off one byte of an
   8-byte payload by using a 9-byte buffer

With this PR, we can write and read EVT pipette serial numbers.

Closes RLIQ-97
Closes RLIQ-98
Closes RLIQ-99

### DIFF
--- a/i2c/firmware/i2c.c
+++ b/i2c/firmware/i2c.c
@@ -1,4 +1,5 @@
 #include "platform_specific_hal_conf.h"
+#include "platform_specific_hal.h"
 #include "i2c/firmware/i2c.h"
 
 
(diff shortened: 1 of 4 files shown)


i2c/firmware/i2c.c
```diff
@@ -7,14 +8,40 @@
...
 {
-    return HAL_I2C_Master_Transmit(handle,
-                            DevAddress, data, size, timeout) == HAL_OK;
+    I2C_HandleTypeDef* i2c_handle = (I2C_HandleTypeDef*)handle;
+    uint32_t tickstart = HAL_GetTick();
+    HAL_StatusTypeDef tx_result = HAL_OK;
+    do {
+        tx_result = HAL_I2C_Master_Transmit(i2c_handle,
+                            DevAddress, data, size, timeout);
+        if (__HAL_I2C_GET_FLAG(i2c_handle, I2C_FLAG_AF)) {
```
Laura-Danielle, line 17: Doesn't the `master_transmit` function set the busy flag already for you?

Laura-Danielle (approved):
One question about the NACK, but looks good in general.

i2c/firmware/i2c.c
sfoster1: this is the NACK (Ack Failure) flag not the busy flag

PR #443, opened 2022-08-24
feature(gripper): add precise linear motion control for brushed motors

ryanthecoder:
Will need to use RET-1097-gripper-evt-basic-position-control branch in the mono-repo to send the new can messages


include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -50,21 +58,72 @@ class BrushedMotorInterruptHandler {
...
+            current_control_pwm = 0;
+        } else {
+            if (move_delta < 0) {
+                hardware.positive_direction();
+            } else if (move_delta > 0) {
+                hardware.negative_direction();
+            }
+            // the compute the pid with the abs of move_delta because the pwm
+            // value is always positive regardless of moving forward or backward
+            double pid_output = hardware.update_control(std::abs(move_delta));
```
fsinapi, line 80: I'm pretty sure that calculating the PID output from an absolute value results in some fundamental changes to how the controller functions. In particular, the integral term will _only_ ever become more positive because every error value is positive. I would generally expect a system like this to calculate the correction value from a signed error value, and then apply the positive/negative direction and convert to an unsigned PWM value after the fact.

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -50,21 +58,72 @@ class BrushedMotorInterruptHandler {
         return queue.has_message_isr();
     }
 
+    // upon advice from hardware, 0.01mm is a good limit for percision
```
sfoster1, line 61:
```suggestion
    // upon advice from hardware, 0.01mm is a good limit for precision
```

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -95,10 +154,26 @@ class BrushedMotorInterruptHandler {
...
+        switch (buffered_move.stop_condition) {
+            case MoveStopCondition::limit_switch:
+                hardware.ungrip();
+                break;
+            case MoveStopCondition::encoder_position:
+                if (hardware.get_encoder_pulses() ==
+                    buffered_move.encoder_position) {
+                    LOG("Attempting to add a gripper move to our current "
+                        "position");
+                    finish_current_move(AckMessageId::position_error);
```
sfoster1, line 168: why would this be an error rather than an immediate completion?

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -50,21 +58,72 @@ class BrushedMotorInterruptHandler {
...
+
+    auto controlled_move_to(int32_t encoder_position) -> int32_t {
+        int32_t move_delta = hardware.get_encoder_pulses() - encoder_position;
+        uint32_t old_control_pwm = current_control_pwm;
+        // pass through early if we're already within acceptable position
+        if (std::abs(move_delta) < acceptable_position_error()) {
+            current_control_pwm = 0;
+        } else {
+            if (move_delta < 0) {
+                hardware.positive_direction();
```
sfoster1, line 74: let's either always use `grip`/`ungrip` or never use `grip`/`ungrip`

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -50,21 +58,72 @@ class BrushedMotorInterruptHandler {
         return queue.has_message_isr();
     }
 
+    // upon advice from hardware, 0.01mm is a good limit for percision
+    auto acceptable_position_error() -> int32_t {
+        return int32_t(gear_conf.get_encoder_pulses_per_mm() * 0.01);
```
sfoster1, line 63: this is the only place we ever use this; can we make `acceptable_position_error` be a value parameter to this class's ctor instead and save a float calculation?

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -50,21 +58,72 @@ class BrushedMotorInterruptHandler {
...
-            if (buffered_move.stop_condition ==
-                    MoveStopCondition::limit_switch &&
-                limit_switch_triggered()) {
-                homing_stopped();
-            } else {
-                if (is_sensing() && is_idle) {
-                    finish_current_move(
-                        AckMessageId::complete_without_condition);
-                }
+            switch (buffered_move.stop_condition) {
```
sfoster1, line 97: this is getting a little complex; maybe we should split it into like an `execute_active_move()` subfunction and a `execute_idle()` subfunction?

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -50,21 +58,72 @@ class BrushedMotorInterruptHandler {
...
+                        finish_current_move(
+                            AckMessageId::complete_without_condition);
+                    }
+                    break;
+                case MoveStopCondition::cap_sensor:
+                    // TODO write cap sensor move code
+                    break;
             }
+        } else if (holding && std::abs(hardware.get_encoder_pulses() -
+                                       hold_encoder_position) >
```
sfoster1, line 124: the `if( (...) abs(pulses() - position) < acceptable)` is the first thing `controlled_move_to` does, right? can we cut it from this `if` and rely on the function call?

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -140,6 +222,10 @@ class BrushedMotorInterruptHandler {
     StatusClient& status_queue_client;
     motor_hardware::BrushedMotorHardwareIface& hardware;
     brushed_motor_driver::BrushedMotorDriverIface& driver_hardware;
+    lms::LinearMotionSystemConfig<lms::GearBoxConfig>& gear_conf;
     BrushedMove buffered_move = BrushedMove{};
+    std::atomic<bool> holding = false;
```
sfoster1, line 227: does this need to be atomic? it's only accessed inside the ISR, right?

sfoster1, line 227: also, does it make more sense for this to be an enum of something like
```
enum class State {
   FORCE_CONTROLLING,
   POSITION_CONTROLLING,
   IDLE 
};
```
rather than a `holding` bool and an implicit "don't disable the motor when you finish a force control move"?

include/can/core/messages.hpp
```diff
@@ -720,6 +720,40 @@ struct SetBrushedMotorPwmRequest
         -> bool = default;
 };
 
+struct AddBrushedLinearMoveRequest
+    : BaseMessage<MessageId::add_brushed_linear_move_request> {
+    uint8_t group_id;
+    uint8_t seq_id;
+    brushed_timer_ticks duration;
```
sfoster1, line 727: why do we need `duration` and `duty_cycle`? those values aren't used, right?

sfoster1 (changes requested):
Looks really good overall, would love a demo. Some details I think we could improve though.

include/can/core/messages.hpp
ryanthecoder: duration currently isn't used by any of the gripper moves, we have a ticket in our queue to use duration to trigger a timeout so moves don't get stuck forever.
and on the monorepo side all the gripper messages have the same payload type for consistency there, so duty cycle is a hold over, I may use this in the future as a duty_cycle MAX value

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -19,6 +21,17 @@ using namespace motor_messages;
...
  */
 
+enum class ControlState {
+    FORCE_CONTROLLING,
+    POSITION_CONTROLLING,
+    IDLE,
+    ACTIVE
+};
+
+// TODO Investigate this value, I'm pretty sure that the timer actually gets
```
ahiuchingau, line 31: Oh you're absolutely right about this. I forgot to factor in the period!

We probably want to shorten this hold-off period a bit.

gripper/firmware/interfaces_grip_motor.cpp
```diff
@@ -31,10 +31,19 @@ struct motor_hardware::BrushedHardwareConfig brushed_motor_pins {
...
+            .pin = GPIO_PIN_2,
+            .active_setting = GPIO_PIN_SET},
+    .encoder_interrupt_freq =
+        double(GRIPPER_JAW_PWM_FREQ_HZ) / double(GRIPPER_JAW_PWM_WIDTH),
+    // the expected behavior with these pid values is that the motor runs at
+        // full power until it's about 2mm away and then it slows down on that
+        // approach. final values as the error delta approachs 0.01mm is ~7
+        // which
+        // is the floor of pwm values that still move the gripper
+        .pid_kp = 0.0008, .pid_ki = 0.0000125, .pid_kd = 0.000015,
```
ahiuchingau, line 46: Could you add the formula for getting these pid values to the comment here?

ahiuchingau (approved):
Very cool! Can't wait to see the behavior in real life.

include/motor-control/firmware/brushed_motor/brushed_motor_hardware.hpp
```diff
@@ -18,20 +20,29 @@ struct BrushedHardwareConfig {
...
     ~BrushedMotorHardware() final = default;
     BrushedMotorHardware() = delete;
     BrushedMotorHardware(const BrushedHardwareConfig& config,
                          void* encoder_handle)
-        : pins(config), enc_handle(encoder_handle) {}
-    BrushedMotorHardware(const BrushedMotorHardware&) = default;
+        : pins(config),
+          enc_handle(encoder_handle),
+          controller_loop{config.pid_kp, config.pid_ki, config.pid_kd,
+                          1 / config.encoder_interrupt_freq} {}
```
sfoster1, line 40:
```suggestion
                          1.f / config.encoder_interrupt_freq} {}
```

just to make it explicit

sfoster1 (changes requested):
Sounds good to me on the message elements front. Couple more little things but this is looking really good.

gripper/firmware/interfaces_grip_motor.cpp
```diff
@@ -31,10 +34,34 @@ struct motor_hardware::BrushedHardwareConfig brushed_motor_pins {
...
+            .pin = GPIO_PIN_2,
+            .active_setting = GPIO_PIN_SET},
+    .encoder_interrupt_freq =
+        double(GRIPPER_JAW_PWM_FREQ_HZ) / double(GRIPPER_JAW_PWM_WIDTH),
+
+    /* the expected behavior with these pid values is that the motor runs at
+     * full power until it's about 2mm away and then it slows down on that
+     * approach. final values as the error delta approachs 0.01mm is ~7
+     * which is the floor of pwm values that still move the gripper
+     * this 7 is from the windup limits, but over very short moves( <
```
Laura-Danielle, line 49: Is this an incomplete thought?

gripper/firmware/motor_encoder_hardware.h
```diff
@@ -6,10 +6,22 @@
 #include "platform_specific_hal_conf.h"
 #include "stm32g4xx_it.h"
 
+// The frequency of one full PWM cycle
+#define GRIPPER_JAW_PWM_FREQ_HZ (32000UL)
+// the number of selectable points in the PWM
+#define GRIPPER_JAW_PWM_WIDTH (100UL)
+// the frequency at which the timer should count so that it
+// does a full PWM cycle in the time specified by GRIPPER_JAW_PWM_FREQ_HZ
+#define GRIPPER_JAW_TIMER_FREQ (GRIPPER_JAW_PWM_FREQ_HZ * GRIPPER_JAW_PWM_WIDTH)
```
Laura-Danielle, line 15: Should we make a define for the encoder interrupt frequency as well -- since we're already doing one for the gripper timer frequency?

include/can/core/messages.hpp
```diff
@@ -720,6 +720,40 @@ struct SetBrushedMotorPwmRequest
         -> bool = default;
 };
 
+struct AddBrushedLinearMoveRequest
+    : BaseMessage<MessageId::add_brushed_linear_move_request> {
+    uint8_t group_id;
```
Laura-Danielle, line 725: What's the difference between this and a "grip" request? Can't we just combine the two CAN messages since we're really just enhancing the functionality of gripping?

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -33,11 +49,17 @@ class BrushedMotorInterruptHandler {
...
+        brushed_motor_driver::BrushedMotorDriverIface& driver_iface,
+        lms::LinearMotionSystemConfig<lms::GearBoxConfig>& gearbox_config)
         : queue(incoming_queue),
           status_queue_client(outgoing_queue),
           hardware(hardware_iface),
-          driver_hardware(driver_iface) {}
+          driver_hardware(driver_iface),
+          gear_conf(gearbox_config) {
+        acceptable_position_error =
+            int32_t(gear_conf.get_encoder_pulses_per_mm() *
```
Laura-Danielle, line 60: Should this also be absolute value since we're comparing it below to the abs of the `move_delta`?

Laura-Danielle (commented):
Some small comments that definitely don't block merging this PR.

gripper/firmware/interfaces_grip_motor.cpp
sfoster1: definitely want to fix this before merging

include/can/core/messages.hpp
sfoster1: the grip request specifies force control, and this specifies position control. they carry similar bits for now, but that won't necessarily always be true - they mean different things

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -19,9 +21,23 @@ using namespace motor_messages;
...
+
+// TODO Investigate this value, I'm pretty sure that the timer actually gets
+// called every 3ms because the clock is 32K hz and the period is set to 100
+// which will call the interupt every 3.125 ish ms and therefor the hold off
+// will be 100 ms
 static constexpr uint32_t HOLDOFF_TICKS =
     32;  // hold off for 1 ms (with a 32k Hz timer)
 
+// upon advice from hardware, 0.01mm is a good limit for precision
+static constexpr double ACCEPTABLE_DISTANCE_TOLLERANCE_MM = 0.01;
```
sfoster1, line 39:
```suggestion
static constexpr double ACCEPTABLE_DISTANCE_TOLERANCE_MM = 0.01;
```

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -33,11 +49,17 @@ class BrushedMotorInterruptHandler {
...
+        lms::LinearMotionSystemConfig<lms::GearBoxConfig>& gearbox_config)
         : queue(incoming_queue),
           status_queue_client(outgoing_queue),
           hardware(hardware_iface),
-          driver_hardware(driver_iface) {}
+          driver_hardware(driver_iface),
+          gear_conf(gearbox_config) {
+        acceptable_position_error =
+            int32_t(gear_conf.get_encoder_pulses_per_mm() *
+                    ACCEPTABLE_DISTANCE_TOLLERANCE_MM);
```
sfoster1, line 61:
```suggestion
                    ACCEPTABLE_DISTANCE_TOLERANCE_MM);
```

sfoster1 (approved):
This is real nice! IMO we should fix that comment @Laura-Danielle noted and the typo in tolerance but looks good!

PR #444, opened 2022-08-26
fix(simulators): parse environment properly

sfoster1:
The options specified through environment variables need to have the
same names as the command line options. You then define what variables
you support in the name transformation lambdas.

With this change we should now properly support environment variable
specification of options.


DerekMaggio (approved):
Validated that usage in both ot3-firmware and opentrons-emulation now accept env variables and does not blow up.

PR #446, opened 2022-09-12
Fix(gripper): reconfigure configure encoder inputs to solve MCU crash

ryanthecoder:
https://opentrons.atlassian.net/browse/RLAB-98
Found a bug that when the motor is driven at high speed, the number of encoder pulses that get sent to reset the slave speed timer is too great. While everything works fine in release mode, in debug mode this overruns the mcu and causes a crash.
Solution is to prescale the slave timer to 4, this reduces the reset from 4x per encoder tick (1 per each rising/falling edge) to 1 per complete encoder tick

### DIFF
--- a/gripper/firmware/motor_encoder_hardware_proto.c
+++ b/gripper/firmware/motor_encoder_hardware_proto.c
@@ -35,19 +35,15 @@ void Encoder_GPIO_Init(void) {
     GPIO_InitStruct.Alternate = GPIO_AF1_TIM2;
     HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
 
-    /*Configure GPIO pin : PA5 */
+    /* There is no need to configure GPIO pin : PA5
+     * for clarity I have left this here commented, as it describes the defualt
+     * state for this pin at startup
     GPIO_InitStruct.Pin = GPIO_PIN_5;
-    GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
+    GPIO_InitStruct.Mode = GPIO_MODE_INPUT;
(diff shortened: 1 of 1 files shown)


gripper/firmware/motor_encoder_hardware.c
```diff
@@ -127,7 +123,7 @@ void TIM4_EncoderGSpeed_Init(void) {
     /* Configure TIM4 in Slave mode */
     sSlaveConfig.SlaveMode = TIM_SLAVEMODE_RESET;
     sSlaveConfig.InputTrigger =
-        TIM_TS_ITR1;  // input trigger uses encoder clock
+        TIM_TS_TI1FP1;  // input trigger uses filtered timer input
```
ahiuchingau, line 126: What does this do?

sfoster1 (approved):
Looks good, but please put a more in-depth description of the problem and solution in the PR description and the eventual squashed commit message.

ahiuchingau (approved):
Got a question, but looks great to me if it fixes the issue we saw on the robot.

gripper/firmware/motor_encoder_hardware.c
ryanthecoder: Not what I thought It did apparently, this makes the crash stop but it is the wrong trigger and breaks encoder_g speed break, gonna re-do and update the PR

ahiuchingau:
> Wait so does this affect the speed that is sensed?

I think the math I did previously was just wrong. What Ryan did here fixes it and also gave the correct formula to let us actually get the right velocity.

ahiuchingau (approved):

sfoster1 (approved):
All right, looks good to me!

vegano1 (approved):

PR #447, opened 2022-09-12
chore(gripper): firmware versioning

ahiuchingau:
This PR allows us to separate the gripper versions, `proto` and `rev1`.


Laura-Danielle (approved):
I think you don't need to switch core files based on revision, and for the simulator you should only need to import one of the core files.

sfoster1 (approved):
Looks good to me!

PR #449, opened 2022-09-21
feat(motor-control): add stealthchop config to tmc2160

ahiuchingau:
This PR allows us to configure stealth chop for the TMC2160 driver. 

We don't currently use it but I figured we might need it after some more testing with the EEs & testing team.


Laura-Danielle (approved):
Steallltthh CHOP

sfoster1 (approved):

PR #452, opened 2022-10-06
refactor(gantry): change the direction of the X motor

Laura-Danielle:
The X motor has been switched to the opposite side of the gantry/pulley and needs to have its direction switched.


sfoster1 (approved):

DerekMaggio (approved):
I can confidently approve this PR. :)

PR #453, opened 2022-10-06
refactor(head): modify the gear ratio for the EVT gantry

Laura-Danielle:
The new ZA motors have a gear train at the top of the linear motors now. It's the same system on both the left and right motors.


Laura-Danielle:
> Obviously has to happen to make it work, but do you know what this does to the max encodable velocity?

Back of the napkin calculation is that the max encodable integer would end up being ~ 1.6 mil since the steps per mm is 25% more now, but that could be super off.

PR #454, opened 2022-10-07
fix(motor-control): Z motors current and encoder fixes

ryanthecoder:
This fixes the math for the tm21* driver set current, fixes the motor config for the gripper z and fixes an encoder issue with the gripper jaw when gripping


gripper/firmware/interfaces_z_motor.cpp
```diff
@@ -68,8 +68,8 @@ static motor_hardware::MotorHardware motor_hardware_iface(motor_pins, &htim7,
  */
 static tmc2130::configs::TMC2130DriverConfig MotorDriverConfigurations{
     .registers = {.gconfig = {.en_pwm_mode = 0x1},
-                  .ihold_irun = {.hold_current = 0x0,
-                                 .run_current = 0x19,
+                  .ihold_irun = {.hold_current = 0xA,
```
sfoster1, line 71: can we drop in some comments for what this ends up commanding in amps

gripper/firmware/interfaces_z_motor.cpp
```diff
@@ -80,10 +80,13 @@ static tmc2130::configs::TMC2130DriverConfig MotorDriverConfigurations{
                                .tbl = 0x2,
                                .mres = 0x4},
                   .coolconf = {.sgt = 0x6},
-                  .pwmconf = {.freewheel = 0x2}},
+                  .pwmconf = {.pwm_ampl = 0x80,
```
sfoster1, line 83: can we add what this does or a quick description of the issue it fixes to the PR description and eventual commit body

sfoster1 (approved):
Looks good but let's add a couple more descriptions of things

PR #457, opened 2022-10-10
feature(can): have all messages respond with a message index in their response or an empty ack with the message index

ryanthecoder:
RET-1247
Paired with https://github.com/Opentrons/opentrons/pull/11527

### DIFF
--- a/bootloader/tests/test_messages.cpp
+++ b/bootloader/tests/test_messages.cpp
@@ -54,7 +54,7 @@ SCENARIO("empty message errors") {
 
 SCENARIO("update data") {
     GIVEN("a message with a data byte count of 0") {
-        auto arr = std::array<uint8_t, 64>{
+        auto arr = std::array<uint8_t, 60>{
             // Message Index
             0xde, 0xad, 0xbe, 0xef,
             // Address
@@ -66,9 +66,10 @@ SCENARIO("update data") {
             // Data
             0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
             0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-            0, 0, 0, 0, 0, 0,
+            0, 0,
             // Checksum.
             0xfc, 0x9a};
+        
         WHEN("parsed") {
             UpdateData result;
             auto error = parse_update_data(arr.data(), arr.size(), &result);
@@ -85,13 +86,13 @@ SCENARIO("update data") {
     }
 
     GIVEN("a message with full data byte count") {
-        auto arr = std::array<uint8_t, 64>{
+        auto arr = std::array<uint8_t, 60>{
             // Message Index
             0xde, 0xad, 0xbe, 0xef,
             // Address
             0x89, 0xab, 0xcd, 0xef,
             // Size
-            52,
+            48,
             // Reserved
             0x0,
             // Data
@@ -99,9 +100,9 @@ SCENARIO("update data") {
             0x17, 0x20, 0x21, 0x22, 0x23, 0x24, 0x25, 0x26, 0x27, 0x30, 0x31,
             0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x40, 0x41, 0x42, 0x43, 0x44,
             0x45, 0x46, 0x47, 0x50, 0x51, 0x52, 0x53, 0x54, 0x55, 0x56, 0x57,
-            0x60, 0x61, 0x62, 0x63,
             // Checksum.
-            0xEF, 0xF6};
+            0xF1, 0x80};
+        
         WHEN("parsed") {
             UpdateData result;
             auto error = parse_update_data(arr.data(), arr.size(), &result);
@@ -109,10 +110,10 @@ SCENARIO("update data") {
             THEN("its fields are populated") {
                 REQUIRE(result.message_index == 0xdeadbeef);
                 REQUIRE(result.address == 0x89abcdef);
-                REQUIRE(result.num_bytes == 52);
+                REQUIRE(result.num_bytes == 48);
                 REQUIRE(result.reserved == 0);
                 REQUIRE(result.data == arr.data() + 10);
-                REQUIRE(result.checksum == 0xEFF6);
+                REQUIRE(result.checksum == 0xF180);
             }
         }
     }
@@ -151,19 +152,19 @@ SCENARIO("update data errors") {
     }
 
     GIVEN("a message with invalid data byte count") {
-        auto arr = std::array<uint8_t, 64>{
+        auto arr = std::array<uint8_t, 60>{
             // Message Index
             0xde, 0xad, 0xbe, 0xef,
             // Address
             0xa, 0xb, 0xc, 0xd,
             // Size
-            53,
+            52,
             // Reserved
             0x0,
             // Data
             0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
             0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-            0, 0, 0, 0, 0, 0,
+            0, 0,
             // Checksum.
             0xFC, 0x9A};
         WHEN("parsed") {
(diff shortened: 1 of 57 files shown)


include/bootloader/core/messages.h
```diff
@@ -20,21 +21,34 @@ typedef struct {
 } UpdateData;
 
 #define UPDATE_DATA_MESSAGE_SIZE    64
-#define UPDATE_DATA_MAX_BYTE_COUNT  56
+#define UPDATE_DATA_MAX_BYTE_COUNT  52
```
sfoster1, line 24: can we make this `#define UPDATE_DATA_MAX_BYTE_COUNT (UPDATE_DATA_MESSAGE_SIZE - sizeof(UpdateData))`?

include/bootloader/core/messages.h
```diff
@@ -20,21 +21,34 @@ typedef struct {
...
  * Contents of the firmware update complete message.
  */
 typedef struct {
+    uint32_t message_index;
     uint32_t num_messages;
     uint32_t crc32;
 } UpdateComplete;
 
-#define UPDATE_COMPLETE_MESSAGE_SIZE    8
+#define UPDATE_COMPLETE_MESSAGE_SIZE    12
```
sfoster1, line 35: `#define UPDATE_COMPLETE_MESSAGE_SIZE (sizeof(UpdateComplete))`?

include/can/core/message_handlers/system.hpp
```diff
@@ -56,24 +56,25 @@ class SystemMessageHandler {
...
         writer.send_can_message(can::ids::NodeId::host, response);
     }
 
     void visit(InitiateFirmwareUpdate &) { app_update_start(); }
 
-    void visit(FirmwareUpdateStatusRequest &) {
-        auto status_response =
-            FirmwareUpdateStatusResponse{.flags = app_update_flags()};
+    void visit(FirmwareUpdateStatusRequest &m) {
+        auto status_response = FirmwareUpdateStatusResponse{
```
sfoster1, line 67: a static member function that builds a response from a request could be helpful here, possibly even one that takes a mostly-premade response so you could use tagged-initialization still. Something like this:

```
struct FirmwareUpdateStatusResponse {
    static auto from_request(FirmwareUpdateStatusResponse premade. FirmwareUpdateStatusRequest request) -> FirmwareUpdateStatusResponse {
        premade.message_id = request.message_id;
        return premade; 
    }
};
```
which could then be called like
```
auto status_response = FirmwareUpdateStatusResponse::from_request({.flags = app_update_flags()}, m);
```
as long as there was a default for the message ids. That way we could establish a pattern of "please build this through the static method" and not have to remember to set the message id.

include/eeprom/core/task.hpp
```diff
@@ -187,6 +189,7 @@ class EEPromMessageHandler {
             m.memory_address, iter, (iter + hw_iface.get_eeprom_addr_bytes()));
 
         auto transaction = i2c::messages::Transaction{
+            .message_index = m.message_index,
```
Laura-Danielle, line 192: So these message indexes for internal messages will have a unique set of errors that the `send_to_can` message will eventually be able to interpret?

include/can/core/messages.hpp
```diff
@@ -41,19 +41,25 @@ struct BaseMessage {
  */
 template <MessageId MId>
 struct Empty : BaseMessage<MId> {
+    uint32_t message_index;
```
sfoster1, line 44: Should we put the `message_index` member in `BaseMessage`? Everything inherits from it anyway. We could then even put a `parse_message_id` and `serialize_message_id` in it.

include/head/core/tasks/presence_sensing_driver_task.hpp
```diff
@@ -81,6 +83,8 @@ class PresenceSensingDriverMessageHandler {
             can_client.send_can_message(
                 can::ids::NodeId::host,
                 can::messages::PushToolsDetectedNotification{
+                    // 0 message index to indicate async message
```
Laura-Danielle, line 86: why?

include/can/core/messages.hpp
```diff
@@ -41,19 +41,25 @@ struct BaseMessage {
...
-        return 0;
+    auto serialize(Output body, Limit limit) const -> uint8_t {
+        auto iter = bit_utils::int_to_bytes(message_index, body, limit);
+        return iter - body;
     }
 
     auto operator==(const Empty&) const -> bool = default;
 };
 
+using Acknowledgment = Empty<MessageId::acknowledgement>;
```
sfoster1, line 61: What if we make a 
```
struct EmptyResponse: public BaseMessage<MessageId::acknowledgement> {
    template <HasMessageId Request>
    static auto from_message(const Request& r) -> Acknowledgement {
        return Acknowledgement{.message_index = r.message_index};
    }
}
```
?

include/motor-control/core/tasks/brushed_move_group_task.hpp
```diff
@@ -75,17 +75,21 @@ class MoveGroupMessageHandler {
...
 
-    void handle(const can::messages::ClearAllMoveGroupsRequest&) {
+    void handle(const can::messages::ClearAllMoveGroupsRequest& m) {
         LOG("Received clear move groups request");
         for (auto& group : move_groups) {
             group.clear();
         }
+        can_client.send_can_message(
+            can::ids::NodeId::host,
+            can::messages::Acknowledgment{.message_index = m.message_index});
```
Laura-Danielle, line 92: Is it common practice to have error messages in `Acknowledgment` messages?

sfoster1 (changes requested):
The core functionality and testing look good but I think if we add a couple strategic helper functions for creating responses from requests and automatically copying the message index we'll make everything a lot simpler to do correctly.

include/bootloader/core/ids.h
```diff
@@ -59,6 +59,7 @@ typedef enum {
     can_messageid_gripper_grip_request = 0x42,
     can_messageid_gripper_home_request = 0x43,
     can_messageid_add_brushed_linear_move_request = 0x44,
+    can_messageid_acknowledgement = 0x50,
```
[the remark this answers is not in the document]
sfoster1: No, node ids and message ids go in different places in the arbitration id

sfoster1:
> Excited for the possibilities! I did have a few questions, and I'm wondering if we need the message index to be that large?

The absolute #1 do not mess this up thing you need with indexes like this is a guarantee that there is never a situation with two live messages with the same ID. this is an issue that has personally caused me great pain in the past. i'm always pro making indices bigger to avoid it.

having the index be large also means that it's generally safe to say "okay this bit is actually a flag now" without making the above situation worse.

include/bootloader/core/messages.h
ryanthecoder: unfortunately no because of C padding structs, unless we pack it the sizeof(updataData) is 32 instead of 12. Left a comment in the code

bootloader/tests/test_messages.cpp
```diff
@@ -118,18 +179,20 @@ SCENARIO("update data errors") {
...
+            52,
             // Reserved
             0x0,
             // Data
             0, 1, 2, 3, 4, 5, 6, 7, 0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16,
             0x17, 0x20, 0x21, 0x22, 0x23, 0x24, 0x25, 0x26, 0x27, 0x30, 0x31,
             0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x40, 0x41, 0x42, 0x43, 0x44,
             0x45, 0x46, 0x47, 0x50, 0x51, 0x52, 0x53, 0x54, 0x55, 0x56, 0x57,
-            0x60, 0x61, 0x62, 0x63, 0x64, 0x65, 0x66, 0x67,
+            0x60, 0x61, 0x62, 0x63,
```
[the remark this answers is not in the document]
ryanthecoder: the size of message data decreased when I added message index so this is the correct size, I also changed the size under the //size line above

include/can/core/messages.hpp
ryanthecoder: spent a few hours trying to get this to work, the parsing and serializing work but it messes up the inheirtance for the non-empty child structs. we would have to change the instantiation of substructs everwhere and add constructors to them all
it's because it makes the BaseMessage a non-aggregate type and we'd have to do all that to get brace initalization working
I'll leave what I got in a branch but I think its better the way it is
For info this is similar to whats going on https://stackoverflow.com/questions/64770166/why-i-can-not-use-designated-initalizers-with-structs-that-are-not-aggregates

include/head/core/tasks/presence_sensing_driver_task.hpp
ryanthecoder: When the firmware sends a message that wasn't the result of a request we need a way to indicate that, so message_index = 0 is how I decided to do that. the way I wrote the monorepo side of this, it will never send a request with a message index of 0

include/eeprom/core/task.hpp
ryanthecoder: Passing these into the i2c or spi transactions allows good responses to contain the message ind or we could generate an error message with the message ind

sfoster1 (approved):
I like it! Looks great!

ryanthecoder:
Going to replace with a different PR

PR #458, opened 2022-10-12
refactor(tmc2160): use the correct method of setting current

Laura-Danielle:
Erroneously, we thought that you needed to set the global scaler rather than the ihold/irun registers directly on the tmc2160. In actuality, you only adjust the global scaler to find a good default value. The EEs have calculated that ~167 is an acceptable value to get us to a max current of 1.5amps when irun is 31.

RET#1261


sfoster1 (approved):
Ohhh okay that makes sense

PR #459, opened 2022-10-13
Enable spread-cycle for Head 

andySigler:
Head should use spread-cycle by default, because it creates a more predictable output current, which we need during pick-up-tip. Downside is that it is louder than stealth mode, but that's an ok tradeoff.

Should follow #458 so it uses the global-scaler setting.

### DIFF
--- a/head/firmware/main_rev1.cpp
+++ b/head/firmware/main_rev1.cpp
@@ -137,19 +137,20 @@ struct motor_hardware::HardwareConfig pin_configurations_right {
 static tmc2160::configs::TMC2160DriverConfig motor_driver_configs_right{
     .registers =
         {
-            .gconfig = {.en_pwm_mode = 1},
+            .gconfig = {.en_pwm_mode = 0},
             .ihold_irun = {.hold_current = 16,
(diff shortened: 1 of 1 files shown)


sfoster1:
> Head should use spread-cycle by default, because it creates a more predictable output current, which we need during pick-up-tip. Downside is that it is louder than stealth mode, but that's an ok tradeoff.
> 
> Should follow #458 so it uses the global-scaler setting.

@andySigler if you retarget this on that pr it'll make the diff  easier to see and should automatically go target on main when merged

sfoster1 (changes requested):
This looks good but the configuration is getting extremely in depth. can we drop some comments in those code files about what it's doing

andySigler:
Ok I did that and now the diff looks crazy.  I'm hoping #458 can be merged and this will look less messy or I mucked it up somehow

sfoster1 (approved):
Looks good (make sure to catch up on the `mres` changes that got merged, I think that's the conflict?) so when the conflicts are fixed we're good

ahiuchingau (approved):

PR #460, opened 2022-10-13
change tools detected if nothing attached

caila-marashaj:
On [main](https://github.com/Opentrons/ot3-firmware) currently, a `tools_detected_notification` will appear when you attach an instrument. After taking off that instrument and sending an `attached_tools_request`, you'll still see the instrument that was previously attached.

This pr allows the detected tools to change following the removal of an instrument without sending an unnecessary `tools_detected_notification`.


sfoster1 (changes requested):
Need some tests for this one I think. However, great fix!

sfoster1 (approved):
This looks good, but why not send an update when the tool goes away?

PR #461, opened 2022-10-17
fix(gripper): use temporary serial number for simulator

ahiuchingau:
closes RQA-318

### DIFF
--- a/gripper/simulator/main.cpp
+++ b/gripper/simulator/main.cpp
@@ -70,13 +70,14 @@ int main(int argc, char** argv) {
     LOG_INIT("GRIPPER", []() -> const char* {
         return pcTaskGetName(xTaskGetCurrentTaskHandle());
     });
+    const uint32_t TEMPORARY_SERIAL = 0x103321;
     auto options = handle_options(argc, argv);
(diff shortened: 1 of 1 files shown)


gripper/simulator/main.cpp
```diff
@@ -70,8 +70,10 @@ int main(int argc, char** argv) {
     LOG_INIT("GRIPPER", []() -> const char* {
         return pcTaskGetName(xTaskGetCurrentTaskHandle());
     });
+    const uint32_t TEMPORARY_SERIAL = 0x103321;
```
Laura-Danielle, line 73: I think this will error if the number isn't `const uint32_t TEMPORARY_SERIAL = 0x10003321` right? Otherwise the model number will be something unexpected rather than 0 or 1?

Laura-Danielle (approved):
Had a comment about the serial number but looks fine otherwise.

DerekMaggio:
@ahiuchingau, do I need to set an env var to create the gripper eeprom file? 
Similar to how it needs to be set for pipettes?

DerekMaggio:
It seems that OT-3 emulation is running now. 
THANK YOU!!!
YOU ARE AWESOME!!!

PR #462, opened 2022-10-20
refactor(pipettes): modify the GPIO pins of the 96-channel for EVT

Laura-Danielle:
## Overview

GPIO pins have been changed around/added to support requests from firmware/to make the board nicer. The 96 channel board is now going to be using a G4 board, but we already have support for both the L5/G4 thanks to previous work.


pipettes/firmware/motor_timer_hardware.c
```diff
@@ -34,7 +34,7 @@ void MX_TIM6_Init(void) {
     TIM_MasterConfigTypeDef sMasterConfig = {0};
 
     htim6.Instance = TIM6;
-    htim6.Init.Prescaler = 499;
+    htim6.Init.Prescaler = 849;
```
sfoster1, line 37: Which timer is this? Is it undoing https://github.com/Opentrons/ot3-firmware/pull/464 ?

sfoster1 (approved):
Looks good overall but confused about the timer freqs

pipettes/firmware/motor_timer_hardware.c
Laura-Danielle: I think accidental unresolved merge.

PR #464, opened 2022-10-20
refactor: increase motor interrupt freqs and usteps

sfoster1:
We'd like to increase the microstep ratios for all the stepper-controlled axes to make motion smoother and quieter. However, this doubles the steps/mm value for each axis, and therefore lowers the maximum speed we can command - we can only issue a step once each interrupt. So, double the interrupt frequency to 200khz to keep the same maximum speed.

This must be tested alongside https://github.com/Opentrons/opentrons/pull/11598 . If that PR isn't being used, the motors will move half as fast as they should.

### DIFF
--- a/head/firmware/main_proto.cpp
+++ b/head/firmware/main_proto.cpp
@@ -152,7 +152,7 @@ static tmc2130::configs::TMC2130DriverConfig motor_driver_configs_right{
                          .hstrt = 0x5,
                          .hend = 0x3,
                          .tbl = 0x2,
-                         .mres = 0x4},
+                         .mres = 0x3},
             .coolconf = {.sgt = 0x6},
         },
     .current_config =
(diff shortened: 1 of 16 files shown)


head/firmware/main_proto.cpp
```diff
@@ -152,7 +152,7 @@ static tmc2130::configs::TMC2130DriverConfig motor_driver_configs_right{
                          .hstrt = 0x5,
                          .hend = 0x3,
                          .tbl = 0x2,
-                         .mres = 0x4},
+                         .mres = 0x3},
```
Laura-Danielle, line 155: We want to make this change on the proto too?

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
```diff
@@ -37,9 +37,6 @@ using namespace motor_messages;
  * Note: The position tracker should never be allowed to go below zero.
  */
 
-// (TODO lc): This should probably live in the motor configs.
-constexpr const int clk_frequency = 85000000 / (5001 * 2);
```
Laura-Danielle, line 41: r.i.p. 🪦

Laura-Danielle (approved):
One question about making this change in the proto firmware, but otherwise lgtm.

head/firmware/main_proto.cpp
sfoster1: there's no reason not to, really

PR #466, opened 2022-10-20
refactor(pipettes): make the sync out line low on boot

Laura-Danielle:
We expect the sync out line to be low until it is triggered by some event. We should set it low by default to make sure that it's not a floating value.


Laura-Danielle:
NVM. Got confused by active low vs active high.

PR #467, opened 2022-10-21
refactor(pipettes): make all motors on the EVT 96 chan use the tmc2160

Laura-Danielle:
All of the motors are now using the tmc2160 motor driver. I made an executive decision to no longer support the gear motors on the L5 boards for the 96 channel because it was getting to be gross to try and support both configurations.

We will only be using the old L5 proto board for plunger testing on the 96 channel.

### DIFF
--- a/include/pipettes/core/gear_motor_tasks.hpp
+++ b/include/pipettes/core/gear_motor_tasks.hpp
@@ -4,8 +4,8 @@
 #include "can/core/ids.hpp"
 #include "can/core/message_writer.hpp"
 #include "motor-control/core/linear_motion_system.hpp"
-#include "motor-control/core/stepper_motor/tmc2130.hpp"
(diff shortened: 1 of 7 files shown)


sfoster1 (changes requested):
Let's not commit code that "builds" but doesn't do anyhting on L5s - let's remove the ability to  build 96 channels on L5.

Laura-Danielle:
> Let's not commit code that "builds" but doesn't do anyhting on L5s - let's remove the ability to build 96 channels on L5.

It will still move the plunger -- which we will be using for testing on the 96 chan plunger prototype. That is why I haven't removed support yet. Please see [this thread](https://opentrons.slack.com/archives/C01TAVBKL73/p1665607038638149).

sfoster1 (approved):
Okay, let's do it then!

PR #468, opened 2022-10-25
chore(cmake): update for poetry fixes

fsinapi:
Pulls latest changes from cmake-utils to enable host configuration on macOS. May require wiping out the `build-host` folder to reconfigure and build correctly.


Laura-Danielle (approved):
nioce

PR #469, opened 2022-10-25
fix(gripper-motor): use the tmc2130 emergency stop pin to reflect new board change

ryanthecoder:
The tmc2130 has a config register that enables an extra emergency_stop pin on the driver. for the gripper we re-worked the mcu board so that the e-stop line coming from the SoM board is routed to this pin instead of the motor-enable pin.
This has the effect of leaving the motor enabled when the e-stop is triggered but the motor driver no longer accepts any steps that are sent to it.
this means that during estop the gripper z motor won't fall straight into the deck and hold steady wherever it is.

The re-work also holds the motor enable pin to ground so the motor driver is always enabled

### DIFF
--- a/gripper/firmware/interfaces_z_motor.cpp
+++ b/gripper/firmware/interfaces_z_motor.cpp
@@ -67,7 +67,7 @@ static motor_hardware::MotorHardware motor_hardware_iface(motor_pins, &htim7,
  * Motor driver configuration.
  */
 static tmc2130::configs::TMC2130DriverConfig MotorDriverConfigurations{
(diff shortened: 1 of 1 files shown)


Laura-Danielle:
Do we also need to change the motor enable pin configuration in the `BrushedHardwareConfig`?

ryanthecoder:
no we only re-worked the gripper z board. the brushed motor driver isn't connected to the e-stop line at all.

sfoster1 (approved):
Looks good to me!

PR #470, opened 2022-10-26
chore(pipettes): support the 96 channel pipette on an EVT head board 

Laura-Danielle:
## Overview

Temporarily, we'll be using a head board to test pick up and drop tip of full tipracks on the 96
channel. All of this code is throw away and should be deleted immediately once EVT 96 chan
electronics have been validated.

## To-Do

- [ ] Validate on hardware.


Laura-Danielle:
No longer necessary 😿

PR #471, opened 2022-10-26
feat(simulators): add state manager socket connection

fsinapi:
This PR adds support for the OT-3 simulators to connect to the State Manager datagram socket server.

- Adds command line options `--state-mgr-host` and `--state-mgr-port` to tell the simulator where to connect to the state manager.
- Adds environment variables `STATE_MANAGER_HOST` and `STATE_MANAGER_PORT` as alternative methods to configure the host and port 
- Adds a class StateManagerConnection to manage the connection to the state manager and send messages over UDP

Testing is as follows:
- Added some debugging lines in `state_manager::create()` to send a short string over UDP upon initialization.
  - Opened a netcat UDP server on the same port as the simulator and verified that it printed out the data
- Added a loop of `await asyncio.sleep(1)` to the State Manager python project, as well as a print statement on datagram reception
  - Started the State Manager and one of the simulators, verified that the State Manager received the message from the simulator


DerekMaggio:
TODO for @DerekMaggio 
1. Add signal handler to gracefully exit.
2. Take the value returned from `start_server` and call `val.close`

DerekMaggio (approved):
Everything looks good as far as I can tell. 
I am working on adding the State Manager as a container to `opentrons-emulation`. 

If everything looks good to someone who knows C++, I would suggest to merge this, and I can test it inside of `opentrons-emulation`. We can iterate if we need to from there.

Laura-Danielle (approved):
Looks great to me!

include/common/simulation/state_manager.hpp
```diff
@@ -0,0 +1,118 @@
...
+    StateManagerConnection(const StateManagerConnection &) = delete;
+    StateManagerConnection(const StateManagerConnection &&) = delete;
+    StateManagerConnection &operator=(const StateManagerConnection &) = delete;
+    StateManagerConnection &&operator=(const StateManagerConnection &&) =
+        delete;
+
+    template <size_t N>
+    auto send(std::array<uint8_t, N> data) -> bool {
+        auto lock = synchronization::Lock(critical_section);
+        if (_socket.is_open()) {
```
sfoster1, line 57: Let's throw if we're not open here

sfoster1 (approved):
Looks good but let's throw if we're not open in `send()` (or just remove the check since we have RAII now)

PR #472, opened 2022-10-26
chore: Install ARM version of clang

DerekMaggio:
When on Linux and using an ARM container install ARM Linux version of clang


DerekMaggio:
Opened this PR in the wrong place. Need to open it in cmake_utils

PR #474, opened 2022-10-27
fix(pipettes): enable reading from encoder

caila-marashaj:
## **Overview**
This code allows us to read values from the pipette encoders, which appear to be very close to the corresponding `current_position_um` values calculated from stepper motor pulses. 

## **Changelog**
- Moves encoder GPIO and timer `TIM2` initialization to `motor_timer_hardware.c`
- Enabled encoder overflow interrupt
- Changes polarity of one of the encoder input clocks- this changes the encoder reading from negative to positive values
- added encoder overflow callback


include/pipettes/firmware/interfaces.hpp
```diff
@@ -63,22 +63,22 @@ auto get_interrupt_queues<PipetteType::THREE_EIGHTY_FOUR_CHANNEL>()
...
-auto get_interrupt(pipette_motor_hardware::MotorHardware& hw,
+auto get_interrupt(motor_hardware::MotorHardware& hw,
                    HighThroughputInterruptQueues& queues)
     -> MotorInterruptHandlerType<linear_motor_tasks::QueueClient>;
-auto get_motor_hardware(motor_configs::LowThroughputPipetteMotorHardware pins)
-    -> pipette_motor_hardware::MotorHardware;
-auto get_motor_hardware(motor_configs::HighThroughputPipetteMotorHardware pins)
-    -> pipette_motor_hardware::MotorHardware;
-auto get_motion_control(pipette_motor_hardware::MotorHardware& hw,
+auto get_motor_hardware(motor_hardware::HardwareConfig pins)
```
caila-marashaj, line 72: I'm changing a couple of these getter functions in this file to use `motor_hardware` class objects instead of `pipette_motor_hadware`- this shouldn't negatively affect anything else for the pipettes, right?

PR #475, opened 2022-10-27
fix(gantry): set up encoder for XY

ahiuchingau:
We didn't actually set up the encoders for XY.. Here it is!


sfoster1 (approved):
It's TIM!

Laura-Danielle (approved):
ayy

PR #476, opened 2022-10-28
refactor(tests): remove intedependencies between tests and simulations

fsinapi:
There was a bit of shared code and dependencies between test and simulation targets that was causing trouble in another branch, primarily due to header generation from the state manager. This PR refactors the tests and simulation targets to try to separate some of those inter-dependencies.

This PR also adds a dependency in common-simulation to ensure the headers are always generated from state manager, and that the state-manager environment is always initialized before header generation.

### DIFF
--- a/common/simulation/CMakeLists.txt
+++ b/common/simulation/CMakeLists.txt
@@ -5,6 +5,8 @@ add_library(common-simulation STATIC
 
 target_link_libraries(common-simulation PUBLIC can-core Boost::boost Boost::date_time pthread)
 
+add_dependencies(common-simulation state_manager)
+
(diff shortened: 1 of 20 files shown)


sfoster1 (approved):
That looks good to me. It's duplicated code right now but I wonder if we can pull out things from one that aren't necessary in the other? And we'll definitely be able to only add things to one and not the other.

Laura-Danielle (approved):
Nice refactor.

fsinapi:
> That looks good to me. It's duplicated code right now but I wonder if we can pull out things from one that aren't necessary in the other? And we'll definitely be able to only add things to one and not the other.

I'm going to be editing the simulator version pretty heavily in another PR, will probably leave it as-is for now to avoid merge conflict hell if that's alright

PR #477, opened 2022-10-28
feat(pipettes): support multiple LED pins for pipettes

Laura-Danielle:
The low throughput EVTs have different LED pins than the high throughput. We should selectively pass in the correct GPIO pin depending on which pipette is selected. This has been verified on the new 96 channel EVT board.


sfoster1 (approved):
let there be light

PR #478, opened 2022-10-31
feat(simulators): add state manager messaging

fsinapi:
Adds support for the OT-3 firmware simulators to send messages to the state manager:
- When the motors are stepped (with the direction inferred from the last time the direction was set)
- When the sync pin is set 
- When reading the sync pin status

For now, every message sent to the state manager comes with a response, so the firmware must wait for that response in order to remain synchronized. This slows things down but will be modified in the future.

Was able to test this functionality through the following:
- Set up an OT3 emulator pulling the firmware from the latest commit on this branch
- Open a can-comm, set up a move group for an axis that will trigger a nonzero number of steps, and then start it
- Using a simple UDP python script, request the position of the aforementioned axis from the state manager.
Every time the move group is executed, the position increases. If the duration is long enough you can watch as the position increases over time.

This should not be merged until the matching [opentrons-emulation PR](https://github.com/Opentrons/opentrons-emulation/pull/214) is merged, otherwise the firmware will hang when trying to read responses.

### DIFF
--- a/common/simulation/CMakeLists.txt
+++ b/common/simulation/CMakeLists.txt
@@ -11,8 +11,11 @@ set_target_properties(common-simulation
         PROPERTIES CXX_STANDARD 20
         CXX_STANDARD_REQUIRED TRUE)
 
-target_include_directories(common-simulation
-        PUBLIC ${CMAKE_SOURCE_DIR}/include)
+target_include_directories(common-simulation PUBLIC 
+        ${CMAKE_SOURCE_DIR}/include
+        ${CMAKE_SOURCE_DIR}/cpp-utils/include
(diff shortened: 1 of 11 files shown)


sfoster1 (commented):
Why do we need synchronous message passing? To get by-the-tick results injection?

fsinapi:
> Why do we need synchronous message passing? To get by-the-tick results injection?

The responses for the motor-step messages don't matter (the state manager just echoes), but when the simulators check the sync pin they have to parse the response. If we don't read the responses to the step messages, we'll become de-synced and unable to read back the correct response for reading the sync pin.

We have a [ticket](https://opentrons.atlassian.net/browse/RCORE-319) to remove the readback on both the python and the firmware side on the same branch.

PR #480, opened 2022-11-02
fix(gantry, head): Updated EVT tmc2160 register settings, plus lowering microstepping

andySigler:
# Context

Big discovery, David Gomez found our EVT units had the wrong sense resistors. This was why there were SO DAMN LOUD. So now, even at 32 microstepping, the gantry sounds just fine.

We found that at 64 microstepping, we aren't yet able to reach the physical max speeds of the gantry. If we lower it to 32, then we can.

David has also been doing testing using a scope to try and get our motor waveforms close to ideal.

# This PR

Makes the following changes:

1. Decreases gantry microstepping from 64 to 32
2. Decreases head microstepping from 32 to 16
3. Explicitly disables Stealth-Chop on both head and gantry


andySigler:
@djgomez-opentrons is my description accurate?

sfoster1 (approved):
Okay, this looks good to me. Glad we don't have to increase the microstepping so much.

ahiuchingau (approved):
What sensor were we using before?? Did we use the wrong sensor for the TMC2130 drivers as well?

djgomez-opentrons (approved):
Looks good. 

For additional context:
- I found the audible difference between 64/32 and 32/16 microstepping wasn't massive. The fastest speeds will be achievable with lower microstepping. 
- The Sense resistor was found to be an incorrectly populated part that had higher inductance than specified - this disrupts the function of the current sense circuit and severely disrupts the current waveform used to drive the motor which creates a lot of audible noise. 
- Haven't checked if any of our TMC2130 designs are impacted by the same issue, this was a manufacturing issue not a design one so it potentially could be some TMC2130 based PCBAs and not others.

PR #481, opened 2022-11-03
refactor(all): use the external oscillator for the system clock

Laura-Danielle:
Since the external oscilator is the same frequency as the internal oscilator, we should be able to use the same clock divisors.

### DIFF
--- a/bootloader/firmware/stm32G4/stm32g4xx_hal_conf.h
+++ b/bootloader/firmware/stm32G4/stm32g4xx_hal_conf.h
@@ -115,7 +115,7 @@
   *        (when HSE is used as system clock source, directly or through the PLL).
   */
 #if !defined  (HSE_VALUE)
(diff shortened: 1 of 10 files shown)


fsinapi (commented):
I think you also have to change the value in `stm32g4xx_hal_conf.h` for each target

sfoster1 (changes requested):
- do this for the bootloaders (also a change in `system_stm32g4xx.c` for the bootloaders)
- for now, explicitly set `RCC_OscInitStruct.HSIState = RCC_HSI_OFF;` since the bootloader will have it enabled until we update the bootloaders on those boards
- agreed about the `hal_conf` change

sfoster1 (approved):
Looks good to me!

ryanthecoder:
Tested on hardware and works well

PR #482, opened 2022-11-03
fix(gripper): use a simpler serial

ahiuchingau:
This allows us to parse the gripper id on the Python side so we can save the gripper offset during calibration.


Laura-Danielle (changes requested):
can we make this a bigger number than 1? If we could get it similar to the pipette serial, that would be great! I think if you add data into the first 11 bytes it _should_ work.

This is what the output looks like for me:

```
choose MessageId (? for list): 2022-11-04 14:28:20.450913 pipette_info_response (pipette_left->host)
 	name: PipetteNameField(value=p1000_single)
 	model: UInt16Field(value=33)
 	serial: SerialDataCodeField(value=b'20200727A05\x00\x00\x00\x00\x00')
```

Laura-Danielle (approved):
🌮

PR #483, opened 2022-11-03
feat(simulators): disable state manager connection for now

fsinapi:
State manager connection has some issues. Will be re-enabled in a future PR.


DerekMaggio (approved):
Looks good to me

Laura-Danielle (approved):

PR #485, opened 2022-11-03
fix(pipettes): enable reading from encoder

caila-marashaj:
## Overview
This code allows us to read values from the pipette encoders, which appear to be very close to the corresponding current_position_um values calculated from stepper motor pulses.

https://opentrons.atlassian.net/browse/RLIQ-235
## Changelog
- Moves encoder GPIO and timer TIM2 initialization to motor_timer_hardware.c
- Enabled encoder overflow interrupt
- Changes polarity of one of the encoder input clocks- this changes the encoder reading from negative to positive values
- added encoder overflow callback


include/pipettes/firmware/interfaces.hpp
```diff
@@ -63,22 +63,21 @@ auto get_interrupt_queues<PipetteType::THREE_EIGHTY_FOUR_CHANNEL>()
 
 namespace linear_motor {
 
-auto get_interrupt(pipette_motor_hardware::MotorHardware& hw,
+auto get_interrupt(motor_hardware::MotorHardware& hw,
                    LowThroughputInterruptQueues& queues)
     -> MotorInterruptHandlerType<linear_motor_tasks::QueueClient>;
-auto get_interrupt(pipette_motor_hardware::MotorHardware& hw,
+auto get_interrupt(motor_hardware::MotorHardware& hw,
```
caila-marashaj, line 69: I'm changing a couple of these getter functions in this file to use `motor_hardware::MotorHardware` objects instead of `pipette_motor_hardware::MotorHardware` - this shouldn't mess with any other pipette firmware, right?

Laura-Danielle: We discussed f2f, but this will affect the L5 project so we should create separate header files to handle that.

pipettes/firmware/motor_timer_hardware.c
```diff
@@ -59,6 +103,10 @@ void HAL_TIM_PeriodElapsedCallback(TIM_HandleTypeDef *htim) {
         plunger_callback();
     } else if ((htim == &htim6) && gear_callback) {
         gear_callback();
+    } else if ((htim == &htim2) && encoder_callback) {
+    uint32_t direction = __HAL_TIM_IS_TIM_COUNTING_DOWN(htim);
```
fsinapi, line 107: Can this get indented?

include/pipettes/firmware/interfaces_g4.hpp
```diff
@@ -0,0 +1,126 @@
...
+// NOLINTNEXTLINE(clang-diagnostic-unknown-warning-option)
+#pragma GCC diagnostic ignored "-Wvolatile"
+#include "motor_timer_hardware.h"
+#pragma GCC diagnostic pop
+
+/*
+ * (CM, 11/3/2022)
+ * This file exists in addition to interfaces.hpp to accommodate the l5 firmware
+ * after the changes in the enable-pipette-encoder-output branch. interfaces.hpp
+ * can be deleted with the l5 firmware.
```
Laura-Danielle, line 23: Can you add in this comment that the `interfaces_g4.hpp` should be renamed `interfaces.hpp` when this happens?

pipettes/firmware/CMakeLists.txt
```diff
@@ -31,8 +31,7 @@ set(PIPETTE_FW_NON_LINTABLE_SRCS
     ${CMAKE_CURRENT_SOURCE_DIR}/i2c_setup.c
     ${CMAKE_CURRENT_SOURCE_DIR}/can.c
     ${CMAKE_CURRENT_SOURCE_DIR}/utility_gpio.c
-    ${CMAKE_CURRENT_SOURCE_DIR}/motor_encoder_hardware.c
-    ${CMAKE_CURRENT_SOURCE_DIR}/motor_hardware.c
+        ${CMAKE_CURRENT_SOURCE_DIR}/motor_hardware.c
```
fsinapi, line 34: Indentation

fsinapi (commented):
Couple of really tiny formatting requests. LGTM but holding off on approving since I helped on this.

pipettes/firmware/utility_gpio.c
```diff
@@ -77,6 +77,35 @@ void limit_switch_gpio_init() {
     }
 }
 
+void encoder_gpio_init() {
+//    PipetteType pipette_type = get_pipette_type();
```
Laura-Danielle, line 81: Why is this commented out?

pipettes/firmware/motor_timer_hardware.c
```diff
@@ -15,6 +18,47 @@ static gear_motor_interrupt_callback gear_callback = NULL;
 // /1 AHB
 // /2 APB1
 // /425 prescaler * 1 count = 200kHz
+
+void TIM2_Encoder_Init(void) {
```
Laura-Danielle, line 22: Can we check the 96 chan schematic and make sure we're also supporting whatever we need to for the encoder?

Laura-Danielle (changes requested):
1. Got a random hex file in your PR for some reason.
2. We no longer need to include `motor_hardware.hpp` in `pipettes/core/interfaces.hpp` since we have a separate header file for the G4 now.

A few other comments, but in general looks good. Nice work!

pipettes/firmware/utility_gpio.c
caila-marashaj: it was because it was being used for the encoder index pin initialization which we're no longer using- I'll take this out

pipettes/firmware/motor_timer_hardware.c
caila-marashaj: checked on this- pins are the same for the 96 channel

fsinapi (approved):
Changes LGTM!

include/pipettes/core/interfaces.hpp
```diff
@@ -3,6 +3,7 @@
 #include "common/core/freertos_message_queue.hpp"
 #include "motor-control/core/motor_messages.hpp"
 #include "motor-control/core/stepper_motor/motion_controller.hpp"
+#include "motor-control/firmware/stepper_motor/motor_hardware.hpp"
```
Laura-Danielle, line 6: Still have the change in here that's not necessary.

Laura-Danielle (approved):
You gotta remove that include of `motor-control/firmware/stepper_motor/motor_hardware.hpp` in `include/pipettes/core/interfaces.hpp`, otherwise it looks good to me.

PR #486, opened 2022-11-03
feat(simulators): make state manager comms asynchronous

fsinapi:
This PR refactors the state manager to make all communications asynchronous. This includes some small revisions to the Python side of things, and more extensive revisions to the C++ side.

Python:
* Not all commands have responses, and the format of the responses has changed slightly (documented in detail on Confluence)
* When the sync pin's value is updated by a message, the state manager broadcasts the new status of the sync pin to _every_ client that it has seen since starting up.
* Added a CLI tool for testing the state manager - allows sending packets and asynchronously monitors for inputs

simulators:
* Switched the udp socket to use asynchronous commands. This involves adding a dedicated thread in each simulator for handling the udp sending/receiving callbacks
* State manager messages no longer wait for responses. Instead, the message is stuffed into a deque to send out.
* The sync pin status is now polled from an atomic variable in the state manager class, which is updated whenever a sync pin status response is parsed
* The state manager class will attempt to "register" itself with the state manager by sending a request for the sync pin status. An asynchronous timer will send this request every 5 seconds until a response comes back.

Tested this by setting up an opentrons-emulation instance (including the necessary change from [this PR](https://github.com/Opentrons/opentrons-emulation/pull/220)):
* After starting up the emulation instance, try using the CLI tool to change the sync pin status. In the emulation logs, can see that each simulator got the update and changed its sync pin status
* Use `can-comm` to run a move group that moves gantry-x (duration 1000000, velocity 1000). Confirm with the CLI tool that the position of the X axis is now 149. Running the move group a second time doubles the position. Putting in a negative velocity and running the move group decreases the position. The position does not decrease below 0.

Tested by just running one of the simulators in a shell:
* If there's no state manager running locally, the connection timer adds its log entry once every 5 seconds.
* Once the state manager is started, the next connection timer log entry is the last.


include/common/simulation/state_manager.hpp
```diff
@@ -167,4 +305,20 @@ auto create(const boost::program_options::variables_map &options)
...
+    using Connection = std::shared_ptr<StateManagerConnection<CriticalSection>>;
+    [[noreturn]] void operator()(Connection *connection_ptr) {
+        if (connection_ptr) {
+            auto connection = *connection_ptr;
+            if (connection) {
+                connection->run();
+            }
+        }
+        LOG("State manager spuriously exited");
+        while (1) {
```
sfoster1, line 319: may want a sleep here

state_manager/state_manager/messages.py
```diff
@@ -26,15 +26,16 @@
...
     is_error: bool = False
+    broadcast: bool = False
 
     def to_bytes(self) -> bytes:
         """Convert Response object into bytes."""
         if self.content is None:
             return b""
-        message = f"ERROR: {self.content}" if self.is_error else self.content
-        return message.encode()
+        message = b"ERROR: " + self.content if self.is_error else self.content
```
sfoster1, line 37: kinda prefer the old way here - the `+` makes it a little tougher to parse what the `if` applies to, i just like f-strings a lot, etc

sfoster1 (approved):
Looks good as a whole! Might be some little fixups though inline

state_manager/state_manager_cli.py
```diff
@@ -0,0 +1,53 @@
+# Util to access the state manager.
+# To send data to the state manager, enter 4 whitespace-separated numbers and hit enter.
```
DerekMaggio, line 2: Would it just be easier to have them enter hex numbers instead of ints?
Since you are just converting to a hex value anyways?

state_manager/tests/test_messages.py
```diff
@@ -423,11 +428,19 @@ def test_valid_handle_get_sync_pin_state_message(ot3_state: OT3State) -> None:
...
     ack = handle_message(HIGH_MESSAGE.to_bytes(), ot3_state)
-    assert ack == HIGH_MESSAGE.to_bytes()
+    assert ack is not None
+    assert ack.broadcast
     assert ot3_state.get_sync_pin_state()
-    assert handle_message(GetSyncPinStateMessage().to_bytes(), ot3_state) == b"1"
+    ack = handle_message(GetSyncPinStateMessage().to_bytes(), ot3_state)
+    assert ack is not None
+    assert not ack.broadcast
+    assert ack.to_bytes() == b"\x031"
```
DerekMaggio, line 437: Can you set b"\x031" to a constant? I am wanting context of what b"\x031" represents.

state_manager/tests/test_messages.py
```diff
@@ -423,11 +428,19 @@ def test_valid_handle_get_sync_pin_state_message(ot3_state: OT3State) -> None:
...
     ack = handle_message(LOW_MESSAGE.to_bytes(), ot3_state)
-    assert ack == LOW_MESSAGE.to_bytes()
+    assert ack is not None
+    assert ack.broadcast
     assert not ot3_state.get_sync_pin_state()
-    assert handle_message(GetSyncPinStateMessage().to_bytes(), ot3_state) == b"0"
+    ack = handle_message(GetSyncPinStateMessage().to_bytes(), ot3_state)
+    assert ack is not None
+    assert not ack.broadcast
+    assert ack.to_bytes() == b"\x030"
```
DerekMaggio, line 446: Same here

state_manager/state_manager_cli.py
DerekMaggio, line 2: Would it just be easier to have them enter hex numbers instead of ints?
Since you are just converting to a hex value anyways?

DerekMaggio (approved):

state_manager/state_manager_cli.py
fsinapi: Yeah probably, didn't write this with any intention of putting in in version control but it ended up being more useful than I expected. Will switch it 👍

PR #490, opened 2022-11-04
fix(sensors): fix a halving error in the capacitive sensor

Laura-Danielle:
## Overview

There was a mismatch on how we were converting values in the capacitive sensor to fixed point. I
decided to match everything to a radix of 16, which is what the python is currently doing.

### DIFF
--- a/include/sensors/core/fdc1004.hpp
+++ b/include/sensors/core/fdc1004.hpp
@@ -1,6 +1,9 @@
 #pragma once
 
 #include <limits>
+#include <cstdlib>
+#include "common/core/logging.h"
+
 
 // TODO (lc 02-16-2022) We should refactor the fixed point
 // helper functions such that they live in a shared location.
@@ -69,8 +72,8 @@ constexpr float MAX_CAPDAC_PF =
 constexpr std::size_t CONVERSION_BITS = 24;
 constexpr float MAX_MEASUREMENT_PF = 15;
 constexpr float MAX_RAW_MEASUREMENT =
-    float(std::numeric_limits<int32_t>::max() >>
-          ((sizeof(int32_t) * 8) - CONVERSION_BITS));
+    float(std::numeric_limits<int32_t>::max() >> 8);
+//          ((sizeof(uint32_t) * 8) - CONVERSION_BITS));
 
 // Because we're doing big gantry moves, we'll probably have to handle
 // the parasitic capacitance of the system shifting around a lot. We'll
--- a/include/sensors/core/fdc1004.hpp
+++ b/include/sensors/core/fdc1004.hpp
@@ -73,9 +73,7 @@ constexpr std::size_t CONVERSION_BITS = 24;
 constexpr float MAX_MEASUREMENT_PF = 15;
 constexpr float MAX_RAW_MEASUREMENT =
     float(std::numeric_limits<int32_t>::max() >>
-          (sizeof(uint32_t) * 8) - CONVERSION_BITS);
-//    float(std::numeric_limits<int32_t>::max() >> 8);
-//          ((sizeof(uint32_t) * 8) - CONVERSION_BITS));
+          ((sizeof(uint32_t) * 8) - CONVERSION_BITS));
 
 // Because we're doing big gantry moves, we'll probably have to handle
 // the parasitic capacitance of the system shifting around a lot. We'll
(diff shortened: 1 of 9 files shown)


fsinapi (approved):
This fix looks good. 

Since this radix is being used all across the can messages, I think it would be really helpful to have it as a constant somewhere widely accessible in this codebase to prevent mismatches like this.

Laura-Danielle:
> This fix looks good.
> 
> Since this radix is being used all across the can messages, I think it would be really helpful to have it as a constant somewhere widely accessible in this codebase to prevent mismatches like this.

great idea! I'll add that.

include/sensors/core/fdc1004.hpp
```diff
@@ -70,7 +73,7 @@ constexpr std::size_t CONVERSION_BITS = 24;
 constexpr float MAX_MEASUREMENT_PF = 15;
 constexpr float MAX_RAW_MEASUREMENT =
     float(std::numeric_limits<int32_t>::max() >>
-          ((sizeof(int32_t) * 8) - CONVERSION_BITS));
+          ((sizeof(uint32_t) * 8) - CONVERSION_BITS));
```
sfoster1, line 76: let's not change this - it makes no difference to the generated code (the value is 4 either way) but everything else is `int32` and it makes things nicer to keep it the same

include/sensors/core/mmr920C04.hpp
```diff
@@ -265,7 +265,7 @@ struct __attribute__((packed, __may_alias__)) Temperature {
     [[nodiscard]] static auto to_temperature(uint32_t reg) -> sq15_16 {
         float temperature =
             CONVERT_TO_CELSIUS * (static_cast<float>(reg) / MAX_SIZE);
-        return convert_to_fixed_point(temperature, 15);
+        return convert_to_fixed_point(temperature, 16);
```
sfoster1, line 268: why don't we just make a `static constexpr S15Q16 = 16` in the same header that provides `convert_to_fixed_point` or something so we can use that instead of writing the number everywhere

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
```diff
@@ -45,8 +45,10 @@ struct ReadCapacitanceCallback {
         auto capacitance = convert_capacitance(
             convert_reads(polling_results[0], polling_results[1]), 1,
             current_offset_pf);
-        auto new_offset = update_offset(capacitance, current_offset_pf);
-        set_offset(new_offset);
+        // TODO (lc 1-2-2022) we should figure out a better strategy for
```
sfoster1, line 48: I still don't quite get why this capdac update strategy is wrong. Plus, won't disabling the capdac entirely break the general purpose use of the sensor?

sfoster1 (changes requested):
This is a good fix but agree with Frank, and I'm a little worried about disabling the capdac entirely

include/sensors/core/mmr920C04.hpp
Laura-Danielle: yep @fsinapi just suggested the same thing.

include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
joecrabtree: I think that's a debug change sneaking into the commit. It's temporarily disable for debugging the other capacitance sensor issues. This change doesn't need to be part of this PR.

Laura-Danielle: I can add it back in and then leave a comment. I think @joecrabtree was/is still worried about us setting the capdac to a spurious value (if we somehow get bad data from the IC) and so we should maybe have a better filtering strategy other than just assuming the data will be good by halving our sampling rate from the sampling rate we set on the IC.

joecrabtree: Yes, the implementation still has room for improvement, but we can do that along with whatever other changes fix the cap sense communication issues.

joecrabtree: If the sample returns to the correct value, yes. However, that could falsely trigger the threshold and cause the motor to stop.

sfoster1 (approved):
Looks good to me! Love the changes

PR #491, opened 2022-11-04
fix(pipettes): fix SPI gpios for the 96 channel

Laura-Danielle:
For some reason I completely forgot to change the SPI GPIOs in the motor configurations file.... 🙃


fsinapi (approved):
Checked against the schematic, looks good!

PR #492, opened 2022-11-07
feat(simulators): simulate encoder counts

fsinapi:
This PR adds functionality for the OT3 firmware simulators to simulate their stepper motor encoder ticks. 

Tested with an emulator running the `can-mon` and `can-comm` scripts. After moving an axis, the `current_position_um` and `encoder_position_um` fields in the `move_completed` message are reasonably close, with differences accounted for by the difference in resolution between the encoder and the motor microstepping.

_Note that the pipettes do not have the encoder configuration. This is pending #485_

Other changes:
* Updated the microstepping for the gantry simulator to match the firmware
* Updated the encoder steps/rev to 0 to disable the encoder on the gripper z motor, to match the firmware


gantry/simulator/interfaces.cpp
```diff
@@ -60,16 +60,18 @@ static tmc2130::configs::TMC2130DriverConfig driver_configs{
         .GPIO_handle = 0,
     }};
 
+static auto linear_motion_sys_config =
+    lms::LinearMotionSystemConfig<lms::BeltConfig>{
+        .mech_config = lms::BeltConfig{.pulley_diameter = 12.7},
+        .steps_per_rev = 200,
+        .microstep = 32,
```
Laura-Danielle, line 67: Is 32 the microstepping we're currently using in the gantry project?

gripper/simulator/interfaces.cpp
```diff
@@ -57,16 +57,17 @@ static tmc2130::configs::TMC2130DriverConfig MotorDriverConfigurations{
...
- */
-static motor_class::Motor motor{
+static auto z_motor_sys_config =
     lms::LinearMotionSystemConfig<lms::LeadScrewConfig>{
         .mech_config = lms::LeadScrewConfig{.lead_screw_pitch = 4},
         .steps_per_rev = 200,
         .microstep = 16,
-        .encoder_pulses_per_rev = 1000},
-    motor_interface,
+        .encoder_pulses_per_rev = 0};
```
Laura-Danielle, line 65: Do we not want this to be 1000?

Laura-Danielle (approved):
Had a few comments, but looks good to me in general :)

gantry/simulator/interfaces.cpp
fsinapi: Seems to be based on https://github.com/Opentrons/ot3-firmware/blob/main/gantry/core/utils.cpp#L20 

HOWEVER, that brings up the point that this should just get pulled from the core/utils file instead of duplicating the config

gripper/simulator/interfaces.cpp
fsinapi: There's no encoder on the gripper Z axis. The encoder is already disabled on the firmware side.

PR #495, opened 2022-11-10
fix(sensors): correctly mask capdac value

fsinapi:
There was a bug in the capacitive sensor driver related to masking the offset capacitance value. An offset of high enough pF would overflow into the CHB configuration field, effectively changing the channel being read to an invalid value and causing all future readings to be exactly the same. 

Before this fix, touching a capacitive sensor (or otherwise driving the capacitance high enough) would cause the sensor to latch onto a value indefinitely. After this fix, the sensor behaves properly.

### DIFF
--- a/include/sensors/core/fdc1004.hpp
+++ b/include/sensors/core/fdc1004.hpp
@@ -142,7 +142,7 @@ inline auto update_offset(float capacitance_pf, float current_offset_pf)
 }
 
 inline constexpr auto device_configuration_msb(uint8_t capdac_raw) -> uint8_t {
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Ahhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

Laura-Danielle (approved):
🤭 Does this mean we weren't actually shorting the sensor as it move to the deck, or is that still a problem?

joecrabtree:
> or

joecrabtree (commented):
Looks correct, capdac is in bits 9:5.

PR #496, opened 2022-11-10
refactor(pipettes): correct a few more bugs in the 96 channel firmware

Laura-Danielle:
Had some incorrect interrupt names for the G4 board + 96 channel data ready line. The GPIO pin for one of the limit switches was also incorrect


sfoster1 (approved):
lgtm!

PR #497, opened 2022-11-10
fix(sensors): environmental sensor bringup

fsinapi:
This PR adds a few fixes necessary to bring up the environmental sensor

* To read from the sensor, you first need to write the command for the autoread mode you want; then you read the data by writing to `AUTO_MEASURE_STATUS`, not the autoread mode setting register
* The polling frequency was 10Hz, which the datasheet claims can lead to internal heat buildup. 1Hz prevents this.
* The polling flags were not being read correctly, so if the host set the flags to sync _and_ report no data would be sent back
* When the host requested a single read, nothing was being sent back because data is only reported if the `report` flag is set
* The data coming back was being read in reverse order. The sensor sends back temperature and then humidity, but the firmware was reading in reverse order.

Tested on an OT3 using both a script to 1) poll one reading at a time 2) bind a report request and watch the data coming back. Temperature is right around room temp, humidity stays at 40%-ish and then goes up if you breathe on the sensor.

### DIFF
--- a/include/sensors/core/hdc3020.hpp
+++ b/include/sensors/core/hdc3020.hpp
@@ -94,6 +94,8 @@ struct __attribute__((packed, __may_alias__)) AutoMeasureStatus {
     static constexpr bool writable = true;
     static constexpr uint32_t value_mask = (1 << 8) - 1;
 
+    static constexpr uint8_t GET_READINGS_CMD = 0x00;
+
     uint8_t initialize : 1 = 1;
     uint8_t minimum_temperature : 1 = 0;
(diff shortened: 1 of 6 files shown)


fsinapi:
>     1. We'll need to update the simulated `include/sensors/hdc3020.hpp` to reflect the reversed order of temperature & humidity

Done

>     2. We'll then need to update the integration test in `hardware/tests/firmware_integration/test_sensors.py` in the monorepo.

I'm pretty sure the monorepo tests don't need to be updated since the simulated values should be the same, right?

PR #498, opened 2022-11-11
fix(sensors): increase capdac rezero threshold

fsinapi:
Another fix for the capacitive sensor to prevent the IC from latching.

The threshold in firmware for rezeroing the capacitive sensor was far lower than it had to be given the range of the base readings. This seemed to be causing the capacitive sensor IC to crash when the capacitance changed too quickly, and the capacitance would get stuck at a negative value.

This was causing issues in a testing script to verify deck-finding. After this fix, the script can run fine and the data never latches up.

### DIFF
--- a/include/sensors/core/fdc1004.hpp
+++ b/include/sensors/core/fdc1004.hpp
@@ -57,6 +57,10 @@ constexpr uint8_t SAMPLE_RATE_MSB = static_cast<uint8_t>(SAMPLE_RATE >> 8);
 constexpr uint8_t SAMPLE_RATE_LSB = static_cast<uint8_t>(SAMPLE_RATE & 0xff);
 constexpr uint16_t DEVICE_ID = 0x1004;
 
+constexpr uint8_t CAPDAC_MSB_MASK = 0x3;
+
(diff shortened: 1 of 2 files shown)


include/sensors/core/fdc1004.hpp
```diff
@@ -146,7 +148,7 @@ inline constexpr auto device_configuration_msb(uint8_t capdac_raw) -> uint8_t {
 }
 
 inline constexpr auto device_configuration_lsb(uint8_t capdac_raw) -> uint8_t {
-    return (DEVICE_CONFIGURATION_LSB | ((capdac_raw << 5) & 0xff));
+    return (DEVICE_CONFIGURATION_LSB | ((capdac_raw << 5) & 0xe0));
```
Laura-Danielle, line 151: Can we name this mask something so it's semi-easier to understand?

Laura-Danielle (approved):
Very excited for this fix! Thanks for looking into it :)

sfoster1 (approved):
Nice work! Let's call the capdac masks `CAPDAC_MASK_LSB` and `CAPDAC_MASK_MSB` or something

PR #500, opened 2022-11-11
fix(sensors): convert pressure data correctly

fsinapi:
This PR fixes some issues with how the pressure sensor data was being read:

* Pressure data value of 0 is perfectly reasonable
* Pressure data was being read as an `int32_t` and then right shifted. This ends up not working correctly because it is a signed 24-bit value. Instead, we want to right shift it _as a uint32_t_ to preserve the signedness 
* The pressure was being converted to a float and then stored into a `uint32_t` which 1) clobbered the accuracy 2) prevented negative values from being stored.


sfoster1 (approved):
Nice

PR #501, opened 2022-11-14
fix(sensors): only set FDC register once on startup

fsinapi:
The capacitive sensor on the pipettes worked fine with recent changes, but the gripper sensors would lock up from time to time. We theorized that the issue was that our poller doesn't check the ready flag for the measurements, but I set up a branch to test that and it didn't help at all - one of the sensors still locked up from time to time.

The code for the capacitive sensors originally wrote to the FDC register _every_ time the offset changed. This is necessary on startup (to configure the sample rate and measurement channel), but it is unnecessary after initial startup. Modifying the code so that the FDC register does _not_ get written after initialization fixed the issue - running the code on a gripper and touching the sensors quickly used to cause them to lock up eventually, but now they work indefinitely.

_Note: It's likely that this was the original cause of the capsense issues on the pipettes as well, but other "fixes" were able to obfuscate the issue due to the lower range of those sensors._

### DIFF
--- a/include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
+++ b/include/sensors/core/tasks/capacitive_sensor_callbacks.hpp
@@ -117,6 +117,12 @@ struct ReadCapacitanceCallback {
         hardware.reset_sync();
     }
 
+    void set_sample_rate() {
+        std::array configuration_data{FDC_CONFIGURATION, SAMPLE_RATE_MSB,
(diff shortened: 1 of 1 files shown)


ahiuchingau (approved):
GOD DAMN!!!

joecrabtree:
Sweet! It's still a good idea to add the check for data ready even though it wasn't the cause of this problem.

PR #502, opened 2022-11-14
fix(pipettes): use correct i2c bus for multichannel pressure sensor

fsinapi:
Closes RET-1280

The firmware was not set up to configure the right I2C bus for the pressure sensor on multichannel boards. This PR just adds that configuration. There is another pressure sensor on the other I2C bus, but it does not currently have the Data Ready line connected so it isn't usable until a sensor board revision.

Tested on a single-channel pipette:
* Flashed single-rev1 firmware and polled every sensor on the single-sensor board -> data came back fine
* Flashed multi-rev1 firmware and polled every sensor with a multi-sensor board attached -> data came back fine


Laura-Danielle (approved):
lgtm

PR #505, opened 2022-11-15
feature(simulators): Enable gripper g simulation

ryanthecoder:


include/motor-control/simulation/motor_interrupt_driver.hpp
```diff
@@ -38,7 +38,7 @@ class MotorInterruptDriver {
                         "%d, duration %ld",
                         move.group_id, move.seq_id, move.duration);
                     do {
-                        if (queue.peek(&move, 0)) {
+                        if (queue.try_read(&move, 0)) {
```
fsinapi, line 41: Pretty sure this wasn't a bug, the interrupt handler should have a reference to the same queue as this class and it manages reading reading messages as they come in (all handled in the `pulse()` function)

include/motor-control/simulation/brushed_motor_interrupt_driver.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+        void operator()() {
+            while (true) {
+                auto move = motor_messages::BrushedMove{};
+                if (queue.peek(&move, queue.max_delay)) {
+                    LOG("Enabling motor interrupt handler for group %d, seq "
+                        "%d, duration %ld",
+                        move.group_id, move.seq_id, move.duration);
+                    do {
+                        if (queue.try_read(&move, 0)) {
```
fsinapi, line 40: Similarly to the stepper simulator, I don't think we want to read the messages here. `brushed_motor_interrupt_handler` already reads from the queue using `try_read_isr`. We just want to peek here to see if the next move should trigger a stop condition because, while we want to make sure the interrupt handler sees the appropriate stop condition, we still need the handler to see the new move and set it as the active buffered movement.

include/motor-control/simulation/motor_interrupt_driver.hpp
ryanthecoder: Yea I got confused because consecutive calls to execute_move_group re-ran all the moves again but in the monorepo hardware controller, we always clear the move groups before starting a new one

Laura-Danielle (approved):
lgtm

PR #506, opened 2022-11-16
feat(motor-control): add motor position validity flags

fsinapi:
This PR adds flags tracking the validity of the stepper motor positions on the firmware boards. For now, these flags are set once the motor homes and they never get reset; in a future PR, they will be cleared by firmware to mark when a motor needs to be re-homed.

* The flags are stored in the Hardware Interface structure. This was done because the interface is shared by both the interrupt handler and the motor controller, which both need access to the flags.
* All operations on the flags are atomic in order to provide safety when accessed from interrupt vs task contexts.
* The flags are reported in two messages, the Encoder Position Response and the Move Completed message.

Tested on both a simulator and a physical OT-3 using `can-mon`:
* On startup, sending an Encoder Position request shows that the flags are at 0x00
* After a homing sequence on an axis is completed, the flags will indefinitely return 0x3 to indicate that both the step count and the encoder position are valid.

This PR should be merged at the same time as the matching Monorepo PR.

### DIFF
--- a/include/can/core/messages.hpp
+++ b/include/can/core/messages.hpp
@@ -357,10 +357,12 @@ struct MoveCompleted : BaseMessage<MessageId::move_completed> {
 struct EncoderPositionResponse
     : BaseMessage<MessageId::encoder_position_response> {
     int32_t encoder_position;
+    uint8_t position_flags;
 
     template <bit_utils::ByteIterator Output, typename Limit>
     auto serialize(Output body, Limit limit) const -> uint8_t {
         auto iter = bit_utils::int_to_bytes(encoder_position, body, limit);
+        iter = bit_utils::int_to_bytes(position_flags, iter, limit);
         return iter - body;
     }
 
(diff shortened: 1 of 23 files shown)


include/can/core/messages.hpp
```diff
@@ -355,10 +357,12 @@ struct MoveCompleted : BaseMessage<MessageId::move_completed> {
 struct EncoderPositionResponse
     : BaseMessage<MessageId::encoder_position_response> {
     int32_t encoder_position;
+    uint8_t position_flags;
```
sfoster1, line 360: can we make this a `PositionResponse` and have both the stepper and the encoder position?

sfoster1 (changes requested):
Can we add stepper position to that position response message? I think it'll be really useful to have around.

sfoster1 (changes requested):
Oh let's also clear the stepper-position-valid flag when we turn off the motor.

include/can/core/messages.hpp
fsinapi: Added this. The hardware interface now caches the steps and gets refreshed by the interrupt handler... I tried moving the position tracker completely, but turns out that ARM doesn't support 8-byte atomics 🙃

sfoster1 (approved):
Awesome, love it!

Laura-Danielle (approved):
this looks great to me!

PR #507, opened 2022-11-17
feature(all): add acknowledment and error message, and message index to all messages

ryanthecoder:
[RET-1244](https://opentrons.atlassian.net/browse/RET-1244)
[RET-1245](https://opentrons.atlassian.net/browse/RET-1245)
[RET-1246](https://opentrons.atlassian.net/browse/RET-1246)
[RET-1247](https://opentrons.atlassian.net/browse/RET-1247)

### DIFF
--- a/include/can/core/ids.hpp
+++ b/include/can/core/ids.hpp
@@ -61,6 +61,7 @@ enum class MessageId {
     gripper_grip_request = 0x42,
     gripper_home_request = 0x43,
     add_brushed_linear_move_request = 0x44,
+    acknowledgement = 0x50,
     read_presence_sensing_voltage_request = 0x600,
     read_presence_sensing_voltage_response = 0x601,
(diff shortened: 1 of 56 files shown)


sfoster1 (approved):
Looks good to me!

include/can/core/ids.hpp
```diff
@@ -125,6 +127,13 @@ enum class ErrorCode {
     hardware = 0x5,
 };
 
+/** Error Severity levels */
+enum class ErrorSeverity {
```
ahiuchingau, line 131: I think this enum is not being used anywhere yet, could we either remove it for now or add it to the bootloader `id.h` in case we forget to do so later?

ahiuchingau (approved):
This looks great! I just got one small request

Laura-Danielle (approved):
still looks good to me.

PR #509, opened 2022-11-21
refactor(pipettes): modify direction pin for 96 channel plunger

Laura-Danielle:
The active setting of the direction pin is flipped for the 96 channel


fsinapi (approved):
nice

PR #510, opened 2022-11-22
fix(head, gantry): update encoder resolution

fsinapi:
Closes RET-1285

Encoder resolution on the Gantry and Head projects was incorrect. The software was set to 1000 ticks/rev, but the actual encoders have 1024 ticks/rev.

The head board also incorrectly initialized the Index pin for the encoder, which would add some extra steps periodically.

The Head change is tested on hardware. The gantry change is being merged as-is, but we have high confidence due to the magnitude of encoder errors with the old values.


Laura-Danielle (approved):
sweet

PR #511, opened 2022-11-23
fix(bootloader) account for the fact that the bootloader is unable to read messages of size 60

ryanthecoder:
the HAl code skips sizes from 48 to 64. and defaults to size=64 when its not one of the defined values in length_from_hal in bootloader/core/utils.c

so this sets the size back to 64 and adjusts the tests accordingly


Laura-Danielle (approved):
lgtm

PR #512, opened 2022-11-28
fix(pipettes): 96 channel enable write pin was using the wrong gpio

Laura-Danielle:
Missed a place to change the enable write pin. Tested on a 96 channel prototype.


sfoster1 (approved):
lgtm

PR #514, opened 2022-11-29
fix(pipettes): make pick up motor timer half the time of the plunger motor

Laura-Danielle:
## Overview

The 96 channel has two timer interrupts to control the 3 stepper motors in the system. The plunger motor timer frequency was bumped up to match the motor frequencies in the rest of the robot system, but the gear motor wasn't changed to accomodate this. As a result, the firmware application was getting overloaded and crashing when both timers were enabled. Hopefully in January we'll be able to do a full study of the system (including freertos tasks/interrupts etc) to figure out where we can better optimize our system.


sfoster1 (approved):
Looks good to me!

PR #515, opened 2022-11-29
feat(motor-control): add StallCheck class

fsinapi:
This PR adds a class for checking the stall status of a stepper motor. The class is not integrated into the motor interrupt in this PR, it is just tested with CI.

Given the ratio of steps : µm for both the encoder and the stepper (using any microstepping ratio), the class keeps a running count of the position of the stepper motor vs the expected position of the encoder. The functions that are expected to be called from interrupts specifically avoid using any floating-point math or any multiplication in an effort to keep their overhead low.

The StallCheck object is set to be able to work with axes without encoders by just never indicating a bad status.

### DIFF
--- a/include/motor-control/core/stall_check.hpp
+++ b/include/motor-control/core/stall_check.hpp
@@ -52,7 +52,7 @@ class StallCheck {
      * @return True if the motor position is OK, false if the encoder
      * indicates a stall occurred.
      */
-    [[nodiscard]] auto check_stall(int32_t encoder_steps) const -> bool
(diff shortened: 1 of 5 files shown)


sfoster1 (changes requested):
I'm pretty uncomfortable with the two versions of check_stall using math that's this far separated. Can we make them both devolve to the same check somehow?

sfoster1 (approved):
A concise way to handle the problem (although since the other stuff is named `itr`, maybe `check_stall` still wants to be)

PR #517, opened 2022-12-02
chore(.github): change cross compile builds to RelWithDebInfo

ryanthecoder:
the Debug builds are too big for some proto boards so build as RelWithDebInfo instead


fsinapi (approved):

sfoster1 (approved):
Looks good to me

PR #518, opened 2022-12-05
feat(motor-control): stall detection in stepper ISR

fsinapi:
This PR adds support for stall detection in the ISR for stepper motors. The stall threshold is set to 500µm on every axis for now, which seems to work from basic testing. The stall check is only run when the stepper motor travels the distance of the stall threshold (i.e. 500µm on any encoder-enabled axes), and is never executed for an axis with no encoder.

Tested on the X, Y, and Z-L axes of an OT-3 by verifying that, if an axis stalls out at its end of travel, the flag for `stepper_position_ok` gets cleared. It is only reenabled if the axis is homed.

### DIFF
--- a/gantry/firmware/interfaces_proto.cpp
+++ b/gantry/firmware/interfaces_proto.cpp
@@ -189,11 +189,17 @@ static motor_class::Motor motor{
                                       .max_acceleration = 2},
     motor_queue};
 
+static stall_check::StallCheck stallcheck(
+    utils::linear_motion_system_config().get_encoder_pulses_per_mm() / 1000.0,
+    utils::linear_motion_system_config().get_steps_per_mm() / 1000.0,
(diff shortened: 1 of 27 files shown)


sfoster1 (approved):
This provisionally looks good to me, but we should probably test more on the stall detection thresholds - they'll mean very different things depending on the axis's motor-to-effector ratio.

fsinapi:
> This provisionally looks good to me, but we should probably test more on the stall detection thresholds - they'll mean very different things depending on the axis's motor-to-effector ratio.

Per offline discussion, stall thresholds of 2 fullsteps worth of micrometers were added + tested on a robot.

ahiuchingau (approved):
Awesome!!

Curious about how we came up with 2 full steps as the threshold?

sfoster1 (approved):
Awesome!

PR #519, opened 2022-12-05
fix(can): fix motor-position-response serialization

fsinapi:
When #507 added the message index to the MotorPositionResponse message, the serialization was not updated correctly and therefore the hardware controller can never get the response.

Tested that this change lets the can_comm and can_mon scripts receive the MotorPositionResponse


Laura-Danielle (approved):
nice :)

PR #520, opened 2022-12-06
feat(motor-control): read brushed motor driver config

ahiuchingau:
This PR allows us to read the current duty cycle and the vref of the brushed motor via CAN.


Laura-Danielle (approved):
nice :)

PR #521, opened 2022-12-08
fix(pipettes): add sync_in to pipette motor configs

caila-marashaj:
Previously the `sync_in` pin assignment was under `SensorHardwareConfiguration`. Moving it to `HardwareConfig` lets the firmware read the correct pin and therefore move with the sync line as a stop condition.

### DIFF
--- a/pipettes/firmware/motor_configurations.cpp
+++ b/pipettes/firmware/motor_configurations.cpp
@@ -114,6 +114,11 @@ auto motor_configs::hardware_config_by_axis(TMC2130PipetteAxis which)
                      .active_setting = GPIO_PIN_SET},
                 // LED PIN B11, active setting low
                 .led = {},
+                .sync_in =
+                    {// NOLINTNEXTLINE(cppcoreguidelines-pro-type-cstyle-cast)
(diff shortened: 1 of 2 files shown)


Laura-Danielle (approved):
Make sure the sync in line is the same for all pipette types, otherwise I'M SO EXCITEDDDDDDDDDDDDDD

YAYAYAYAYAY

pipettes/firmware/motor_configurations.cpp
```diff
@@ -213,6 +228,11 @@ auto motor_configs::hardware_config_by_axis(TMC2160PipetteAxis which)
                      .active_setting = GPIO_PIN_SET},
                 // LED PIN C11, active setting low
                 .led = {},
+                .sync_in =
+                    {// NOLINTNEXTLINE(cppcoreguidelines-pro-type-cstyle-cast)
+                     .port = GPIOD,
```
caila-marashaj, line 233: I made the `sync_in` pin D2 for the left gear motor, right gear motor, and linear motor. Is this what we want to do?

PR #522, opened 2022-12-08
fix(pipettes): change comparison to pascals, add sync_out configurations back

caila-marashaj:
This code changes the comparison for reaching a threshold from cm of water to pascals. It converts the pressure reading from the sensor to pascals, and then does the comparison to the threshold value.

A couple possible concerns though:
  - this assumes the threshold sent from Python is in Pascals
  - this code now leaves the value stored in `_registers.pressure.reading` unused, and instead uses the value stored in the private variable in `pressure_pascals`  

**Update:**
 in a [previous pr](https://github.com/Opentrons/ot3-firmware/pull/485), I removed the `sync_out` pin configuration from the sensor hardware configurations mistakenly, so I'm adding that back in this pr

### DIFF
--- a/include/sensors/core/tasks/pressure_driver.hpp
+++ b/include/sensors/core/tasks/pressure_driver.hpp
@@ -173,6 +173,9 @@ class MMR920C04 {
     }
 
     auto send_pressure(uint32_t message_index) -> void {
+        // Pressure is sent via CAN in pascals
+        auto pressure_pascals =
+            mmr920C04::Pressure::to_pressure(_registers.pressure.reading);
         auto pressure_fixed_point =
             convert_to_fixed_point(pressure_pascals, S15Q16_RADIX);
         auto message = can::messages::ReadFromSensorResponse{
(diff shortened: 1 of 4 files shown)


sfoster1 (approved):
I think this is fine, because
- yes, you have to get the units right. let's document that
- we do actually store the data in the register, we're just not using it at least right now

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -322,7 +320,8 @@ class MMR920C04 {
     uint16_t number_of_reads = 0x1;
     // TODO(fs, 2022-11-11): Need to figure out a realistic threshold. Pretty
     // sure this is an arbitrarily large number to enable continuous reads.
-    float threshold_cmH20 = 2457.0F;
+    float threshold_pascals = 100.0F;
+    float pressure_pascals = 0.0F;
```
Laura-Danielle, line 324: I don't think we should have this be a private variable on the pressure driver class.

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -173,10 +173,8 @@ class MMR920C04 {
     }
 
     auto send_pressure(uint32_t message_index) -> void {
-        auto pressure =
-            mmr920C04::Pressure::to_pressure(_registers.pressure.reading);
         auto pressure_fixed_point =
-            convert_to_fixed_point(pressure, S15Q16_RADIX);
+            convert_to_fixed_point(pressure_pascals, S15Q16_RADIX);
```
Laura-Danielle, line 177: can we put some comments in the send pressure functions that we are sending over pascals? We should probably also document [this in confluence](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3513712720/Pressure+Sensor).

Laura-Danielle (changes requested):
In general, I agree with the change. Have two tiny requests and then I will approve.

Laura-Danielle (approved):
lgtm

PR #523, opened 2022-12-12
feat(motor-control): add command to estimate stepper position from encoder

fsinapi:
This PR adds a command that allows each motor axis with an encoder to estimate its position from the current encoder value. This PR should be merged before the matching [monorepo PR](https://github.com/Opentrons/opentrons/pull/11868).

The Motor Interrupt Driver maintains the state of the motor controller, so the message must be handled by it. However, if there's an active move, it is important to prioritize the handling of that movement over the message; a priority queue would be ideal to solve this, but in lieu of that we simply add a second queue (and therefore have to touch a million source files). This secondary queue is only read if the motor ISR has NO pending move messages and is NOT currently handling a move.

The message handling logic is as follows:
* If the motor is currently disabled, the message isn't passed to the Interrupt Handler and is instead ack'd with the current motor position data (which is unchanged)
* When the motor interrupt handler receives the message:
  * If the Encoder Position OK flag is not set, don't change anything
  * If the flag _is_ set, reset the current motor position to be an estimate from the encoder position, and **set** the flag that the Motor Position is ok

Tested the following on Gantry and Head boards, monitoring responses with `can_mon`:
* Home the axis, and then send a move_rel command to change the position
* Send `UpdateMotorPositionRequest`, verify that the position flag remains set and the position is the same (or changes by a small number of µm due to differing resolution)
* Disable the axis, then send `UpdateMotorPositionRequest`, verify the `motor position ok` flag remains **unset** in the response
* Enable the axis, then send `UpdateMotorPositionRequest`, verify the `motor position ok` flag is now **set** and the motor position approximately matches the encoder position.


include/motor-control/core/stepper_motor/motion_controller.hpp
```diff
@@ -206,6 +219,12 @@ class PipetteMotionController {
         queue.try_write(msg);
     }
 
+    [[nodiscard]] auto update_position(
```
sfoster1, line 222: IMO to keep things consistent, we should treat this the same as the "encoder-position-ok-flag not set" case

sfoster1 (changes requested):
This is in general good but I think the gera motor stuff should just be the exact same as the other axes, but the encoder position is never good.

I really like the position update response approach. I think sending back an unchanged state is an appropriate way to say "yeah i'm not gonna do that" for the encoder-not-valid case. However, I think it's not a great idea for the "you-sent-this-while-i-was-moving" case - that's a straight up programmer error or unexpected system behavior, and IMO it should be an error response.

sfoster1 (approved):
Looks good to me!

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
```diff
@@ -110,11 +114,13 @@ class MotorInterruptHandler {
          * This function is called from a timer interrupt. See
          * `motor_hardware.cpp`.
          */
-        if (!has_active_move && has_messages()) {
+        if (!has_active_move && has_move_messages()) {
             update_move();
+            handle_update_position_queue_error();
```
Laura-Danielle, line 119: how much time does this check add to the work done in the isr?

include/can/core/messages.hpp
```diff
@@ -465,6 +468,27 @@ struct MotorPositionResponse : BaseMessage<MessageId::motor_position_response> {
     auto operator==(const MotorPositionResponse& other) const -> bool = default;
 };
 
+// This response is the exact same payload as MotorPositionResponse
+struct UpdateMotorPositionResponse
```
Laura-Danielle, line 472: nit you can ignore. As I understand it, this message is for the instances where you might want to read the current position without having issued a `Move`. Could we instead call this something like `GetCurrentMotorPosition` or something to that effect?

Laura-Danielle (approved):
In general, looks great to me. I am curious how much extra "work" this could potentially add in the ISR.

include/can/core/messages.hpp
fsinapi: The `UpdateMotorPositionRequest` command is used to update the position tracker in the firmware, and this is the response to it. I don't think that `GetCurrentMotorPosition` is an accurate representation of what the command does, and the functionality you're describing exists as `MotorPositionRequest`. 

Disambiguating this command by naming it `UpdateMotorPositionFromEncoder` would make sense to me!

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
fsinapi: I don't have a quantitative answer, but this is a relatively low-overhead thing to add in because it doesn't have any hardware interaction, most of the time is interacting with FreeRTOS queues. Tested running a home command and then sending an UpdateMotorPositionRequest during the movement on the Head board with both interrupts active and the result was as expected, an error message came back and the motor interrupt kept running as usual.

fsinapi:
Rebased with latest Main and tested on a robot

PR #524, opened 2022-12-14
fix(gripper): do not engage z-motor at boot

ahiuchingau:
We've added a brake to the z-motor, so we don't need to engage the motor at boot anymore.


Laura-Danielle (commented):
This is true for EVT grips?

ahiuchingau:
ya!

sfoster1 (approved):
There's no good grip emoji

PR #525, opened 2022-12-15
fix(pipettes): add position flags and action type to TipActionResponse

Laura-Danielle:
# Overview
The TipActionResponse was missing a few attributes in the message which caused it to fail unpacking the response on the python side.


fsinapi (approved):
We should probably have a test for this message in `can/tests/test_messages.cpp` but LGTM

PR #526, opened 2022-12-15
Check against commit in opentrons-emulation

DerekMaggio:


DerekMaggio:
Started getting failures after triggering the GH actions to run again. 
Need to diagnose before I put up for review.

PR #528, opened 2022-12-19
chore(all): enable codecoverage

fsinapi:
This PR adds code coverage tracking to the firmware and includes a few refactors to the cmake setup:
* Changed the following into dedicated static libraries:
  * i2c
  * spi
  * version.c (this one is required because of version.c being shared between common-core and bootloader-core)
* Removed extraneous instances of `find_package(Clang)` for cleaner output during configuration stage
* Added lcov coverage generation for any relevant targets
  * This does NOT include a lot of the `core` folders. The code in here depends on FreeRTOS and therefore isn't covered by tests (and realistically will not be covered), but if we think it is worthwhile then it would be possible to add this to the test coverage.
 * Added a CI job to generate coverage reports on PR's


sfoster1 (approved):
Looks good to me!

PR #531, opened 2022-12-30
feat(pipettes): add specific can messages for 96 channel tip motors

Laura-Danielle:
# Overview

Rather than having a specific node id for the tip motors on the 96 channel, I decided to create specific CAN messages for certain motor functionalities. Unlike the gripper, the tip motors are pretty intrinsic to the rest of the pipetting functionality and these CAN messages are only used in very specific instances.

See the python side of this change [here](https://github.com/Opentrons/opentrons/pull/11934).


include/motor-control/core/tasks/gear_tmc2160_motor_driver_task.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+namespace gear {
+
+using TaskMessage = tmc::tasks::GearTaskMessage;
+/**
+ * The handler of motor driver messages
+ */
+template <class Writer, can::message_writer_task::TaskClient CanClient,
+          class TaskQueue>
+class MotorDriverMessageHandler {
```
ahiuchingau, line 26: Could you use inheritance here to make use of the original 2160 driver task behaviors to avoid repeating code?

sfoster1: unfortunately no because then it would accept the other messages. you could make a single class that is templated with the kinds of messages to accept but that's kind of a lot.

sfoster1 (approved):
Looks good structurally but we need some tests for this I think.

PR #532, opened 2023-01-03
fix(coverage): run coverage on pushes to main

fsinapi:
Coverage wasn't being generated on the main branch because the codecov action was configured to run on pushes to `edge`, but this repository uses `main`.


sfoster1 (approved):
👍

PR #533, opened 2023-01-05
feature(gripper): add detection for dropping labware or collision detection

ryanthecoder:
This PR watches the gripper g encoder during force control moves (when homed or gripping something) and if there is unexpected movement we report an error
https://opentrons.atlassian.net/browse/RLAB-74
https://opentrons.atlassian.net/browse/RLAB-150
https://opentrons.atlassian.net/browse/RLAB-151

### DIFF
--- a/include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
+++ b/include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
@@ -117,7 +117,8 @@ class BrushedMotorInterruptHandler {
                 break;
             // linear move
             case MoveStopCondition::encoder_position:
-                move_delta = hardware.get_encoder_pulses() - buffered_move.encoder_position;
+                move_delta = hardware.get_encoder_pulses() -
+                             buffered_move.encoder_position;
                 controlled_move_to(move_delta);
                 if (std::abs(move_delta) < acceptable_position_error) {
                     finish_current_move(AckMessageId::stopped_by_condition);
@@ -142,10 +143,13 @@ class BrushedMotorInterruptHandler {
         if (has_messages()) {
             update_and_start_move();
         } else if (motor_state == ControlState::POSITION_CONTROLLING) {
-            int32_t move_delta = hardware.get_encoder_pulses() - hold_encoder_position;
+            int32_t move_delta =
+                hardware.get_encoder_pulses() - hold_encoder_position;
             controlled_move_to(move_delta);
-            // we use double the acceptable position here just to allow the pid loop the oppertunity to maintain small movements that occure from motion and vibration
-            if (move_delta > 2*acceptable_position_error) {
+            // we use double the acceptable position here just to allow the pid
+            // loop the oppertunity to maintain small movements that occure from
+            // motion and vibration
+            if (move_delta > 2 * acceptable_position_error) {
                 cancel_and_clear_moves(can::ids::ErrorCode::collision_detected);
             }
         } else if (motor_state == ControlState::FORCE_CONTROLLING_GRIP ||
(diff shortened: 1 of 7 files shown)


ahiuchingau (approved):
Nice!!

Curious about the behavior on the robot - what happens if a collision happens while the gripper is transferring a labware (and the labware doesn't drop)?

ryanthecoder:
> n
Right now if it hits something hard enough with the labware still being gripped to trigger it, it will say labware dropped even if it isn't

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -99,10 +99,10 @@ class BrushedMotorInterruptHandler {
         if (old_control_pwm != current_control_pwm) {
             driver_hardware.update_pwm_settings(current_control_pwm);
         }
-        return move_delta;
     }
 
     void execute_active_move() {
+        int32_t move_delta = 0;
```
sfoster1, line 105: move this into the case at line 120? otherwise it's unused, and it just feels very k&r up here

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -140,7 +141,26 @@ class BrushedMotorInterruptHandler {
         if (has_messages()) {
             update_and_start_move();
         } else if (motor_state == ControlState::POSITION_CONTROLLING) {
-            controlled_move_to(hold_encoder_position);
+            int32_t move_delta =
+                hardware.get_encoder_pulses() - hold_encoder_position;
+            controlled_move_to(move_delta);
+            // we use double the acceptable position here just to allow the pid
+            // loop the oppertunity to maintain small movements that occure from
```
sfoster1, line 148:
```suggestion
            // loop the opportunity to maintain small movements that occur from
```

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -140,7 +141,26 @@ class BrushedMotorInterruptHandler {
...
+            // motion and vibration
+            if (move_delta > 2 * acceptable_position_error) {
```
sfoster1, line 150: maybe make this a named constant? it's easy to miss the digit in all the variable names.

sfoster1 (changes requested):
kind of minor, but up until now we've done a pretty good job of keeping the word "grip" out of the brushed motion controller, which nominally is independent of the gripper itself. now we've added the `[get,set]_stay_gripping` functions; a `FORCE_CONTROLLING_GRIPPING` state; and a labware-dropped error. Is there any way that we can rephrase these to keep them generic?

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
ryanthecoder: Complier complains that it "Jumps the case statement" if you put it down there. Bothers me too

ryanthecoder:
removed as many references to grip/ungrip that were low hanging fruit, I think we should make another ticket to fully generalize this since I think refactoring out the labware dropped into a gripper specific file will bloat this PR

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
sfoster1: needs braces in the case statement if you're declaring a variable:
```
case MoveStopCondition::encoder_position: {
int32_t move_delta //...
break;
}

File: include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp [hunk 4]
```diff
@@ -140,7 +145,26 @@ class BrushedMotorInterruptHandler {
         if (has_messages()) {
             update_and_start_move();
         } else if (motor_state == ControlState::POSITION_CONTROLLING) {
-            controlled_move_to(hold_encoder_position);
+            int32_t move_delta =
+                hardware.get_encoder_pulses() - hold_encoder_position;
+            controlled_move_to(move_delta);
+            // we use double the acceptable position here just to allow the pid
```
sfoster1, line 151: this comment's out of date now right?

sfoster1 (approved):
Couple more little things, but looks solid!

PR #534, opened 2023-01-05
fix(motor-control): reduce GPIO write operations during stepper interrupts

fsinapi:
This PR reduces the overhead of the stepper motor interrupt on the OT-3 firmware boards. In particular, the head board was struggling with the overhead of the two stepper interrupts resource-starving all of the FreeRTOS tasks, including the highest-priority watchdog task. 

Tested cycling a head board's motors up and down repeatedly. Before this change, the status LED would not be blinking during movements and the watchdog would usually expire. After the change, the board functions as expected.


sfoster1 (approved):
Looks good. I wonder why those calls are so slow - I have a sad suspicion it's the virtualization and we may need to do something a bit drastic (and template based) to recover it

Laura-Danielle (approved):
nioce

PR #535, opened 2023-01-06
fix(pipettes): use the correct gpio pins on rev b single/multi boards

ryanthecoder:


sfoster1 (approved):
Cool. We have all rev b boards in house right?

ryanthecoder:
We have some protos around still  (which is what I used to test) and rev c for single/multi is on its way but they don't change the pins

single/multi rev a | PA9
single/multi rev b | PC12
single/multi rev c1 | PC12
single/multi rev c2 | PC12
96 A | pa 9
96 b | pb9

PR #537, opened 2023-01-09
feat(motor-control): add ability to terminate movements from stalls

fsinapi:
This PR adds the ability for the motor controller to terminate a movement and send an error if a stall is detected. This is paired with a [monorepo PR](https://github.com/Opentrons/opentrons/pull/11967).

When the stepper motor controller clears the Stepper Position OK flag, it will now check if the end condition of the movement is the Stall condition. If it is, an error is sent up to the host and the move queue is cleared. 

This PR also moves the MoveStopCondition to be an autogenerated enum from the monorepo.

Tested on robot that, if an axis is forced to stall while moving, it detects the condition and bubbles up an error _only_ if the end condition of the movement is `stall`. Also verified that cycling the axes at high speed does not cause spurious stall events.


sfoster1 (approved):
Looks good, but I'm a little worried by how `gripper_stall` wasn't checked for before... is that condition not getting loaded in to the interrupt message?

PR #539, opened 2023-01-09
refactor(motor-control): parameterize motor hardware in stepper interrupt

fsinapi:
In our ongoing quest to reduce the overhead of the motor interrupts, virtual calls seems like a good target. By adding a template parameter for the type used for the motor hardware in the stepper interrupt, the compiler is now able to optimize out some virtual hardware calls.

I tested that this at least doesn't slow down the interrupts by running it on a Head unit, but I didn't quantify the difference. However, I checked the disassembly with `objdump` and gcc is definitely calling the hardware functions directly (and was using virtual calls before), so I'm 99% confident that this is an improvement (even if it's only marginal).

There are other places we could do the same thing with a bit more work, but this was pretty low-hanging fruit.

### DIFF
--- a/include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
+++ b/include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
@@ -39,8 +39,9 @@ using namespace motor_messages;
  */
 
 template <template <class> class QueueImpl, class StatusClient,
-          typename MotorMoveMessage>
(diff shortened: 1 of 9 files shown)


Laura-Danielle (approved):
lgtm!

sfoster1 (approved):
Oh hell yes that's awesome i was having nightmares about CRTP but this is a way easier way to devirtualize

PR #540, opened 2023-01-10
feat(motor-control): update all stall thresholds to 0.5mm

fsinapi:
Per discussions with hardware, we want the default stall threshold for the motors to always be 0.5mm. This has the nice side effect of making the initialization code look really clean and simple.


sfoster1 (approved):
5️⃣

PR #542, opened 2023-01-11
fix(motor-control): don't send a Stop Request after stalling

fsinapi:
In one of the primary use cases for stall detection (tip force pickup), we don't want to send a Stop Request to all of the other axes because that will disable each motor and invalidate motor positions. This PR removes the broadcast after a stall is detected - if a full gantry stop makes sense, it can be sent by the hardware controller.


sfoster1 (approved):
Yeah that sounds correct. Nice.

Laura-Danielle (approved):
makes sense to me

PR #544, opened 2023-01-12
refactor: use logic voltages for mount detect

sfoster1:
In EVT and previous, we had a scheme (or set of frequently-changing schemes) to have a passives network establish an appropriate voltage for a given (pipette type, attach point) pair that could be read by both the attached pipette and the head board. This kept having problems with rise time, consistency between assembled units, the quality of the onboard ADCs, and so on.

Instead, we'll eventually have a system where the head sets its mount id pins to input with internal pulls. This establishes a voltage the pipettes can see; they'll read it, and then they'll drive the pin to the opposite voltage, which the head can detect to indicate pipette presence.

However, this scheme won't work with EVT pipettes or head boards because of the presence of external pulls which are stronger than the internal pulls. That means the head board has to drive its detection pins directly, and _that_ means the pipettes have to keep their mount ID pins on input so we're not driving current between two GPIOs. To get the above scheme to work, either both sides need to be DVTs or they need to be modified to remove the external pulls.

For now, and what's in this PR, what we're going to do is drive the pins from the head and keep the pipette pins inputs. This removes the ability for the head to detect the presence of a pipette, so we're going to keep this change in a branch sent to the factory to do QC of DVT pipettes and not distribute it further. We'll add, in a separate branch, the new behavior that allows detection again; and eventually, when we get more DVT hardware we'll either move away from the remaining EVT hardware or modify it to work with the new scheme.

What we have now is:
- The head drives its left/z mount id high
- The head drives its right/a mount id low
- The pipettes sense this and assign their ids from it
Note: The above is switched from the first push of this PR because driving the voltage high on the left/z mount of EVT systems results in an unusable intermediate voltage because of the values of the resistor divider for the old analog system. 
Test cases:
- [x] Pipette plugged into left at system boot
- [x] Pipette plugged into right at system boot
- [x] Pipette plugged into left after system boot
- [x] Pipette plugged into right after system boot
- [x] Pipette sent to bootloader on left
- [x] Pipette sent to bootloader on right

Closes RET-1297


bootloader/core/pipette_node_id.c
```diff
@@ -1,13 +1,6 @@
...
-#define A_CARRIER_MAX_VOLTAGE_MV (2900)
-
-
 CANNodeId determine_pipette_node_id(
-    uint16_t mount_voltage_mv) {
-    if ((mount_voltage_mv > A_CARRIER_MIN_VOLTAGE_MV) && (mount_voltage_mv < A_CARRIER_MAX_VOLTAGE_MV)) {
-        return can_nodeid_pipette_right_bootloader;
-    }
-    return can_nodeid_pipette_left_bootloader;
+    bool mount_id) {
```
Laura-Danielle, line 4: nit: could we call this `mount_id_left` since the boolean is True only if it's a `left_pip`

bootloader/firmware/stm32L5/node_id_stm32l5xx.c
```diff
@@ -5,92 +5,18 @@
...
-    return ((((uint32_t)adc_raw_reading) * ADC_REF_MV) / ADC_MAX_VAL);
-}
-
 
 static CANNodeId update_dynamic_nodeid() {
-    if (!adc_init()) {
-        return can_nodeid_pipette_left_bootloader;
-    }
+    // B1: mount id
+    __HAL_RCC_GPIOB_CLK_ENABLE();
```
Laura-Danielle, line 12: I don't even know if I would bother changing this file. Pre-EVT pipettes aren't compatible with today's robot, so the likelihood that they would be attached to an EVT/DVT head board is pretty much zero.

head/core/tasks_proto.cpp
```diff
@@ -82,16 +76,11 @@ void head_tasks::start_tasks(
     auto& spi3_task = spi3_task_builder.start(5, "spi bus 3", spi3_device);
     spi3_task_client.set_queue(&spi3_task.get_queue());
 
-    auto& presence_sensing = presence_sensing_driver_task_builder.start(
```
Laura-Danielle, line 85: shouldn't we keep presence sensing on pre-EVT head boards if it's not too much of a pain?

Sidenote, I feel like we're getting pretty close to where we can delete proto code for a bunch of stuff.

pipettes/firmware/utility_gpio.c
```diff
@@ -120,7 +120,7 @@ void LED_drive_gpio_init() {
     HAL_GPIO_Init(hardware.port, &GPIO_InitStruct);
 }
 
-void sync_drive_gpio_init() {
+static void sync_drive_gpio_init() {
```
Laura-Danielle, line 123: What was the purpose of making these static?

Laura-Danielle (approved):
lgtm. imo, I feel like we should just delete the proto code rather than change it to match with DVT behavior since the likelihood that proto boards + dvt units would be used together is basically zero.

You can choose to not do that in this PR though.

head/core/tasks_proto.cpp
sfoster1: Well, maybe... but I'm more inclined to agree with the second statement (for a different PR since this one is so time-constrained)

sfoster1: Also presence sensing will have to look really different anyway

pipettes/firmware/utility_gpio.c
sfoster1: Same as making stuff `const` in cpp and other languages - in c, whatever functions can be static should be static. static applied to a function definition in c means that it's not accessible from other translation units, which means it doesn't pollute the global namespace

sfoster1:
Yeah this PR is a little time constrained but the followup that re-adds how tool detection should work should maybe also just delete the proto code.

sfoster1:
In EVT and previous, we had a scheme (or set of frequently-changing schemes) to have a passives network establish an appropriate voltage for a given (pipette type, attach point) pair that could be read by both the attached pipette and the head board. This kept having problems with rise time, consistency between assembled units, the quality of the onboard ADCs, and so on.

Instead, we'll eventually have a system where the head sets its mount id pins to input with internal pulls. This establishes a voltage the pipettes can see; they'll read it, and then they'll drive the pin to the opposite voltage, which the head can detect to indicate pipette presence.

However, this scheme won't work with EVT pipettes or head boards because of the presence of external pulls which are stronger than the internal pulls. That means the head board has to drive its detection pins directly, and that means the pipettes have to keep their mount ID pins on input so we're not driving current between two GPIOs. To get the above scheme to work, either both sides need to be DVTs or they need to be modified to remove the external pulls.

For now, and what's in this PR, what we're going to do is drive the pins from the head and keep the pipette pins inputs. This removes the ability for the head to detect the presence of a pipette, so we're going to keep this change in a branch sent to the factory to do QC of DVT pipettes and not distribute it further. We'll add, in a separate branch, the new behavior that allows detection again; and eventually, when we get more DVT hardware we'll either move away from the remaining EVT hardware or modify it to work with the new scheme.

What we have now is:

    The head drives its left/z mount id high
    The head drives its right/a mount id low
    The pipettes sense this and assign their ids from it
    Note: The above is switched from the first push of this PR because driving the voltage high on the left/z mount of EVT systems results in an unusable intermediate voltage because of the values of the resistor divider for the old analog system.
    Test cases:
    Pipette plugged into left at system boot
    Pipette plugged into right at system boot
    Pipette plugged into left after system boot
    Pipette plugged into right after system boot
    Pipette sent to bootloader on left
    Pipette sent to bootloader on right

Closes RET-1297

PR #546, opened 2023-01-16
chore(build): add preset to build all binary hex files for use in oe-core builds.

vegano1:
# Overview

In order to update the instruments on the OT3, we need to include binary hex files in the filesystem as part of the ot-core system build. This will add a preset to build all hex binary files as well as a new script to compile the subsystem versions into a json file for the convenience of the oe-core build.

Part of [RCORE-168](https://opentrons.atlassian.net/browse/RCORE-168)

# Changes

- Add preset "all" that will build all binary hex files intended to be included in the ot3 system build, right now this points to rev1 but might change later.
- Adds a script/subsystem_versions.py script that generates a json file with the version, shortsha, and branch for all subsystems.

# Test Plan

- [x] generate a single target
- [x] generate multiple targets
- [x] generate all targets
- [x] print output to stdout
- [x] save output to filepath given by --output arg

# Review Requests

- [x] Does this make sense?
- [x] Do we want to use rev1 binary hex files or something else?

# Risk Assessment

Low, OT3 dev only change.

[RCORE-168]: https://opentrons.atlassian.net/browse/RCORE-168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfoster1 (approved):
nice!

CMakePresets.json
```diff
@@ -77,6 +77,23 @@
     }
   ],
   "buildPresets": [
+    {
+      "name": "all",
```
sfoster1, line 81: sort of a nit but I don't love this name - it really feels like it should build _everything_, but it doesn't and can't - doesn't build tests, doesn't build simulators, doesn't build image hexes, etc. It just builds the application firmwares. Maybe call it `all-application-firmware` or something?

scripts/subsystem_versions.py
```diff
@@ -0,0 +1,69 @@
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Creates a json file of the submodule version info.")
+    parser.add_argument(
+        "--target",
+        default="all",
+        nargs="+",
+        choices=SUBSYSTEMS,
+        help="subsystem to generate file for; leave blank for all.",
+    )
+    parser.add_argument(
```
sfoster1, line 64: typically what you want to do with an output arg like this is make it a filetype: `parser.add_argument('--output`, type=argparse.FileType('w'), default=sys.stdout)`. What this does is automatically try and open the specified file and give you an io object that you can write to, with a default of stdout which is treated the exact same as a file.

scripts/subsystem_versions.py
```diff
@@ -0,0 +1,69 @@
...
+            json.dump(version_info, fh)
+    return version_info
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Creates a json file of the submodule version info.")
+    parser.add_argument(
+        "--target",
+        default="all",
+        nargs="+",
```
sfoster1, line 60: IMO a more natural way to do this is `nargs="*"` and have the logic in `main()` look like `subsystems = args.target or SUBSYSTEMS`.

scripts/subsystem_versions.py
```diff
@@ -0,0 +1,69 @@
...
+    "pippetes-384",
+]
+
+
+def main(args):
+    version_info = {}
+    for subsystem in SUBSYSTEMS:
+        if args.target == "all" or subsystem in args.target:
+            try:
+                version = (
```
sfoster1, line 27: i get that this is more extensible but we really are building everything off the same commit, so let's hoist this logic out of the loop to make it faster?

sfoster1 (approved):
This looks like it will work but I have a lot of nits with it; I think we can make the script and cmake targets a little more ergonomic. I'd also like to see manifest generation become a cmake target itself, but I would be OK with merging this as is due to time pressure.

scripts/subsystem_versions.py
```diff
@@ -0,0 +1,69 @@
...
+                    subprocess.check_output(
+                        ["git", "describe", "--tags", "--always", "--match=v*"]
+                    )
+                    .decode()
+                    .strip()
+                )
+                # version should be int
+                version = int(re.search(VERSION_REGEX, version).group(1)) if re.search(VERSION_REGEX, version) else 0
```
Laura-Danielle, line 35: We could probably pull this into a function that passes in the target we're looking for.

That way, you could use the function to build a version for a single target or for all the subsystems.

Laura-Danielle (approved):
Agree with seth's comments and added a few of my own. Once those are addressed I'm good for this to be merged.

CMakePresets.json
```diff
@@ -77,6 +77,23 @@
...
+      "jobs": 4,
+      "targets": [
+        "head-rev1-hex",
+        "gantry-x-rev1-hex",
+        "gantry-y-rev1-hex",
+        "gripper-rev1-hex",
+        "pipettes-single-rev1-hex",
+        "pipettes-multi-rev1-hex",
+        "pipettes-96-rev1-hex",
+        "pipettes-384-rev1-hex"
```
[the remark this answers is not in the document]
vegano1: will keep so we dont have to worry about it for testing.

vegano1:
@sfoster1 @Laura-Danielle 
Thanks for the feedback, should be much cleaner now.

scripts/subsystem_versions.py
```diff
@@ -0,0 +1,68 @@
...
+    subsystems = args.target or SUBSYSTEMS
+
+    try:
+        version = subprocess.check_output(["git", "describe", "--tags", "--always", "--match=v*"]).decode().strip()
+        # version should be int
+        version = int(re.search(VERSION_REGEX, version).group(1)) if re.search(VERSION_REGEX, version) else 0
+        shortsha = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode().strip()
+        branch = subprocess.check_output(["git", "rev-parse", "--abbrev-ref", "HEAD"]).decode().strip()
+    except Exception as e:
+        print(f"Could not get the version info for {subsystem}", file=sys.stderr)
```
fsinapi, line 33: `subsystem` isn't initialized at this point

sfoster1 (approved):
Looks good to me!

scripts/subsystem_versions.py
vegano1: good catch, leftover from before refactor.

fsinapi (approved):

PR #548, opened 2023-01-17
refactor(pipettes): Only monitor for a TipActionRequest in the move motor task and add a motor id

Laura-Danielle:
## Overview

Previously, the gear move and motion tasks were both looking for TipActionRequests on the CAN dispatcher. This meant that any time you sent a tip action request, it would send back four responses rather than two responses.

In addition, I added a "GearMotorId" to help determine which motor is sending back the response. Required changes on the python side can be found [here](https://github.com/Opentrons/opentrons/pull/12018).

See RLIQ-295 for more information.

# Test Plan

- [x] Test 96 channel tip motor movement with the diagnostic script
- [x] Check that this fixes tip pick up behavior in LPC where the 96 channel starts homing before pick up tip is completed.

### DIFF
--- a/can/tests/test_messages.cpp
+++ b/can/tests/test_messages.cpp
@@ -426,7 +426,7 @@ SCENARIO("message serializing works") {
     }
 
     GIVEN("a tip action response") {
-        constexpr uint8_t MESSAGE_SIZE = 18;
+        constexpr uint8_t MESSAGE_SIZE = 19;
         auto message =
             TipActionResponse{.message_index = 0x1234,
                               .group_id = 1,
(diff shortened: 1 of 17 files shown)


sfoster1 (approved):
This looks like it'll work, yeah. I'm a bit confused by why each move status reporter task seems to get the responses from both motors? Might be better to have that not happen instead.

Laura-Danielle:
> This looks like it'll work, yeah. I'm a bit confused by why each move status reporter task seems to get the responses from both motors? Might be better to have that not happen instead.

@sfoster1 

Because each motion controller task receives the request to do a motor move (which is correct behavior), the motor interrupt will be handling two `GearMotorMoves`. Since each queue class has a move status reporter, and both are looking for GearAcks, then they would both respond with each GearAck they saw.

I definitely want to keep the motor ids since we have no other way of determining which response came from which motor. I could try to refactor the code so that there is only one move status reporter. Let me know what you think.

sfoster1 (approved):
Looks good to me now, and works. Let's just make sure to update the pr description and squash commit message to reflect what actually changed.

PR #550, opened 2023-01-19
refactor(head): enable motors on boot

Laura-Danielle:
## Overview

The 96 channel is extremely heavy. We have the ebrake to prevent the pipette from falling when the robot is off, however immediately on boot of the system the pipette starts falling. This means that we will need to enable the Z motor on boot. I decided to just enable both motors to make the behavior the same, but I am fine with only doing the mount that matters as well.


sfoster1 (approved):
do hate to have it all fall down

PR #551, opened 2023-01-19
refactor: handle the DVT mount/pipette ID scheme

sfoster1:
Implements the slot detection scheme for DVT.

# Overview
On EVT and previous, the head, mount, and pipette/gripper main boards were equipped with a set of pull resistors that when combined would give a unique voltage output for every tool and mount combination. This voltage could be read by ADCs on the pipettes to determine which mount they were connected to, and by ADCs on the head to determine what pipettes were connected where. This scheme mostly worked, but it was not robust to noise and passive component tolerances.

On DVT, we made the system much more reliable by removing the various pull up and pull down resistors and switching to an active signalling scheme. Now,
- The head configures its mount ID pins as inputs, with a pull up for the A mount and a pull down for the Z and gripper mounts
- Tools initially configure their mount ID pins as inputs, and read the voltage caused by the internal pulls on the head as digital values to determine what mount they're connected to and set their node IDs
- After a tool sets its node ID, it switches its mount pin to an output and drives the line to the opposite of what the head pulled it to
- The head, because its mount ID pins are inputs, knows that a tool is present when the mount ID line has the opposite value of its configured internal pulls

This allows a lot of simplification of code and reduction of time spent waiting for line voltages to stabilize.

The gripper is pretty much the same, but it always knows its node ID so it can just always drive its mount ID pin high.

# Changelog and affected devices
- Head (+bootloader), all pipettes (+bootloaders), gripper (+bootloader) are affected by this change
- Remove tool id voltage lookup table and machinery
- Remove ADC machinery
- Reduce pipette settling time from 1500ms to 2ms
- Pipettes now read digital inputs to determine node IDs
- After pipettes and gripper determine node ID, they set the pin to tell the head they're present

# Test Plan
- [x] Connect a DVT pipette to both mounts of a DVT head and check that its node ID is correctly assigned and the head correctly detects it, both generating tool notifications and responding correctly to attached-tool requests; switch it to bootloader, and check that the bootloader gets the right node ID and notifications are generated
- [X] Connect a DVT pipette to an EVT head with reworked (to remove the pull resistor) mount boards and do the same tests
- [X] Connect a reworked (to remove the pull resistor) EVT pipette to both mounts of a DVT head board and do the same tests
- [X] Connect a reworked (to remove the pull resistor) EVT gripper to a DVT head board and do the same test
Closes RET-1298

# Note on merging

This change handles a hardware change going from EVT to DVT. It must not be merged until either all machines are DVT or all remaining EVT mount boards, pipette main boards, and gripper main boards have been reworked to remove the pull resistors. If not, the internal pulls on the head microcontroller will be dominated by the external pulls scattered around the system, and connected pipettes will not correctly determine their node ID.


include/head/core/presence_sensing_driver.hpp
```diff
@@ -5,22 +5,25 @@
...
-    PresenceSensingDriver(adc::BaseADC& adc,
-                          attached_tools::AttachedTools current_tools)
-        : adc_comms(adc), current_tools(current_tools) {}
-    auto get_readings() -> adc::MillivoltsReadings {
-        return adc_comms.get_voltages();
-    }
+    PresenceSensingDriver()
+        : PresenceSensingDriver(attached_tools::AttachedTools{}) {}
+    explicit PresenceSensingDriver(const PresenceSensingDriver&) = default;
+    auto operator=(const PresenceSensingDriver&)
```
Laura-Danielle, line 18: Do we want this class to be moveable/copy-able?

head/simulator/main.cpp
```diff
@@ -133,10 +132,8 @@ static motor_interrupt_driver::MotorInterruptDriver sim_interrupt_left(
     motor_queue_left, motor_interrupt_left, motor_interface_left,
     update_position_queue_left);
 
-static auto adc_comms = adc::SimADC{};
-
 static auto presence_sense_driver =
-    presence_sensing_driver::PresenceSensingDriver(adc_comms);
+    presence_sensing_driver::PresenceSensingSimulator();
```
Laura-Danielle, line 136: Should we default this to all instruments being attached rather than unattached? Otherwise I think the hardware controller will not return that instruments are attached if using the `OneShotDetector` class (or whatever it was called).

Laura-Danielle (approved):
Implementation looks good.

Two overall comments:
- Did we want to take the opportunity here to delete proto code from main?
- Do we still need the long delay before ID attaching? Would be great to reduce this time if not.

pipettes/firmware_l5/utility_gpio.c
```diff
@@ -94,7 +94,7 @@ void LED_drive_gpio_init() {
     HAL_GPIO_Init(hardware.port, &GPIO_InitStruct);
 }
 
-void sync_drive_gpio_init() {
+static void sync_drive_gpio_init() {
```
caila-marashaj, line 97: What's the advantage of making these functions static?

include/head/core/presence_sensing_driver.hpp
sfoster1: it doesn't really matter IMO it doesn't capture any references to things

head/simulator/main.cpp
sfoster1: it'll asplode if it's wrong though i think

sfoster1:
> Implementation looks good.
> 
> Two overall comments:
> 
>     * Did we want to take the opportunity here to delete proto code from main?
Not in this PR, this is pretty crucial functionality and I don't want to pollute it
>     * Do we still need the long delay before ID attaching? Would be great to reduce this time if not.
We need a delay but a much shorter delay, like 2ms, so I should change that

PR #552, opened 2023-01-20
chore: update poetry to 1.3.2

sfoster1:
This probably solves problems with installing both opetnrons and opentrons-shared-data.

# Test plan
- [x] run `cmake --preset=host` and `cmake --build --preset=simulators` and see if poetry explodes

# review requests
- do the test plan on your machine and see if it works


DerekMaggio:
Weird that this fixes it.

DerekMaggio (approved):

PR #553, opened 2023-01-23
refactor(all): Modify calculation for usteps per mm and update microstepping to 1/16th for tip motors

Laura-Danielle:
## Overview

The steps per mm calculation was slightly off for the tip motors. This is because we have been using two different types of gear ratios -- the gear reduction ratio and the regular gear ratio. The regular gear ratio is `driving gear teeth / driven gear teeth` whereas the reduction ratio is `larger teeth / smaller teeth`.

The current way we do the calculations worked out OK for the gripper motor because the reduction ratio is the inverse of the regular gear ratio (since the larger gear motor is the driven gear). However, logically, it makes a bit more sense to multiple these gear ratios by the leadscrew pitch (which represents mm per rev) because gear trains affect the mm per rev. 

In addition, when we changed the motor drivers for the tip motors to use tmc2160s, we did not change the microstepping to 1/16th. The issue hadn't been found thus far because the tip motors were also executing moves twice so it would appear to be moving mostly to the correct position.

### DIFF
--- a/gantry/firmware/interfaces_proto.cpp
+++ b/gantry/firmware/interfaces_proto.cpp
@@ -207,7 +207,7 @@ static motor_class::Motor motor{
 
 static stall_check::StallCheck stallcheck(
     utils::linear_motion_system_config().get_encoder_pulses_per_mm() / 1000.0F,
-    utils::linear_motion_system_config().get_steps_per_mm() / 1000.0F,
+    utils::linear_motion_system_config().get_usteps_per_mm() / 1000.0F,
     utils::STALL_THRESHOLD_UM);
 
 /**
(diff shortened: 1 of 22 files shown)


fsinapi (approved):
Nice. Can the PR title be edited to mention that it's 1/32 microstepping though? I think that would be clearer in the commit history.

sfoster1 (approved):
Heck yes

fsinapi (approved):
nice!

sfoster1 (approved):
Okay, this looks good to me - the idea is that none of the actual values should change aside from the tip pickup motor,s sounds good

caila-marashaj (approved):
nice work!

PR #554, opened 2023-01-25
refactor(motor-control): rename cap_sensor stop condition to sync_line

caila-marashaj:
# Overview
Before this, the moves created in the liquid sensing code used `MoveStopCondition.cap_sensor` to check the sync line during motion. Since there's no reason to differentiate between the two sensors for the purpose of the `MoveStopCondition` enum, both sensors will use the same condition, which will now be called `sync_line`.

# Changelog

- changed `MoveStopCondition.cap_sensor` to `MoveStopCondition.sync_line`
- combined `sync_triggered()` and `calibration_stopped()` to make naming a little less redundant
- moved sync line check out of nested `if` to appease the linter

### DIFF
--- a/include/can/core/ids.hpp
+++ b/include/can/core/ids.hpp
@@ -213,7 +213,7 @@ enum class MotorPositionFlags {
 enum class MoveStopCondition {
     none = 0x0,
     limit_switch = 0x1,
-    cap_sensor = 0x2,
+    sync_line = 0x2,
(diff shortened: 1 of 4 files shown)


Laura-Danielle (approved):
no cap ❌ 

Also pls just double check on a robot if you haven't already :)

sfoster1 (approved):
LGTM when everything's green!

PR #555, opened 2023-01-25
refactor(head): flip directions for dvt

sfoster1:
There is one fewer gear and therefore the direction is now different.

Closes RET-1314


Laura-Danielle (approved):
😎

PR #557, opened 2023-01-26
feat(rear-panel) add an ot 3 firmware project for the rear panel mcu

ryanthecoder:
https://opentrons.atlassian.net/browse/RET-1302
Initial bring up with blinking deck lights.

Create a new subdirectory for the rear-panel project

This new project is not connected to the can bus so it uses a different linking and updating code which was mostly copied from the tempdeck-gen3 and adjusted for the different chip package

This project is responsible for controlling the Deck and UI leds
watching the estop button and signaling the rest of the system,
connected to the door closed sensor
connect to aux devices like the hepa hood


include/rear-panel/core/tasks.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include "i2c/core/hardware_iface.hpp"
+#include "i2c/core/messages.hpp"
+#include "i2c/core/tasks/i2c_poller_task.hpp"
+#include "i2c/core/tasks/i2c_task.hpp"
+#include "i2c/core/writer.hpp"
+
+namespace rear_tasks {
+
+/**
+ * Start gripper tasks.
```
sfoster1, line 15: let's go ahead and clean up the copy/paste artifacts before doing the first commit

include/rear-panel/firmware/led_hardware.h
```diff
@@ -0,0 +1,37 @@
...
+	GREEN_UI_LED,
+	BLUE_UI_LED,
+	WHITE_UI_LED
+}LED_DEVICE;
+
+extern TIM_HandleTypeDef htim2;
+extern TIM_HandleTypeDef htim3;
+extern TIM_HandleTypeDef htim15;
+
+void update_pwm(uint32_t duty_cycle, LED_DEVICE led);
```
sfoster1, line 32: let's make these prefixed names like `led_hardware_update_pwm()` and `led_hardware_initialize_leds()`

sfoster1 (changes requested):
This looks good, but I think we need some cleanup work on it. Stuff like
- let's sanitize out the copy/pasted and often commented out code, or remove the comments and get it working
- let's add a readme to the rear panel subdirectory that talks about why the rear panel is so different, needing USB and so on, and how it doesn't use a lot of the common code
- let's add a more substantive PR description that will end up as the commit message in the squash commit

sfoster1 (approved):
Nice, thanks for the changes! We'll also need a github workflow to build this but we don't need to do that right now.

rear-panel/startup/application-startup/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+        DEPENDS ${TARGET}.bin)
+    
+    set_source_files_properties(${CMAKE_CURRENT_FUNCTION_LIST_DIR}/startup_main.c
+        PROPERTIES
+        OBJECT_DEPENDS ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/STM32_MODULE_STARTUP.ld)
+endfunction()
+
+# Incurs at least a relink when you change the linker file (and a recompile of main
+# but hopefully that's quick)
+set_source_files_properties(./module-startup.S
```
fsinapi, line 66: Can delete this

rear-panel/startup/application-startup/module-startup.S
```diff
@@ -0,0 +1,207 @@
+/**
+  ******************************************************************************
+  * @file   module-startup.S
```
fsinapi, line 3: This file can go, it's a holdover from an earlier rev of the startup app. Should honestly get deleted from the modules repo too.

fsinapi (approved):
Nice! Just had comments on stuff you can delete 🗑️

PR #558, opened 2023-01-27
Updates for DVT head and pipette behavior

sfoster1:
This is a branch that contains a couple PRs updating behavior for DVT. It should not be merged through the UI, but rebase-merged when we're ready.


sfoster1:
Superceded by #571, though the branch and tag will remain.

PR #559, opened 2023-01-27
refactor(head): set dvt gearing

sfoster1:
On DVT, there are only three gears, and the net reduction ratio is now 3.2 instead of 4.

Closes RET-1315


fsinapi (approved):
Math checks out to me 🔢

ahiuchingau (approved):
⚙️

PR #560, opened 2023-01-27
chore(build): fix small type in pipettes-384 entry

vegano1:


scripts/subsystem_versions.py
```diff
@@ -15,7 +15,7 @@
     "pipettes-single",
     "pipettes-multi",
     "pipettes-96",
-    "pippetes-384",
+    "pipetes-384",
```
fsinapi, line 18:
```suggestion
    "pipettes-384",
```

sfoster1 (approved):
lgtm

scripts/subsystem_versions.py
vegano1: smh

PR #561, opened 2023-01-30
feature(rear-panel):  bring up the usb interface of the rear panel mcu

ryanthecoder:
Compile in the HAL USB driver into the rear panel board
Assign it the VID 0x04D8 and PID 0xEF01 (appropriate confluence page updated)
template out the message handling by implementing an echo back on the USB interface


ryanthecoder:
needs to be tested still

include/rear-panel/core/tasks/host_comms_task.hpp
```diff
@@ -0,0 +1,134 @@
...
+#include "common/core/message_queue.hpp"
+#include "rear-panel/core/double_buffer.hpp"
+#include "rear-panel/core/messages.hpp"
+
+namespace host_comms_task {
+
+using TaskMessage =
+    std::variant<std::monostate, messages::IncomingMessageFromHost>;
+
+template <class ResposneQueue>
```
sfoster1, line 16:
```suggestion
template <class ResponseQueue>
```

sfoster1: and so on

include/rear-panel/core/tasks/host_comms_task.hpp
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    ResposneQueue &resp_queue;
+};
+
+/**
+ * The task type.
+ */
+template <template <class> class QueueImpl>
+requires MessageQueue<QueueImpl<TaskMessage>, TaskMessage>
```
sfoster1, line 79: we might need to rework this into the ot3-firmware style with a message handler idk

rear-panel/firmware/host_comms_task/freertos_comms_task.cpp
```diff
@@ -0,0 +1,178 @@
...
+    using namespace host_comms_control_task;
+    ssize_t remaining_buffer_count =
+        (_local_task.rx_buf.committed()->data()
+         // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+         + _local_task.rx_buf.committed()->size()) -
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic,cppcoreguidelines-pro-type-reinterpret-cast)
+        reinterpret_cast<char *>(Buf + *Len);
+
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+    if ((std::find_if(Buf, Buf + *Len,
```
sfoster1, line 153: idk if we want to do it in this PR (probably not) but we'll need to replace this with the binary messaging handling stuff instead of text mode

sfoster1 (commented):
couple little immediate things

include/rear-panel/core/messages.hpp
```diff
@@ -0,0 +1,31 @@
...
+// messages and include them from another file
+enum class MessageType : uint16_t {
+    ECHO = 0x00,
+};
+
+template <MessageType MType>
+struct BinaryFormatMessage {
+    static const auto message_type = MType;
+    const uint16_t length;
+    const char* data;
```
sfoster1, line 22: is the idea that this data will physically remain in the double buffer until it's done processing, which maybe is just until the message is received? if we want to copy it out of the double buffer immediately we'll need this to be a `std::array<char>` of appropriate max size.

sfoster1: Also if these are binary messages it's a bad idea IMO to put them in a native stringlike type because you reasonably read a native stringlike type as zero-terminated which these definitely won't be. at the very least imo this should be `uint8_t*`.

sfoster1 (commented):
This is looking good, I'd like some further thought on having the message structs capture a pointer to the double buffer but I understand if the answer is "it's an early implementation it'll change"

ryanthecoder:
tested by running the following script over ssh:
import serial
ser = serial.Serial('/dev/ttyACM0',115200)
len = ser.write(b"hello there")
data = ser.read(size=len)
print(data)
ser.close()

sfoster1 (approved):
Looks good to me once tested!

PR #562, opened 2023-01-30
refactor(pipettes): disable the pressure sensor interrupt for the 96 channel

Laura-Danielle:
## Overview

We will not have a fix for the noise on the data ready line of the 96 channel until DVT. To make testing easier, we should just disable this in main until then.


sfoster1 (approved):
:cry:

PR #564, opened 2023-02-02
refactor(all): modify motor hardware to be templated rather than use virtual functions

Laura-Danielle:
## Overview

Closes RET-1313 and RET-1316.

## TODO

- Still need to fully fix the simulators

## Test plan

- [ ]  Test that booting the system doesn't cause the 96 channel to drop anymore

- [ ] Run a full protocol with all of the boards flashed to this branch


include/motor-control/core/tasks/motor_hardware_task.hpp
```diff
@@ -85,7 +83,7 @@ class PipetteMotorHardwareTask {
     }
 
   private:
-    motor_hardware::PipetteStepperMotorHardwareIface* hardware;
+    PipetteStepperHardware* hardware;
```
sfoster1, line 86: looking at this file again, if you made all the hardwares required to have a `periodic_tasks()` method that did basically `read_estop_in()`, `read_limit_switch()`, `read_sync_in()`, and in the case of pipette `read_tip_sense()` you could get rid of the pipette motor task

motor-control/firmware/stepper_motor/motor_hardware.cpp
```diff
@@ -4,64 +4,28 @@
...
-void MotorHardware::read_sync_in() { sync = (gpio::is_set(pins.sync_in)); }
-
-void MotorHardware::set_LED(bool status) {
-    if (status) {
-        gpio::set(pins.led);
-    } else {
-        gpio::reset(pins.led);
-    }
+template <>
+void MotorHardware<HardwareConfig>::deactivate_motor() {
```
sfoster1, line 15: I see what you're going for here, but we really shouldn't have explicit specializations for application-dependent types in library code like this - and having these specializations and _only_ these specializations live in a .cpp file instead of the header makes it harder to read. Given that `MotorHardware` is kind of trivial, i think it's probably better to just have a `HeadMotorHardware` that lives in `include/head/firmware` that does this and can explicitly handle the brake pins, and get these definitions unspecialized and in the header file.

pipettes/firmware_l5/main.cpp
```diff
@@ -69,6 +69,21 @@ class EEPromHardwareIface : public eeprom::hardware_iface::EEPromHardwareIface {
         }
     }
 };
+
+auto convert_to_motor_hardware(
+    pipette_motor_hardware::HardwareConfig motor_config)
```
sfoster1, line 74: this strikes me as slightly worrying but I'm not sure why - why does this need to exist? can `convert_to_motor_hardware` be templated or something?

sfoster1 (commented):
Some initial comments

pipettes/firmware_l5/main.cpp
Laura-Danielle: It shouldn't exist. There was a typo on my part awhile back where the hardware configs were accidentally using the pipette hardware configs.

The way that we're doing tip sensing is going to change so the need for a pipette hardware class seems moot to me anyways.

I didn't want to address that in this PR since it's already a large-ish refactor. I'll make a ticket to track the work.

motor-control/firmware/stepper_motor/motor_hardware.cpp
Laura-Danielle: okayyyy 😢

include/motor-control/core/tasks/motor_hardware_task.hpp
Laura-Danielle: Same comment as below, but:

The way that we're doing tip sensing is going to change so the need for a pipette hardware class seems moot to me anyways.

I didn't want to address that in this PR since it's already a large-ish refactor. I'll make a ticket to track the work.

include/motor-control/core/stepper_motor/motion_controller.hpp
```diff
@@ -20,14 +20,18 @@ using namespace motor_hardware;
...
+template <lms::MotorMechanicalConfig MEConfig,
+          StepperMotorHardwareIface StepperHardware>
+// requires
+// StepperMotorHardwareIface<MotorHardwareType<HardwareConfigurationsType>>
 class MotionController {
   public:
     using GenericQueue = freertos_message_queue::FreeRTOSMessageQueue<Move>;
     using UpdatePositionQueue = freertos_message_queue::FreeRTOSMessageQueue<
         can::messages::UpdateMotorPositionEstimationRequest>;
+    // using StepperHardware = MotorHardwareType<HardwareConfigurationsType>;
```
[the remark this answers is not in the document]
Laura-Danielle: yes

Laura-Danielle:
Closing in favor of simply creating an extended class for the head board 😿

PR #567, opened 2023-02-07
refactor: add revision to device info

sfoster1:
The revision of the PCBA that is currently running the firmware is part
of the firmware image. This seems circular, but it is established by the
firmware image that is programmed at the factory when the PCBA is
integrated into its subassembly - at that time, we'll program using an
image that has a bootloader with the appropriate revision compiled in.
From there, we'll only update micros with application firmware for the
revision that matches the bootloader. That means we don't need to store
the firmware revision in the eeprom, and compiling it in is enough.

The revision needs to be communicated from a microcontroller to the host
for this to work. That happens by adding it to the device info
message (with tests) for all the bootloaders and all the application
firmwares (all the ones that currently handle device info, anyway - it's
not implemented yet in rear-panel).

There will still be some bootloaders in the wild that don't and will
never have this change because it's impractical to reflash the boards,
but they are all DVT boards of various varieties and can be defaulted as
such in the python side of the update process.

This is a PR that touched a lot of files because of further CMake concerns; I've 
split it into commits by what the point of the changes were, so that's a good
way to review it.

This PR can be merged in any order relative to https://github.com/Opentrons/opentrons/pull/12111 
because the python side will ignore extra data in CAN messages, but that PR will 
be necessary to test that this PR sends the right data.

## Test plan
- [ ] Run on some boards and check that the revision is sent appropriately

Closes RET-1321

### DIFF
--- a/include/can/core/messages.hpp
+++ b/include/can/core/messages.hpp
@@ -113,25 +113,20 @@ using DeviceInfoRequest = Empty<MessageId::device_info_request>;
 
 struct DeviceInfoResponse : BaseMessage<MessageId::device_info_response> {
     /**
-     *   TODO (al, 2021-09-13)
-     *   Seth's thoughts on future of payload
-     *   IMO we should set this up for a couple more things than just version,
-     * and then can care about version in general a little less. I don't think
-     * it's necessarily critical to get right the first time, but we could do
-     * something like:
-     *   - two bits of build type (dev, testing, release)
-     *   - a byte or two of like message schema version
-     *   - a four byte incrementing version seems fine i suppose
-     *   - a byte or two for well-known-id system identification through some
-     * centrally defined enum
-     *   - at some point we'll want a serial number probably
-     *   - a hardware revision unless we want to fold that into the
-     * well-known-id
+     * The device info tells us some information that's useful for all nodes on
+     * the canbus - e.g., nothing specific to pipettes or grippers, but what's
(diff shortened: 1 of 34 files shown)


Laura-Danielle (approved):
lgtm

include/can/core/messages.hpp
```diff
@@ -113,25 +113,20 @@ using DeviceInfoRequest = Empty<MessageId::device_info_request>;
...
+     * a CI build or a dev build; and the hardware revision, which is necessary
+     * to specify what firmware this device should be updated with.
      */
     uint32_t message_index;
     uint32_t version;
     uint32_t flags;
     std::array<char, VERSION_SHORTSHA_SIZE> shortsha;
+    char primary_revision;
+    char secondary_revision;
+    std::array<char, 2> tertiary_revision;
```
Laura-Danielle, line 129: Stray comment -- I thought we weren't tracking tertiary revisions in SW

vegano1 (approved):

include/can/core/messages.hpp
vegano1: I believe thats there as a placeholder just in case.

sfoster1: yeah. if we ever do, we'll do it in the eeprom and report it here, but I wanted to carve out the space. for now these will always be 0.

sfoster1:
Tested that both applications and bootloaders now send the correct revision.

PR #568, opened 2023-02-08
refactor: deliver all firmware revisions to OE

sfoster1:
Now that we have the capability to build all the electrical revisions, we need to deliver them to the built system images. This PR does that and a couple other little refactors:

- Alter `subsystem_versions.py` to look for hex files in the `dist/applications` directory that `cmake --install` will populate, which means we don't have to worry about image.hex files getting in there; also dynamically pick up subsystems
- Make some changes to make OE's job a little easier, such as: bump the minimum required version, to bump the preset version, to allow configure presets without binary dirs, to let OE specify a build directory in _OE's_ build directory, and do some other nice little changes

## Testing
- [ ] Look at the outcome of https://github.com/Opentrons/oe-core/pull/53 and make sure that hex files are properly included

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,8 +1,8 @@
 {
-  "version": 2,
+  "version": 3,
   "cmakeMinimumRequired": {
     "major": 3,
-    "minor": 20,
+    "minor": 21,
     "patch": 0
   },
   "configurePresets": [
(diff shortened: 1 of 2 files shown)


.github/workflows/pipettes.yaml
```diff
@@ -128,10 +128,6 @@ jobs:
         run: cmake --build ./build-host --target pipettes-96-simulator
         working-directory: ot3-firmware
 
-      - name: 'Build 384 simulator'
```
Laura-Danielle, line 131: A note for both PRs,

I believe some work that @vegano1 was doing included compiling/looking for the 384 firmware. I would just triple check on everything to do w firmware updates and make sure removing the presets etc doesn't break anything.

CMakePresets.json
```diff
@@ -13,35 +13,40 @@
...
     },
-    {
-      "name": "cross",
-      "displayName": "STM32 G4 OT-3 cross-compilation",
-      "description": "Build application firmware for OT-3 systems that use STM32, for flashing onto boards",
+    {"name": "cross-no-directory-reqs",
+      "displayName": "STM32 G4 OT-3 cross-compilation with no directory specifications",
+      "description": "Build application firmware for OT-3 systems that use STM32",
       "generator": "Unix Makefiles",
-      "binaryDir": "${sourceDir}/build-cross",
```
Laura-Danielle, line 21: Why don't we need the binary directory anymore to build hex files?

Laura-Danielle (approved):
Two comments, but looks good to me.

.github/workflows/pipettes.yaml
vegano1: oe-core looks for pioettes-384 hex file to install, but that's not a breaking change. But that's most likely getting changed as part of the oe-core portion of the changes Seth is working on.

vegano1:
Very nice!
This looks good to me

vegano1 (approved):

.github/workflows/pipettes.yaml
sfoster1: The nice thing about that is that it's actually all controlled in here by the `subsystem_versions.json` - so it all works still

CMakePresets.json
sfoster1: We do, but we want a way to build the firmware where we can specify the binary directory.
The `cross` preset is behaviorally unchanged - you still don't have to specify anything. What this does is add a new preset where you have to specify the build dir with `-B`. That's really specifically and only (but very) useful in openembedded, where by convention you build stuff somewhere completely unrelated from the source directory. This lets OE specify an arbitrary directory to build in while keeping the dev experience unchanged.

PR #569, opened 2023-02-09
fix(motor_control, pipettes): Only change the gpio state if it is stable for two ticks in a row

ryanthecoder:
Due to a HW problem in pipettes our old denouncing code wasn't working well enough.

The new code makes sure that the state of the relevant pins is only saved if it is stable across two ticks

### DIFF
--- a/motor-control/firmware/brushed_motor/brushed_motor_hardware.cpp
+++ b/motor-control/firmware/brushed_motor/brushed_motor_hardware.cpp
@@ -33,21 +33,11 @@ void BrushedMotorHardware::activate_motor() { gpio::set(pins.enable); }
 void BrushedMotorHardware::deactivate_motor() { gpio::reset(pins.enable); }
 
 void BrushedMotorHardware::read_limit_switch() {
-    // only set the state if the bounce matches the current gpio_is_set
-    // on the first state change it won't match but on the second tick it will
-    // and we can set it to the new state.
-    std::atomic_bool new_state = gpio::is_set(pins.estop_in);
-    limit.store(new_state == limit_bounce ? new_state : limit);
-    limit_bounce.store(new_state);
+    gpio::debouce_update(gpio::is_set(pins.limit_switch), limit, limit_bounce);
 }
 
 void BrushedMotorHardware::read_estop_in() {
-    // only set the state if the bounce matches the current gpio_is_set
-    // on the first state change it won't match but on the second tick it will
-    // and we can set it to the new state.
-    std::atomic_bool new_state = gpio::is_set(pins.estop_in);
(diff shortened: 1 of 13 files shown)


motor-control/firmware/brushed_motor/brushed_motor_hardware.cpp
```diff
@@ -33,11 +33,21 @@ void BrushedMotorHardware::activate_motor() { gpio::set(pins.enable); }
 void BrushedMotorHardware::deactivate_motor() { gpio::reset(pins.enable); }
 
 void BrushedMotorHardware::read_limit_switch() {
-    limit = gpio::is_set(pins.limit_switch);
+    // only set the state if the bounce matches the current gpio_is_set
+    // on the first state change it won't match but on the second tick it will
+    // and we can set it to the new state.
+    std::atomic_bool new_state = gpio::is_set(pins.estop_in);
```
sfoster1, line 39: this local doesn't have to be atomic

motor-control/firmware/brushed_motor/brushed_motor_hardware.cpp
```diff
@@ -33,11 +33,21 @@ void BrushedMotorHardware::activate_motor() { gpio::set(pins.enable); }
...
+    limit.store(new_state == limit_bounce ? new_state : limit);
+    limit_bounce.store(new_state);
 }
 
 void BrushedMotorHardware::read_estop_in() {
-    estop = gpio::is_set(pins.estop_in);
+    // only set the state if the bounce matches the current gpio_is_set
+    // on the first state change it won't match but on the second tick it will
+    // and we can set it to the new state.
+    std::atomic_bool new_state = gpio::is_set(pins.estop_in);
```
sfoster1, line 48: ditto

sfoster1 (changes requested):
The temporary variables in the functions don't have to be atomic.

I also think this is the same code used 6 different times - maybe make a really simple struct that owns two atomic_bools and has an `update(bool)` method that has the logic? It can even be in core and be tested.

sfoster1 (approved):
This looks good and we can merge it once we fix the typo, but I'd really prefer if
- This was in `core/` rather than `firmware/` and thus was testable, and had tests
- This was gathered into a separate struct that just debounces bools and nothing else, so the hardware handler can just have a bunch of instances of `Debouncer sync; Debouncer estop; Debouncer limit;` and then update them with `sync.update_debounce(gpi::is_set(pins_sync.in));` and query them with `sync.debounced_value()` or whatever

However, also down to make those in a follow-up.

Also also we should test this on hardware before merging.

sfoster1 (approved):
Excellent, thank you!

PR #570, opened 2023-02-09
chore: Add --no-ansi flag when running poetry

DerekMaggio:
Poetry is dying in CI because of [this issue](https://github.com/python-poetry/poetry/issues/7184). Workaround is to add `--no-ansi` flag. 

Dumb. But it works.


fsinapi (approved):
Nice fix. Gotta love the Wall Of Mentions on that issue.

PR #571, opened 2023-02-10
apply dvt fixes to main

sfoster1:
Now that we can separate out firmware by the revision of the PCBA that it applies to, we can merge in the changes from #558 and apply them only on the appropriate revision.

Note - the tool ID changes here are retroactive. That's because it's fundamentally a really bad idea to ever mix the EVT and DVT behavior, and pipettes are a user-swappable part that are not mechanically distinct between revs. Instead, we'll rework all EVT pipettes and head boards to have the new tool ID hardware.

## Test Plan
- [x] Make sure that DVT heads get the new gear behavior
- [x] Make sure that the EVT heads don't
- [x] Make sure that everything gets the new tool id behavior

## BIG MERGE NOTE
Because the new behavior was individually ticketed and individually committed, I plan to massage this PR's history and then _rebase_ merge it instead of squashing.


fsinapi (approved):

Laura-Danielle (approved):
Good for me to merge in 👍

PR #572, opened 2023-02-13
feature(rear-panel):  add abinary messaging protocol to the rear panel

ryanthecoder:
Create a system for defining messages then lets the protocol automatically parse and create the message type
using the incoming binary message from the host. 

Also create the serialize to take messages and convert them to a binary blob to send to the host.

add tests to verify the serializing and parsing

### DIFF
--- a/include/rear-panel/core/binary_parse.hpp
+++ b/include/rear-panel/core/binary_parse.hpp
@@ -22,7 +22,7 @@ concept Parsable = requires(std::array<uint8_t, 0> buffer) {
 };
 
 /**
- * Parser of can message bodies.
+ * Parser of usb message bodies.
  *
  * @tparam T The types of Parsables (messages) this parser will support.
  */
--- a/include/rear-panel/core/binary_parse.hpp
+++ b/include/rear-panel/core/binary_parse.hpp
@@ -43,9 +43,9 @@ class Parser {
                const Iterator& limit) -> Result {
         auto result = Result{std::monostate{}};
         // Fold expression over Parsable template type.
-        // Create a lambda that accepts the id and body.
+        // Create a lambda that accepts the type and body.
         ((([&result, message_type, &payload, &limit]() -> void {
-             // If message id matches Parsable's id parse will build the result.
+             // If message type matches Parsable's type parse will build the result.
              if (message_type == T::message_type) {
                  result = T::parse(payload, limit);
              }
--- a/include/rear-panel/core/binary_parse.hpp
+++ b/include/rear-panel/core/binary_parse.hpp
@@ -1,0 +1,71 @@
+#pragma once
+
+#include <variant>
+
+#include "common/core/bit_utils.hpp"
+#include "rear-panel/core/bin_msg_ids.hpp"
+#include "rear-panel/core/messages.hpp"
+
+namespace binary_parse {
+
+/**
+ * The concept describing the interface of a parsable can message
+ * @tparam T
+ */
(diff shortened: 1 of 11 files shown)


include/rear-panel/core/messages.hpp
```diff
@@ -1,31 +1,132 @@
...
+#include "rear-panel/core/bin_msg_ids.hpp"
+#include "rear-panel/core/binary_parse.hpp"
 
-// TODO(ryan): as part of RET-1307 create an auto gen of the ids like the can
-// messages and include them from another file
-enum class MessageType : uint16_t {
-    ECHO = 0x00,
-};
+// this is the max size of the double_buffer used in freertos_message_queue
+constexpr size_t MAX_MESSAGE_SIZE = 512U * 4;
```
sfoster1, line 12: this is... pretty big. I think 1024 might be better since I think that's the max payload size for highspeed data payloads per https://www.beyondlogic.org/usbnutshell/usb3.shtml

include/rear-panel/core/messages.hpp
```diff
@@ -1,31 +1,132 @@
...
+            bit_utils::int_to_bytes(uint16_t(message_type), body, limit);
+        iter = bit_utils::int_to_bytes(length, iter, limit);
+        iter = bit_utils::int_to_bytes(version, iter, limit);
+        iter = bit_utils::int_to_bytes(flags, iter, limit);
+        iter =
+            std::copy_n(&shortsha[0],
+                        std::min(limit - iter,
+                                 static_cast<ptrdiff_t>(VERSION_SHORTSHA_SIZE)),
+                        iter);
+
```
sfoster1, line 115: we should add the revision in here as in the other firmwares

include/rear-panel/core/tasks/host_comms_task.hpp
```diff
@@ -63,46 +62,59 @@ class HostCommMessageHandler {
...
+    requires std::forward_iterator<InputIt> &&
+        std::sized_sentinel_for<InputLimit, InputIt>
+    auto visit_message(rearpanel::messages::DeviceInfoRequest &msg,
+                       InputIt tx_into, InputLimit tx_limit) -> InputIt {
+        std::ignore = msg;
+        const auto *ver_info = version_get();
+        auto response = rearpanel::messages::DeviceInfoResponse{
+            .length = rearpanel::messages::DeviceInfoResponse::get_length(),
+            .version = ver_info->version,
+            .flags = ver_info->flags};
```
sfoster1, line 82: add revision here also

sfoster1 (changes requested):
Looks good overall, but let's add the revision to the device info

include/rear-panel/core/binary_parse.hpp
```diff
@@ -0,0 +1,72 @@
...
+concept Parsable = requires(std::array<uint8_t, 0> buffer) {
+    /**
+     * It has a static parse factory method.
+     */
+    { T::parse(buffer.begin(), buffer.end()) } -> std::same_as<T>;
+    { T::parse(buffer.cbegin(), buffer.cend()) } -> std::same_as<T>;
+};
+
+/**
+ * Parser of can message bodies.
```
fsinapi, line 25:
```suggestion
 * Parser of usb message bodies.
```

fsinapi (approved):
Nice, LGTM

sfoster1 (approved):
Nice, looks good!

PR #574, opened 2023-02-17
fix(pipettes): we did not update the mount detect GPIO for the 96 channel for DVT changes

Laura-Danielle:
## Overview

Correct the mount detect GPIO pins.

### DIFF
--- a/include/bootloader/core/pipette_type.h
+++ b/include/bootloader/core/pipette_type.h
@@ -7,4 +7,14 @@ typedef enum Type {
     THREE_EIGHTY_FOUR_CHANNEL
 } PipetteType;
 
+
+#ifdef PIPETTE_TYPE_DEFINE
+
 static inline PipetteType get_pipette_type() { return PIPETTE_TYPE_DEFINE; }
+
+#else
+
+static inline PipetteType get_pipette_type() { return SINGLE_CHANNEL; }
+#endif  // pipette type define
+
+
--- a/include/bootloader/core/pipette_type.h
+++ b/include/bootloader/core/pipette_type.h
@@ -1,0 +1,10 @@
+#pragma once
+
+typedef enum Type {
+    SINGLE_CHANNEL,
+    EIGHT_CHANNEL,
+    NINETY_SIX_CHANNEL,
+    THREE_EIGHTY_FOUR_CHANNEL
+} PipetteType;
+
+static inline PipetteType get_pipette_type() { return PIPETTE_TYPE_DEFINE; }
--- a/pipettes/firmware/utility_gpio.c
+++ b/pipettes/firmware/utility_gpio.c
@@ -208,13 +208,24 @@ static void estop_input_gpio_init() {
 }
 
 static void mount_id_init() {
-    // B0: mount id
-    __HAL_RCC_GPIOB_CLK_ENABLE();
-    GPIO_InitTypeDef GPIO_InitStruct = {0};
-    GPIO_InitStruct.Pin = GPIO_PIN_0;
-    GPIO_InitStruct.Mode = GPIO_MODE_INPUT;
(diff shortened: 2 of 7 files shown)


pipettes/firmware/utility_gpio.c
```diff
@@ -208,13 +208,24 @@ static void estop_input_gpio_init() {
...
-    __HAL_RCC_GPIOB_CLK_ENABLE();
-    GPIO_InitTypeDef GPIO_InitStruct = {0};
-    GPIO_InitStruct.Pin = GPIO_PIN_0;
-    GPIO_InitStruct.Mode = GPIO_MODE_INPUT;
-    GPIO_InitStruct.Pull = GPIO_PULLDOWN;
-    HAL_GPIO_Init(GPIOB, &GPIO_InitStruct);
+    PipetteType pipette_type = get_pipette_type();
+    if (pipette_type == NINETY_SIX_CHANNEL) {
+        // C3: mount id
+        __HAL_RCC_GPIOC_CLK_ENABLE();
```
ryanthecoder, line 214:
```suggestion
        __HAL_RCC_GPIOB_CLK_ENABLE();
```

ryanthecoder (changes requested):
looks good just use the right clock since the pin is on GPIOB

pipettes/firmware/utility_gpio.c
```diff
@@ -208,13 +208,24 @@ static void estop_input_gpio_init() {
...
+        GPIO_InitTypeDef GPIO_InitStruct = {0};
+        GPIO_InitStruct.Pin = GPIO_PIN_3;
+        GPIO_InitStruct.Mode = GPIO_MODE_INPUT;
+        GPIO_InitStruct.Pull = GPIO_PULLDOWN;
+        HAL_GPIO_Init(GPIOB, &GPIO_InitStruct);
```
fsinapi, line 219:
```suggestion
        HAL_GPIO_Init(GPIOC, &GPIO_InitStruct);
```

fsinapi (changes requested):
Referencing the Altium, looks like C3 is correct. Looks like the code is initializing B3.

caila-marashaj (approved):
lgtm!

Laura-Danielle:
> Referencing the Altium, looks like C3 is correct. Looks like the code is initializing B3.

Yeah it should be C3/port C, bouncing around to too many things.

fsinapi (approved):
nice

vegano1 (approved):

sfoster1 (approved):
noice

include/bootloader/core/pipette_type.h
```diff
@@ -0,0 +1,20 @@
...
+} PipetteType;
+
+
+#ifdef PIPETTE_TYPE_DEFINE
+
+static inline PipetteType get_pipette_type() { return PIPETTE_TYPE_DEFINE; }
+
+#else
+
+static inline PipetteType get_pipette_type() { return SINGLE_CHANNEL; }
```
fsinapi, line 17: [nit] Because `PIPETTE_TYPE_DEFINE` exists for every target in cmake, I think a `#error` here makes more sense than having a default value - the only case where we should hit this `#else` is if something is wrong with the build config.

fsinapi (approved):
Changes LGTM just had one nit

include/bootloader/core/pipette_type.h
Laura-Danielle: `node_id_stm32g4xx.c` which uses `pipette_type.h` is compiled by all the bootloader types whereas only the pipette bootloaders actually specify a `PIPETTE_TYPE_DEFINE`. If we threw an error, all the other bootloaders wouldn't be able to compile successfully. It's kinda gross, but I didn't want to further split out files for this.

PR #575, opened 2023-02-17
feat(rear-panel):  add an enter bootloader message to the rear panel mcu

ryanthecoder:
create a freertos task to handle system tasks and add a c file that disables peripherals and jumps to the bootloader

`
>>> import subprocess
>>> import time
>>> try:
...     ser = serial.Serial("/dev/ttyACM0",115200)
...     ser.write(bytes.fromhex("00050000"))
...     data = ser.read(size=5)
...     print(data)
...     ser.close() 
... except Exception as e:
...     print(e)
... 
4
[58096.797118] usb 1-1.1: USB disconnect, device number 22
read failed: device reports readiness to read but returned no data (device disconnected or multiple access on port?)
>>> time.sleep(0.5)
[58097.047217] usb 1-1.1: new full-speed USB device number 23 using ci_hdrc
>>> 
>>> output = subprocess.run(["lsusb"], capture_output=True)
>>> success =(output.stdout.find(b"DFU")) > 0 
>>> print("Device in dfu mode {success}".format(success=success))
Device in dfu mode True
>>> 
`

### DIFF
--- a/include/rear-panel/core/messages.hpp
+++ b/include/rear-panel/core/messages.hpp
@@ -24,16 +24,15 @@ struct Echo : BinaryFormatMessage<MessageType::ECHO> {
     std::array<uint8_t, MAX_MESSAGE_SIZE> data;
 
     template <bit_utils::ByteIterator Input, typename Limit>
-    static auto parse(Input body, Limit limit) -> Echo {
+    static auto parse(Input body, Limit limit)
+        -> std::variant<std::monostate, Echo> {
         uint16_t type = 0;
         uint16_t len = 0;
         std::array<uint8_t, MAX_MESSAGE_SIZE> data{};
         body = bit_utils::bytes_to_int(body, limit, type);
         body = bit_utils::bytes_to_int(body, limit, len);
         if (body + len > limit) {
-            // we got something wrong from this figure out what we want to
-            // do in this situation
-            len = uint16_t(limit - body);
+            return std::monostate{};
         }
         len = std::min(static_cast<size_t>(len), data.size());
         std::copy_n(body, len, data.begin());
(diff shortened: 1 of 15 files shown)


include/rear-panel/core/messages.hpp
```diff
@@ -135,10 +135,51 @@ struct DeviceInfoResponse : BinaryFormatMessage<MessageType::DEVICE_INFO_RESP> {
...
+    uint16_t length;
+
+    template <bit_utils::ByteIterator Input, typename Limit>
+    static auto parse(Input body, Limit limit) -> EnterBootlader {
+        uint16_t type = 0;
+        uint16_t len = 0;
+        body = bit_utils::bytes_to_int(body, limit, type);
+        body = bit_utils::bytes_to_int(body, limit, len);
+        if (len > 0) {
+            // we got something wrong from this figure out what we want to
```
fsinapi, line 148: This actually brings up the point that `binary_parse` doesn't have a way to fail parsing AFAIK. Doesn't have to be in this PR, but I think it'd make sense for the `Parsable` concept to dictate that `parse` returns an _optional_ value so there's a way to propagate a parsing error.

fsinapi (commented):
Looks good, we should probably add tests for at least the basic task message passing though.

include/rear-panel/core/messages.hpp
ryanthecoder: upgraded the parsing so that if a body of a message is invalid it will return monostate, and the serial handling task will automatically send out a AckFailed if a message on the serial bus is not parsed into a valid message

fsinapi (approved):
groovy

PR #576, opened 2023-02-21
fix(motor-control): clear move group after stall

fsinapi:
During stall repeatability testing, we noticed that motor axes would sometimes continue to drive into the endstop after stalling. The culprit appears to be that, when the Stop message was removed in #542, there was no replacement for clearing out the interrupt's queue.

This PR adds functionality to the interrupt handler to clear out its queue when a stall occurs. We've previously had trouble using the actual `clear` function for FreeRTOS queues in interrupts, so instead we set a flag to "drain" the queue (one message per interrupt) until it's empty.

Tested by:
* Move an axis very close to its endstop
* Command the axis to move through the endstop with the `_check_stalls` flag enabled. Make the distance at least twice that of the distance to the endstop, to ensure it will collide during the acceleration phase.
* Send an "update position" request, and then move the axis back away from the endstop to make sure the interrupt is behaving normally

Tested on head to ensure the queue-draining functionality doesn't cause timing issues, and tested on the Gantry-X with a slow motion video to verify that the axis doesn't overdrive into the spring-loaded endstop.

### DIFF
--- a/include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
+++ b/include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
@@ -136,11 +136,6 @@ class MotorInterruptHandler {
          * This function is called from a timer interrupt. See
          * `motor_hardware.cpp`.
          */
-        if (clear_queue_until_empty && has_move_messages()) {
-            if (!pop_and_discard_move()) {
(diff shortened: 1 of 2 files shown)


ahiuchingau (approved):
Sweet!

sfoster1 (approved):
Nice

PR #577, opened 2023-02-22
refactor(gripper): hardware configurations

ahiuchingau:
We had too many c files for gripper motor configuration, which made managing new revisions really difficult. '

So instead of having multiple headers files, let's just keep all of these defs in one master `motor_hardware.h`.

We will have separate source files for each axis: `motor_hardware_g.c` and `motor_hardware_z.c` for axis-specific configurations; as for functions that are shared, such as HAL timer callbacks, we will put those in `motor_hardware_shared.c`.


gripper/firmware/motor_hardware_g.c
```diff
@@ -0,0 +1,377 @@
...
+    }
+}
+
+/**
+ * @brief DAC1 Initialization Function
+ * @param None
+ * @retval None
+ */
+static void DAC1_Init(void) {
+    hdac1.Instance = DAC1;
```
fsinapi, line 330: I know this isn't actually new code, but while you're in these files - we should really be explicitly setting the `State`variable in these HAL . The bootloader and main application use the same RAM and anything that isn't explicitly initialized may get changed by the bootloader - [this was an issue with the TC2](https://github.com/Opentrons/opentrons-modules/pull/400).

gripper/firmware/motor_hardware_z.c
```diff
@@ -0,0 +1,71 @@
+#include "motor_hardware.h"
+
+TIM_HandleTypeDef htim7;
+
+static void TIM7_Base_Init(void) {
+    TIM_MasterConfigTypeDef sMasterConfig = {0};
+
+    htim7.Instance = TIM7;
```
fsinapi, line 8: Same comment about the `State` variable

gripper/firmware/motor_hardware_g.c
```diff
@@ -0,0 +1,377 @@
...
+TIM_HandleTypeDef htim1;
+TIM_HandleTypeDef htim2;
+TIM_HandleTypeDef htim3;
+TIM_HandleTypeDef htim4;
+
+TIM_OC_InitTypeDef htim1_sConfigOC = {0};
+TIM_OC_InitTypeDef htim3_sConfigOC = {0};
+
+
+uint32_t round_closest(uint32_t dividend, uint32_t divisor) {
```
fsinapi, line 17: This and `calc_prescaler` either belong in `motor_hardware_shared.c` or they can be static functions.

fsinapi (commented):
Had a few comments. I really like this approach over the separate `rev1`/`proto` source files, much cleaner!

gripper/firmware/utility_gpio.c
```diff
@@ -79,18 +67,97 @@ static void estop_input_gpio_init() {
...
+    // Enable
+    GPIO_InitStruct.Pin = GPIO_PIN_9;
+    GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
+    HAL_GPIO_Init(GPIOA,  // NOLINT(cppcoreguidelines-pro-type-cstyle-cast)
+                  &GPIO_InitStruct);
+}
+
 void utility_gpio_init(void) {
+    /* GPIO Ports Clock Enable */
+    __HAL_RCC_GPIOC_CLK_ENABLE();
```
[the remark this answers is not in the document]
ahiuchingau: Yeah that's true!

gripper/firmware/motor_hardware_z.c
```diff
@@ -0,0 +1,71 @@
...
+
+};
+
+/**
+ * @brief SPI MSP De-Initialization
+ * This function freeze the hardware resources used in this example
+ * @param hspi: SPI handle pointer
+ * @retval None
+ */
+void HAL_SPI_MspDeInit(SPI_HandleTypeDef* hspi) {
```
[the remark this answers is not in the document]
ahiuchingau: I moved the GPIO setup for SPI to the `utility_gpio.c` file

ryanthecoder (approved):

fsinapi (approved):
Nice, LGTM!

sfoster1 (approved):
Agree with @fsinapi but looks great other than that!

PR #578, opened 2023-02-23
feat(gripper): add support for DVT encoder

ahiuchingau:
This PR adds z-axis encoder support for DVT gripper (requires #577).


gripper/firmware/interfaces_z_motor.cpp
```diff
@@ -71,8 +71,13 @@ struct motion_controller::HardwareConfig motor_pins {
 /**
  * The motor hardware interface.
  */
+#if PCBA_PRIMARY_REVISION == 'c'
+static constexpr void* enc_handle = &htim7;
```
fsinapi, line 75: Should this be `htim8`?

fsinapi (commented):
In addition to switching the TIM handle, you're going to have to edit some initialization parameters in `gripper/firmware/interfaces_z_motor.cpp` (and the simulator equivalent) to add parameters for the encoder.

gripper/firmware/interfaces_z_motor.cpp
ahiuchingau: oh man yeah

gripper/firmware/utility_gpio.c
```diff
@@ -3,6 +3,18 @@
 #include "platform_specific_hal_conf.h"
 #include "stm32g4xx_hal_gpio.h"
 
+#if PCBA_PRIMARY_REVISION == 'c'
```
sfoster1, line 6: this feels like it should be a different PR since it's doing something pretty unrelated

sfoster1 (changes requested):
Couple inline things, but also IMO when we add an `#ifdef` for a new revision, we should always make it open-ended for the future. If we add a bunch of `#if PCBA_PRIMARY_REVISION == 'c'`, then whenever we add new revision `d` we have to go and edit those to make them `#if PCBA_PRIMARY_REVISION == 'c' || PCBA_PRIMARY_REVISION == 'd'`, and if we forget we won't get compiler errors or anything we'll just silently get the old behavior.

Instead, we should start out by making it `#if PCBA_PRIMARY_REVISION != 'b' && PCBA_PRIMARY_REVISION != 'a'` (or, well, at some point we should add a `REVISION_GREATER_EQUAL` macro) so that whenever we add revision `d` we'll automatically get the `c` behavior to build on.

gripper/firmware/interfaces_z_motor.cpp
```diff
@@ -116,16 +122,33 @@ static freertos_message_queue::FreeRTOSMessageQueue<
     can::messages::UpdateMotorPositionEstimationRequest>
     update_position_queue("Position Queue");
 
+#if PCBA_PRIMARY_REVISION == 'c'
```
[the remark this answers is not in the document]
ahiuchingau: This one can be collapsed to the block above, but the stallcheck stuff in line 138 will needs to be separate

gripper/firmware/utility_gpio.c
ahiuchingau: Oh I should've put it in the description; we had to switch these GPIOs out in order to add the new encoder. So I'm gonna try to keep these here, otherwise, the encoder wouldn't work properly.

gripper/firmware/interfaces_z_motor.cpp
```diff
@@ -116,16 +122,33 @@ static freertos_message_queue::FreeRTOSMessageQueue<
...
+static constexpr float encoder_pulses = 1024.0;
+#else
+static constexpr float encoder_pulses = 0.0;
```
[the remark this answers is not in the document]
ahiuchingau: Fixed by #583

PR #579, opened 2023-02-23
chore(scripts): add the firmware to the oe builds

ryanthecoder:
update the manifest version to 2
add a usb_subsystems category after subsystems
add rear panel to the usb_subsystems


sfoster1 (approved):
Looks good to me, properly formats the output.

PR #581, opened 2023-02-27
feat(head): utilize the ebrake when motors are disabled on the ZA

Laura-Danielle:
## Overview

Closes RET-1313. Now, whenever head motors are disabled, it will automatically re-enable the ebrake so that heavy instruments (such as a 96 channel pipette) do not fall. At some point in the future, hopefully we can refactor these hardware classes and make the inheritance a little cleaner. For now, I just chose to separate everything out which is very
similar to how `pipette_motor_hardware` works.


sfoster1 (approved):
Looks good to me once tests+lint pass

fsinapi:
If the only difference is that we  use the ebrake for the Head motor, have we considered adding the ebrake as just an optional configuration for the MotorHardware class? Maybe something along the lines of making the value in HardwareConfig be `std::optional<gpio::PinConfig> ebrake = std::nullopt;` and only setting it if it's defined - not the cleanest solution, but I worry about having three almost-identical copies of this class.

Laura-Danielle:
> 

Yeah I agree, which is why I had attempted to do that [big refactor ](https://github.com/Opentrons/ot3-firmware/pull/564) which didn't turn out the way I wanted it to.

The pipette motor hardware class will be going away as the implementation for tip sensing will be different.

I am happy to change ebrake to an optional param. @sfoster1 what do you think about that?

sfoster1:
Okay, lets do it!

sfoster1 (approved):
That looks really nice

fsinapi (approved):
🆒

PR #582, opened 2023-02-28
fix(head): flip encoder on C2

sfoster1:
Encoder goes the wrong direction now because the gears flipped and the encoder is on the motor axle.

## Testing
- [x] See if the encoder goes the right direction on C2
- [x] See if the encoder goes the right direction on B and previous

### DIFF
--- a/head/firmware/motor_hardware_common.c
+++ b/head/firmware/motor_hardware_common.c
@@ -232,20 +232,20 @@ void MX_GPIO_Init(void) {
 }
 
 #if PCBA_PRIMARY_REVISION == 'b' || PCBA_PRIMARY_REVISION == 'a'
-    static const int encoder_direction = TIM_ICPOLARITY_RISING;
+    static const int encoder_chan1_direction = TIM_ICPOLARITY_RISING;
 #else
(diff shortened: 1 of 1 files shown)


fsinapi (approved):
LGTM if it works

head/firmware/motor_hardware_common.c
```diff
@@ -231,14 +231,21 @@ void MX_GPIO_Init(void) {
     initialize_rev_specific_pins();
 }
 
+#if PCBA_PRIMARY_REVISION == 'b' || PCBA_PRIMARY_REVISION == 'a'
+    static const int encoder_direction = TIM_ICPOLARITY_RISING;
```
Laura-Danielle, line 235: The direction is flipped for both motors?

Laura-Danielle (approved):
had a q, but lgtm in general

ahiuchingau (approved):
🧭

head/firmware/motor_hardware_common.c
sfoster1: yeah, both motors need the change right?

sfoster1: since they both changed the number of gears

PR #584, opened 2023-02-28
refactor(pipettes): remove pipette hardware class

Laura-Danielle:
## Overview

The pipette hardware class was originally created to add the tip sense feature onto the pipette. This isn't really related to "motor hardware" at all and should thus be removed. Instead, we should add this functionality to the sensor hardware manager class.


sfoster1 (changes requested):
yessss

sfoster1 (approved):
nooooo

PR #585, opened 2023-02-28
fix(bootloader): update move stop condition enum name

ahiuchingau:
Changes are created by the header generator — seems like we forgot to edit the enum in bootloader after updating the `can/core/ids.hpp`.


fsinapi (approved):
(github doesn't have a sink emoji so just pretend) :sink:

sfoster1 (approved):

PR #586, opened 2023-02-28
fix(rear-panel): some fixes that were needed during update testing

ryanthecoder:
minor discrepancy on a struct size, and the rear panel application needs to be a .bin and not a .hex for dfu to work properly

### DIFF
--- a/include/rear-panel/core/messages.hpp
+++ b/include/rear-panel/core/messages.hpp
@@ -128,7 +128,7 @@ struct DeviceInfoResponse : BinaryFormatMessage<MessageType::DEVICE_INFO_RESP> {
 
     static auto get_length() -> uint16_t {
         //         version              flags          shortsha            revision
-        return sizeof(uint32_t) + sizeof(uint32_t) + VERSION_SHORTSHA_SIZE + 4;
+        return sizeof(uint32_t) + sizeof(uint32_t) + VERSION_SHORTSHA_SIZE + sizeof(revision);
     }
     auto operator==(const DeviceInfoResponse& other) const -> bool = default;
 };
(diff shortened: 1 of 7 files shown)


rear-panel/firmware/CMakeLists.txt
```diff
@@ -229,12 +229,22 @@ macro(rearpanel_loop)
...
       DEPENDS ${HEX_IMG_NAME})
   list(APPEND ${PROJECT_NAME}-IMAGES ${REVISION_TARGET}-image-hex)
-  list(APPEND ${PROJECT_NAME}-APPLICATIONS ${REVISION_TARGET}-hex)
+  list(APPEND ${PROJECT_NAME}-APPLICATIONS ${REVISION_TARGET}-bin)
   install_if_latest_revision(
     FILENAME ${HEX_IMG_FILENAME}
     COMPONENT robot-images
     THISREVISION ${REVISION}
     REVISIONS ${REVISIONS})
+  install_if_latest_revision(
```
sfoster1, line 238: we want to install _all_ revisions here - we only do `install_if_latest_revision` for `robot-images` because `robot-images` builds the zip we'll send to the factory. this can be 
```
install(
      FILES ${CMAKE_CURRENT_BINARY_DIR}/${REVISION_TARGET}.bin
      DESTINATION applications/
      COMPONENT Applications)
```

include/rear-panel/core/messages.hpp
```diff
@@ -127,8 +127,8 @@ struct DeviceInfoResponse : BinaryFormatMessage<MessageType::DEVICE_INFO_RESP> {
     }
 
     static auto get_length() -> uint16_t {
-        //         version              flags          shortsha
-        return sizeof(uint32_t) + sizeof(uint32_t) + VERSION_SHORTSHA_SIZE;
+        //         version              flags          shortsha            revision
+        return sizeof(uint32_t) + sizeof(uint32_t) + VERSION_SHORTSHA_SIZE + 4;
```
sfoster1, line 131: i think there's a `struct revision` in `version.h` you can get the sizeof

sfoster1 (approved):
Looks good to me!

PR #587, opened 2023-03-01
feat(gripper): make gripper jaw error tolerance adjustable

ahiuchingau:
**Changes:**
* New CAN message to change the gripper jaw error tolerance. 
* Added a new class `BrushedMotorErrorTolerance` to keep the tolerance values.
* The message is handled by the motion controller task which modifies the tolerance values in `BrushedMotorErrorTolerance`.


sfoster1 (approved):
Looks good, but let's maybe drop in a test where you change the value through a message or something and see it update.

PR #588, opened 2023-03-02
feat(pipettes): add tip sense check to the sensor hardware class

Laura-Danielle:
## Overview

Tip sensing falls more under the category of general sensor hardware. The pin configurations are also more conducive to having "multiples" of the tip sense. The next bit of work will involve setting up a task that can fire off tip presence data once a state change has been detected.

Some notes for next steps. For single/eight channel the ejector stage flag is active when in 'resting' position. For ninety-six the ejector stage flag is inactive when in 'resting' position.

The ninety six channel tip presence routine will be much more involved than the single/eight because it involves moving the pick up motors slightly 😞

### DIFF
--- a/include/sensors/firmware/sensor_hardware.hpp
+++ b/include/sensors/firmware/sensor_hardware.hpp
@@ -13,6 +13,12 @@ class SensorHardware : public SensorHardwareBase {
         : hardware(hardware) {}
     auto set_sync() -> void override { gpio::set(hardware.sync_out); }
     auto reset_sync() -> void override { gpio::reset(hardware.sync_out); }
+    auto check_tip_presence() -> bool override {
+        if (hardware.tip_sense.has_value()) {
+            return gpio::is_set(hardware.tip_sense.value());
+        }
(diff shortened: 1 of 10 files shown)


include/sensors/firmware/sensor_hardware.hpp
```diff
@@ -13,6 +13,12 @@ class SensorHardware : public SensorHardwareBase {
         : hardware(hardware) {}
     auto set_sync() -> void override { gpio::set(hardware.sync_out); }
     auto reset_sync() -> void override { gpio::reset(hardware.sync_out); }
+    auto check_tip_presence() -> bool override {
+        if (hardware.tip_sense.has_value()) {
+            return gpio::is_set(hardware.tip_sense.value());
```
sfoster1, line 18: can't wait for this https://en.cppreference.com/w/cpp/utility/optional/and_then

sfoster1 (approved):
Looks good, definitely makes more sense than having it in the motor class!

PR #590, opened 2023-03-02
feat(pipettes): add baseline sensor response

caila-marashaj:
This pr just adds a `BaselineSensorRespone` message. This is intended to replace the stream of `ReadFromSensorResponse` messages in the case of auto-zeroing. However, we do want to keep limited polling available for testing and dev purposes, so some additional firmware to differentiate between that and production auto-zeroing may be in order for the sensors.


sfoster1 (approved):
Looks good to me!

vegano1 (approved):

PR #591, opened 2023-03-03
refactor(pipettes): remove the proto builds of pipette firmware

Laura-Danielle:
There aren't even any full builds of proto pipettes left in the office so it should be completely safe to finally remove this! That way we don't need to modify feature development in two places anymore


sfoster1 (approved):
what a diffstat. you love to see it

PR #592, opened 2023-03-06
build: fix version generation in OE

sfoster1:
Openembedded puts the build directories outside the git repo. The source directories should still be in there, so let's make sure we run generate_version in the source directory - this will still work fine for output because we give explicit output directories.

## testing
- [x] do a normal build locally and check that the sha gets found
- [x] do a `cmake --preset=cross-no-directory-reqs` with a `-B` somewhere outside the ot3-firmware directory, build that, and check that the sha is found
- [ ] do an OE build and check that the sha is found

### DIFF
--- a/common/core/CMakeLists.txt
+++ b/common/core/CMakeLists.txt
@@ -1,5 +1,6 @@
 add_custom_target(generate_version
         ${CMAKE_COMMAND} -DGENERATE_VERSION_TARGET_DIR=${CMAKE_CURRENT_BINARY_DIR} -P ${CMAKE_CURRENT_SOURCE_DIR}/generate_version.cmake
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
         BYPRODUCTS ${CMAKE_CURRENT_BINARY_DIR}/version.c)
 
 add_library(version-lib STATIC ${CMAKE_CURRENT_BINARY_DIR}/version.c)


ryanthecoder (approved):
looks good LGTM assuming the SHA shows up in the CI build

vegano1 (approved):

vegano1:
ditto, I can confirm once the build is done.

vegano1:
@sfoster1 
Looking good, heres the deviceInfoResponse from gantry-x before and after installing the binary included with this build, notice that the version, flags, and shortshas are now correct. 

**Before**
2023-03-07 15:56:31.498678 device_info_response (gantry_x->host)
 	message_index: UInt32Field(value=0)
 	version: UInt32Field(value=0)
 	flags: VersionFlagsField(value=)
 	shortsha: FirmwareShortSHADataField(value=b'no-hash\x00')
 	revision: OptionalRevisionField(primary=b, secondary=1, tertiary=None)

**After**
choose MessageId (? for list): 2023-03-07 16:00:07.070391 device_info_response (gantry_x->host)
 	message_index: UInt32Field(value=0)
 	version: UInt32Field(value=2)
 	flags: VersionFlagsField(value=BUILD_IS_EXACT_COMMIT)
 	shortsha: FirmwareShortSHADataField(value=b'329cdadc')
 	revision: OptionalRevisionField(primary=b, secondary=1, tertiary=None)

sfoster1:
Thanks for testing @vegano1 ! Guess I can merge it

PR #594, opened 2023-03-07
build: add workflow for ot3 builds

sfoster1:
This workflow should hook into the same mechanics that do builds for the monorepo and cause ot3 builds on pushes to main and pull requests that are labeled with 'ot3-build'.


Laura-Danielle (approved):
lgtm

sfoster1:
https://github.com/Opentrons/oe-core/actions/runs/4355585698 is a build started by the test pr #595

PR #597, opened 2023-03-08
build: add push script

sfoster1:
This is a python script (for cross-platform compat since we don't have make here) that lets you push all the firmware to a flex. The firmware goes to /usr/lib/firmware and gets picked up by the robot server when you start an update.

### DIFF
--- a/push
+++ b/push
@@ -1,0 +1,150 @@
+#!/usr/bin/env python3
+
+"""
+Push a firmware set to an OT-3.
+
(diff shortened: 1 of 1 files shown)


fsinapi (approved):
🆒

Laura-Danielle (approved):
lgtm

vegano1:
very nice, tested on the flex

vegano1 (approved):

PR #600, opened 2023-03-14
feat(rear-panel): add enable/disable estop and sync

ryanthecoder:
This PR adds the enable/disable messages for estop and sync,
it also adds the handlers to system_task

Tested and working with https://github.com/Opentrons/opentrons/pull/12290 via the new usb_comm script and probing the lines with a multimeter


sfoster1 (changes requested):
Looks good overall, love the gpio header, but I think we might want to rename some stuff and make others defines sadly

sfoster1 (approved):
Looks good to me!

PR #601, opened 2023-03-15
feat(pipettes): add GPIO EXTI interrupt support for the tip sense GPIOs

Laura-Danielle:
## Overview

We want to decide that a tip has been detected based on an EXTI interrupt. This PR adds the piping in HAL to support an EXTI interrupt. While making these changes, I also edited the code on the pressure sensor to support two data ready lines for the 96 channel. We'll also want to do this for DVT 8 channel pipettes since there are now two separate data ready lines.

## Test Plan

- [x] Test that the pressure sensor still functions normally on the single/eight channel

### DIFF
--- a/pipettes/firmware/hardware_config.c
+++ b/pipettes/firmware/hardware_config.c
@@ -33,10 +33,21 @@ static PipetteHardwarePin get_gpio_ht(PipetteHardwareDevice device) {
     }
 }
 
-IRQn_Type get_interrupt_line(const PipetteType pipette_type) {
-    switch (pipette_type) {
-        case NINETY_SIX_CHANNEL:
-        case THREE_EIGHTY_FOUR_CHANNEL:
+
+static IRQn_Type get_interrupt_line_lt(PipetteGPIOInterrupt gpio_pin_type) {
+    switch (gpio_pin_type) {
+        case pipette_gpio_data_ready:
+            return EXTI3_IRQn;
+        case pipette_gpio_tip_sense:
+        default:
+            // External interrupt line 3
+            return EXTI3_IRQn;
+    }
+}
+
+static IRQn_Type get_interrupt_line_ht(PipetteGPIOInterrupt gpio_pin_type) {
+    switch (gpio_pin_type) {
+        case pipette_gpio_data_ready:
             // External interrupt lines 15:10
             // TODO(lc: 10-20-2022) This will only
             // return the front sensor board data ready line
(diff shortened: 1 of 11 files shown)


pipettes/firmware/utility_gpio.c
```diff
@@ -166,9 +169,9 @@ static void data_ready_gpio_init() {
         GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
         HAL_GPIO_Init(hardware.port, &GPIO_InitStruct);
 
-        /* EXTI interrupt init*/
-        HAL_NVIC_SetPriority(exti_line, 10, 0);
-        HAL_NVIC_EnableIRQ(exti_line);
+        // /* EXTI interrupt init*/
```
fsinapi, line 172: Guessing this should be uncommented before merging?

pipettes/firmware/stm32g4xx_it.c
```diff
@@ -118,12 +118,37 @@ void DebugMon_Handler(void) {}
 /*  file (startup_stm32g4xxxx.s).                                             */
 /******************************************************************************/
 
+void EXTI2_IRQHandler(void) {
+    EXTI->PR1 = EXTI_LINE_2;
```
fsinapi, line 122: Similar to checking `PR1`, can we clear using `__HAL_GPIO_EXTI_CLEAR_FLAG`?

pipettes/firmware/hardware_config.c
```diff
@@ -33,23 +33,45 @@ static PipetteHardwarePin get_gpio_ht(PipetteHardwareDevice device) {
...
             // PC6 -> GPIO_EXTI6 (EXTI9_5_IRQn)
             // PC11 -> GPIO_EXTI11 (EXTI15_10_IRQn)
-            return EXTI15_10_IRQn;
+            return EXTI15_10_IRQn | EXTI9_5_IRQn;
+        case pipette_gpio_tip_sense:
+        default:
+            // tip sense
+            // PC7 -> GPIO_EXTI6 (EXTI9_5_IRQn)
+            // PC12 -> GPIO_EXTI11 (EXTI15_10_IRQn)
+            return EXTI15_10_IRQn | EXTI9_5_IRQn;
```
fsinapi, line 61: I don't think that this is going to give you the behavior you want. The `IRQn` enums are not single-bit values, but instead indices into the NVIC bitfields. Returning two IRQn values OR'd together is probably going to end up enabling a completely different IRQ.

fsinapi (changes requested):
Had a few comments but the big one is how `get_interrupt_line_ht` tries to return two different IRQ lines, I'm pretty sure we're going to have to use a different approach there.

Laura-Danielle:
Yeah I needed to test this on the firmware to see if that would actually work. I know you can combine GPIOs but it wasn't clear to me in the NVIC documentation whether the same thing would work or not. I can just change the implementation (and probably make the code more readable by doing so.)

sfoster1 (approved):
Looks good to me!

caila-marashaj (approved):
Exciting!!

Laura-Danielle:
Tested liquid sensing, auto calibration and the new tip sense functionality on a robot. Everything is functioning as expected ✔️

PR #602, opened 2023-03-16
feat(can): add tip presence notification CAN message

Laura-Danielle:
## Overview

To support detecting tip presence (which is an ejector stage flag) we will need to have a CAN notification message.


vegano1 (approved):
Looks good

PR #603, opened 2023-03-16
feat(gripper): update new gripper eeprom chip

ryanthecoder:
Adds the new DR variant of the chip and use that for the rev c version of the gripper board

### DIFF
--- a/include/gripper/core/gripper_info.hpp
+++ b/include/gripper/core/gripper_info.hpp
@@ -90,7 +90,8 @@ class GripperInfoMessageHandler : eeprom::accessor::ReadListener {
     void visit(const InstrumentInfoRequest &m) {
         // Start a serial number read. Respond with CAN message when read
         // completes.
-        // serial_number_accessor.start_read();
+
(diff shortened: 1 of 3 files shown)


include/gripper/core/gripper_info.hpp
```diff
@@ -90,7 +90,8 @@ class GripperInfoMessageHandler : eeprom::accessor::ReadListener {
     void visit(const InstrumentInfoRequest &m) {
         // Start a serial number read. Respond with CAN message when read
         // completes.
-        // serial_number_accessor.start_read();
+
+#if PCBA_PRIMARY_REVISION == 'b'
```
Laura-Danielle, line 94: Is there an "a" revision of the gripper?

Laura-Danielle (approved):
lgtm, except I did have a quick question about the revisions available on the gripper.

include/gripper/core/gripper_info.hpp
ryanthecoder: we have a proto board but I don't think any of them exist anymore

PR #604, opened 2023-03-16
fix(gripper): collection of fixes found during bringup

ryanthecoder:
Was testing on the bench setup and found a few issues with the C1 version of the firmware,
- motor driver setting flag we don't need anymore needed to be cleared so the motor could move
         * old flag was a hack to get the z motor not to fall down when motor disabled
- add a common gpio pin define since they didn't match in different parts of the code
         * there was a mismatch since we used the GPIO_PIN_* in differed places
- remove an IRQ setup since we don't have downstream timers for the Z motor encoder
         *caused a crash when encoder was conencted

Bench setup was tested with the following
all z motor hardware connected and talking via can
called home via ot3repl and saw z motor moving
used limit switch to complete home
moved it relative in ot3repl


fsinapi (approved):
Nice

PR #605, opened 2023-03-16
fix(gripper): reduce gripper error spam

ryanthecoder:
hardware said we should switch this to 2mm to make errors less spamy until further testing can find a better value

### DIFF
--- a/include/motor-control/core/brushed_motor/error_tolerance_config.hpp
+++ b/include/motor-control/core/brushed_motor/error_tolerance_config.hpp
@@ -5,7 +5,7 @@
 namespace error_tolerance_config {
 
 // upon advice from hardware, 0.01mm is a good limit for precision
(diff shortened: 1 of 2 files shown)


Laura-Danielle (approved):

include/motor-control/core/brushed_motor/error_tolerance_config.hpp
ahiuchingau, line 1: Could we change the acceptable distance tolerance as well? To like 1 mm if not 2? 

We've been using 5 mm for both for science. I think we'll need to update the tests as well

ahiuchingau (dismissed):
Could we change the acceptable distance tolerance as well? To like 1 mm if not 2?

We've been using 5 mm for both for science. I think we'll need to update the tests as well

PR #606, opened 2023-03-17
feat(rear-panel):  bring up estop switch monitor task

ryanthecoder:
Tested and working

The new task monitors the GPIO input lines for estop in and estop button detection and sends a message over usb when the state changes

The first part of this will mimic the estop detection message that CAN devices send, and will allow the rear-panel to update the status indicator lights when estop has been activated.

The second part of this sends a message to the hardware controller when an estop button is added to the system and which of the Aux ports it is attached to.


include/rear-panel/core/tasks/hardware_task.hpp
```diff
@@ -0,0 +1,117 @@
...
+ * The task entry point.
+ */
+
+struct GpioInputState {
+    bool estop_in = false;
+    bool estop_aux1_det = false;
+    bool estop_aux2_det = false;
+    // TODO:
+    // door switch
+    // Aux device present
```
sfoster1, line 27: separate from estop present? or is this out of date now

sfoster1 (approved):
Looks good, but please put a more in-depth explanation of what this does and why in the PR description and subsequent squash commit

include/rear-panel/core/tasks/hardware_task.hpp
ryanthecoder: Aux device is a separate pin that will be used by Hepa etc.

PR #607, opened 2023-03-17
feat(pipettes): support tip ejector notification messages

Laura-Danielle:
## Overview

Now that we have a hardware interrupt available for the tip ejector photointerrupter, we should use
that information to notify freertos to send a CAN notification that the status has changed.
Currently, I am only handling one sensor right now but I will support the 96 channel in a follow-up.


sfoster1 (approved):
It looks good, but like with tool notifications we're not really well set up to handle asynchronous notifications from firmware - can we add a can message the host can send that will trigger a response of this message with the current tip-presence state?

Laura-Danielle:
> It looks good, but like with tool notifications we're not really well set up to handle asynchronous notifications from firmware - can we add a can message the host can send that will trigger a response of this message with the current tip-presence state?

Moving this out of a DM. I am planning on tackling Asychronous notification handling this coming week because we have some other dependencies directly related to tool detection.

Both tool detection and tip status will rely on this new methodology going forward.

PR #608, opened 2023-03-17
chore(rear-panel): add CI workflow

fsinapi:
Adding a CI workflow and some other little changes

* Added a github workflow configuration to build + test the rear panel on pushes/PR's
* Fixed the format target (was not working)
* Changed name of the `build-and-test` target to prefix with `rear-panel` instead of `rear_panel`


sfoster1 (approved):
Nice, thank you!

PR #609, opened 2023-03-20
feat(rear-panel): add light task

fsinapi:
Adds a light task to the rear panel board. The light task is responsible for controlling the lights. Right now it just sets the deck light and the white status bar LED on.

The message queue for the Light Control Task includes an empty message that instructs the task to update the LED's. The message is sent every 5ms by a FreeRTOS timer. Right now the task always sets the lights to the same value, but in the future the periodic updates will be used for animations.

Also fixed up the hardware timer initialization so that all of the LED outputs work. Tested by modifying the task to have an incrementing `uint8_t` that would set one of the 5 outputs. For each output, the light would pulse in a sawtooth pattern as expected.

### DIFF
--- a/rear-panel/firmware/main.cpp
+++ b/rear-panel/firmware/main.cpp
@@ -47,42 +47,6 @@ class EEPromHardwareInterface
 static auto eeprom_hw_iface = EEPromHardwareInterface();
 */
 
-static constexpr uint32_t TASK_STACK_DEPTH = 512;
-
-class LED_BLINK_TASK {
-  public:
-    LED_BLINK_TASK(const char* task_name) : task_name{task_name} {}
-    LED_BLINK_TASK(const LED_BLINK_TASK& c) = delete;
-    LED_BLINK_TASK(const LED_BLINK_TASK&& c) = delete;
-    auto operator=(const LED_BLINK_TASK& c) = delete;
-    auto operator=(const LED_BLINK_TASK&& c) = delete;
-    ~LED_BLINK_TASK() = default;
-
-    void start_task() {
-        xTaskCreateStatic(this->task_function, task_name, TASK_STACK_DEPTH,
-                          (void*)1, 6, backing.data(), &static_task);
-    }
-
-    /**
-     * Task entry point.
-     */
-    [[noreturn]] static void task_function(void* unused) {
-        std::ignore = unused;
-        for (;;) {
-            vTaskDelay(500);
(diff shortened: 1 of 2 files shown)


include/rear-panel/core/tasks/light_control_task.hpp
```diff
@@ -0,0 +1,90 @@
...
+            if (queue.try_read(&message, queue.max_delay)) {
+                handle_message(message, hardware);
+            }
+        }
+    }
+
+    [[nodiscard]] auto get_queue() const -> QueueType& { return queue; }
+
+  private:
+    auto handle_message(const TaskMessage& message,
```
Laura-Danielle, line 70: This works just fine, but we've been separating out the concept of "message handler" vs "task entry point" for many of our other tasks.

include/rear-panel/core/tasks/light_control_task.hpp
```diff
@@ -0,0 +1,90 @@
...
+                        LightControlInterface& hardware) -> void {
+        std::visit([this, &hardware](auto m) { this->handle(m, hardware); },
+                   message);
+    }
+
+    auto handle(std::monostate&, LightControlInterface&) -> void {}
+
+    auto handle(rearpanel::messages::UpdateLightControlMessage&,
```
Laura-Danielle, line 78: If you separated this into a message handler, you could save an instance of the light hardware to that message handler rather than having to pass it into the message handle function.

I don't really have any strong opinions, but again, it's just the pattern we've been following for a bit now.

Laura-Danielle (approved):
Looks great to me! One comment about the task pattern you have, but ultimately it doesn't really matter and you can ignore.

rear-panel/firmware/main.cpp
```diff
@@ -83,19 +83,21 @@ class LED_BLINK_TASK {
...
 static auto lbt = LED_BLINK_TASK{"Blinkenlights"};
+static auto light_hardware = light_control_hardware::LightControlHardware();
 
 auto main() -> int {
     HardwareInit();
     // RCC_Peripheral_Clock_Select();
     utility_gpio_init();
+    light_hardware.initialize();
 
     // initialize_leds();
```
ryanthecoder, line 94: We don't need this anymore

fsinapi: Talked in person, deleted the commented out `initialize_leds` call

PR #610, opened 2023-03-20
feat(rear-panel): bring up door switch in the rear-panel

ryanthecoder:
Simple PR that maintains the state of the door switch, and fires off a message is the state changes

also adds an endpoint to request the current state of the door switch


include/rear-panel/core/bin_msg_ids.hpp
```diff
@@ -18,6 +18,8 @@ enum class MessageType : uint16_t {
     RELEASE_SYNC_REQUEST = 0x0A,
     ESTOP_STATE_CHANGE = 0x0B,
     ESTOP_BUTTON_DETECTION_CHANGE = 0x0C,
+    DOOR_SWITCH_STATE_REQUEST = 0x0D,
+    DOOR_SWITCH_STATE_RESPONSE = 0x0E,
```
sfoster1, line 22: we should probably name this something that makes it clearer it can be asynchronously sent by the rear panel

sfoster1 (approved):
Looks good once tested, maybe rename that message though

PR #611, opened 2023-03-21
build: add a workflow to make gh releases from tags

sfoster1:
It will add artifacts to the releases and autogenerate some changelogs also.

Here's an example of a release created by this workflow: https://github.com/Opentrons/ot3-firmware/releases/tag/untagged-3121c9b9dc94a5ac5aee

It will stay as a draft until somebody clicks release, but will automatically get the release artifacts uploaded to it.


Laura-Danielle (approved):
lgtm

fsinapi (approved):
Nice, this is really cool.

PR #612, opened 2023-03-21
build(rear-panel): autogenerate comm protocol constants

fsinapi:
Adds the rear-panel binary message protocol to the `generate_headers.py` script and integrates it with cmake workflows. Also generated the header and fixed references to it in code.

Also fixed a little issue where the autogen script was adding an extra newline that the formatter doesn't like.


ryanthecoder (approved):
Looks good, LGTM

PR #613, opened 2023-03-21
feat(pipettes): support the secondary pressure sensor on 8 and 96channel pipettes

Laura-Danielle:
## Overview

Now that we are in DVT phase of hardware. We want to be able to test the sensors connected to the front channel of the 8ch pipette. This should also give us the 96 channel pressure sensor for free, though we won't be able to test the pressure sensor properly on the 96 channel until we have DVT hardware for that.


pipettes/firmware/main.cpp
```diff
@@ -1,5 +1,7 @@
 // clang-format off
 #include "FreeRTOS.h"
+#include <typeinfo>
```
fsinapi, line 3: What do we need `typeinfo` for?

include/sensors/core/message_handlers/sensors.hpp
```diff
@@ -76,7 +76,15 @@ class SensorHandler {
                 break;
             }
             case can::ids::SensorType::pressure: {
-                client.send_pressure_sensor_queue(m);
+                switch (id) {
+                    case can::ids::SensorId::S1: {
```
fsinapi, line 80: Is there protection to keep us from dereferencing a `null` queue pointer if there's only one pressure sensor on the pipette and a CAN message tries to specify the one that isn't configured? If not, we should have a NULL check in the `send_pressure_sensor_queue` functions.

fsinapi (approved):
Looks good! Had a question about a new import and I think there might be a possible `null`-dereference case (but if I'm wrong then this looks good to go!)

include/sensors/core/message_handlers/sensors.hpp
Laura-Danielle: good call!

include/pipettes/core/sensor_tasks.hpp
```diff
@@ -73,23 +85,27 @@ struct QueueClient : can::message_writer::MessageWriter {
...
+    void send_pressure_sensor_queue_rear(const sensors::utils::TaskMessage& m);
+
+    void send_pressure_sensor_queue_front(const sensors::utils::TaskMessage& m);
 
     freertos_message_queue::FreeRTOSMessageQueue<eeprom::task::TaskMessage>*
         eeprom_queue{nullptr};
     freertos_message_queue::FreeRTOSMessageQueue<sensors::utils::TaskMessage>*
         environment_sensor_queue{nullptr};
     freertos_message_queue::FreeRTOSMessageQueue<sensors::utils::TaskMessage>*
-        capacitive_sensor_queue_front{nullptr};
```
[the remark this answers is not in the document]
Laura-Danielle: Not yet. The primary sensor is actually on the "rear" part of the pipette so I changed the name in this pr.

I'm working on a separate thing to bring up the secondary capacitive sensor.

pipettes/firmware/main.cpp
```diff
@@ -121,24 +122,31 @@ void encoder_callback(int32_t direction) {
...
+    utility_configs::sensor_configurations<PIPETTE_TYPE>();
+static auto sensor_hardware_container =
+    utility_configs::get_sensor_hardware_container(pins_for_sensor);
 
-static auto sensor_hardware =
-    sensors::hardware::SensorHardware(pins_for_sensor.primary);
 static auto data_ready_gpio_primary =
     pins_for_sensor.primary.data_ready.value();
+static auto data_ready_gpio_secondary =
+    pins_for_sensor.primary.data_ready.value();
```
caila-marashaj, line 132:
```suggestion
    pins_for_sensor.secondary.data_ready.value();
```

Laura-Danielle: good catch!

PR #614, opened 2023-03-22
chore(pipettes): disable the pressure sensor for C2 pipette revisions

Laura-Danielle:
## Overview

We do not have filtering on the data ready line and thus the pressure sensor can be noisy. This causes issues with auto-calibration and so for now we should disable the pressure sensor.

### DIFF
--- a/pipettes/firmware/main.cpp
+++ b/pipettes/firmware/main.cpp
@@ -134,8 +134,14 @@ static auto tip_sense_gpio_primary = pins_for_sensor.primary.tip_sense.value();
 
 static auto& sensor_queue_client = sensor_tasks::get_queues();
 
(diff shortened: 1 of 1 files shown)


pipettes/firmware/main.cpp
```diff
@@ -134,15 +134,22 @@ static auto tip_sense_gpio_primary = pins_for_sensor.primary.tip_sense.value();
 
 static auto& sensor_queue_client = sensor_tasks::get_queues();
 
+#if PCBA_PRIMARY_REVISION == 'c' && PCBA_SECONDARY_REVISION == '2'
+static bool pressure_sensor_available = false;
```
sfoster1, line 138:
```suggestion
static constexpr bool pressure_sensor_available = false;
```

sfoster1 (approved):
Looks good but let's make them constexpr

PR #615, opened 2023-03-22
fix(gripper): add encoder timer interrupt handler & fix encoder dir

ahiuchingau:
During initial DVT gripper testing, @ryanthecoder and I saw that the z encoder wasn't behaving correctly.

We found out that (1) encoder timer interrupt was not added, and the (2) encoder direction was flipped. This PR fixes the issues we saw on the Z motor.


sfoster1 (approved):
Looks good to me once tested!

PR #616, opened 2023-03-22
chore(pipettes): disable the primary pressure sensor for C2 pipette revisions

Laura-Danielle:
## Overview

We do not have filtering on the data ready line and thus the pressure sensor can be noisy. This causes issues with auto-calibration and so for now we should disable the pressure sensor.


sfoster1 (approved):
yay

PR #617, opened 2023-03-23
feat(rear-panel): light action handler

fsinapi:
This PR adds in an AnimationHandler class that is responsible for handling rear panel light animations. The AnimationHandler is backed by an AnimationBuffer (a double-buffered array) which stores a series of individual "actions". Each action contains:
- A color to transition to
- A transition type
  - Instant will instantly change the color and then hold at it for the remainder of the step
  - Linear will just change the color linearly over time from the previous setting to the new setting
  - Sinusoid will transition in a smoother, pulse-like fashion between the previous color and the new color
- A transition time, which is the amount of time in milliseconds that this step should take

The double buffer is utilized to provide an "active" queue, for the animation that is currently running; and a "staging" queue, where individual Actions can be added to create an animation. When the controller wants to run a new animation, it can send all of the individual steps over USB (in a future PR) and then request to start an animation (either looping or single shot), which will swap the staging buffer to be the active buffer and begin running through it.

Also requires https://github.com/Opentrons/opentrons/pull/12345

### DIFF
--- a/include/rear-panel/core/lights/animation_handler.hpp
+++ b/include/rear-panel/core/lights/animation_handler.hpp
@@ -9,7 +9,12 @@
 namespace lights {
 
 /**
- * @brief Acts as an interface for an AnimationQueue
+ * @brief Acts as an interface for an AnimationQueue.
+ * 
+ * @note This class if NOT thread safe - it accesses the same queue in
+ * multiple functions and, if they are called from different threads,
+ * race conditions may arise. If this class is accessed from different
+ * threads, wrap access to all member functions with a mutex.
  *
  */
 template <size_t Size>
--- a/include/rear-panel/core/lights/animation_handler.hpp
+++ b/include/rear-panel/core/lights/animation_handler.hpp
@@ -44,8 +44,8 @@ class AnimationHandler {
             ret = calculate_step_power();
             if (_timer_ms >= _current_step.value().transition_time_ms) {
                 _step_starting_color = _current_step.value().color;
-                _timer_ms = 0;
-                _current_step = _queue.get_next_active_step();
+                // Signal that we need a new message next time through
+                _current_step = std::nullopt;
             }
         }
         _most_recent_color = ret;
--- a/include/rear-panel/core/lights/animation_handler.hpp
+++ b/include/rear-panel/core/lights/animation_handler.hpp
@@ -41,7 +41,10 @@ class AnimationHandler {
         if (_current_step.has_value()) {
             // We are in a step, calculate the next increment
             _timer_ms += time_increment_ms;
-            ret = calculate_step_power();
+            ret = lights::math::color_interpolate(
+                _step_starting_color, _current_step.value().color,
+                _current_step.value().transition, _timer_ms,
+                _current_step.value().transition_time_ms);
             if (_timer_ms >= _current_step.value().transition_time_ms) {
                 _step_starting_color = _current_step.value().color;
                 // Signal that we need a new message next time through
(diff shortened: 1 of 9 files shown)


include/rear-panel/core/lights/animation_handler.hpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    auto add_to_staging(const Action &action) -> bool {
+        return _queue.add_to_staging(action);
+    }
+
+    auto clear_staging() -> void { _queue.clear_staging(); }
+
+  private:
+    auto calculate_step_power() -> Color {
```
sfoster1, line 72: if we made this pure and called it like `color_interp(Color first, Color second, TransitionType transition, (int start, int end) | (float progress)) -> Color` it could even also go into `color::math`

rear-panel/core/lights/animation_math.cpp
```diff
@@ -0,0 +1,56 @@
...
+    -> double {
+    // Simple linear regression. At count of 0ms we return start_power, and at
+    // the end we return the end_power.
+    double slope = end_power - start_power;
+    double scale =
+        static_cast<double>(ms_count) / static_cast<double>(ms_total);
+    return (scale * slope) + start_power;
+}
+
+static auto calculate_sin_power(double start_power, double end_power,
```
sfoster1, line 20: based on my confusion and the number of comments I had to delete, can you annotate this function with the part of the range of `cos` that we're using (i.e. peak-to-peak including the zero-crossing)

rear-panel/core/lights/animation_math.cpp
```diff
@@ -0,0 +1,56 @@
...
+                                uint32_t ms_count, uint32_t ms_total)
+    -> double {
+    double slope = end_power - start_power;
```
sfoster1, line 23: nit: `slope` is kind of a bad name for this, it would only be the slope if it was divided by the time. can we call it `powerdiff` or something

sfoster1 (changes requested):
The functionality looks good but I have some thoughts about the design - I think we can make it clearer at first glance.

include/rear-panel/core/lights/animation_handler.hpp
```diff
@@ -0,0 +1,99 @@
...
+        if (!_current_step.has_value()) {
+            // We were not in an active animation last update, but
+            // one may have started since then. Try to get a new step.
+            _timer_ms = 0;
+            _current_step = _queue.get_next_active_step();
+        }
+
+        if (_current_step.has_value()) {
+            // We are in a step, calculate the next increment
+            _timer_ms += time_increment_ms;
```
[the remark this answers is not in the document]
fsinapi: The issue with that approach is that it ends up repeating states. Example, if you have two steps that are 3ms long and we tick at 1ms:

- First step: [0, 1, 2, 3]
- Second step [0, 1, 2, 3]

But the state of the first step at 3 is the same as the second step at 0. 

Since we're now [only grabbing](https://github.com/Opentrons/ot3-firmware/pull/617#discussion_r1146214499) from the queue at the start of the `animate` function, you end the step when the counter equals the total transition time but then the next time through you have to start at time = 0.

Agree that, either way, it should look the same to any human.

include/rear-panel/core/lights/animation_handler.hpp
```diff
@@ -0,0 +1,99 @@
...
+                _step_starting_color = _current_step.value().color;
+                _timer_ms = 0;
+                _current_step = _queue.get_next_active_step();
+            }
+        }
+        _most_recent_color = ret;
+        return ret;
+    }
+
+    auto start_staged_animation(AnimationType animation) -> void {
```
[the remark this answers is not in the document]
fsinapi: We could add mutex protection to the queue, but in practice it's only going to be called from message handlers from a single task that can't interrupt itself. Maybe we can just add some language that this isn't a thread-safe class?

include/rear-panel/core/lights/animation_handler.hpp
fsinapi: 💯

rear-panel/core/lights/animation_math.cpp
fsinapi: lol

include/rear-panel/core/lights/animation_handler.hpp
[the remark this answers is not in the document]
fsinapi: Ahhhhh actually, after the pseudo-visitor refactor you suggested, we can do the following:
* When grabbing a new step from the queue, set the timer to the increment amount instead of 0
* Increment the timer _after_ calculating the interpolation

And we get the correct behavior without skipping anything. Calling `animate` gives the values for the _current_ timer and then advances it.

include/rear-panel/core/lights/animation_queue.hpp
```diff
@@ -0,0 +1,124 @@
...
+ * The AnimationQueue holds both an active queue and a staging queue. New
+ * actions are added to the staging queue linearly, and then when a new
+ * animation is started the staging queue is swapped with the active queue.
+ * This allows a new animation to be constructed in the background while the
+ * current animation continues running.
+ *
+ * @tparam Size The maximum number of elements in the queue. Must be nonzero.
+ */
+template <size_t Size>
+class AnimationQueue {
```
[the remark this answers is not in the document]
fsinapi: Yeah this sounds 1000% nicer, will refactor this guy.

sfoster1 (approved):
Looks great to me!

PR #618, opened 2023-03-23
chore: add rear panel and manifest to application artifact

ryanthecoder:
The previous file only installed .hex files but we also want the rear panel's .bin file and the manifest .json file.

with this new workflow you can download the application artifact and unzip it into your ot3firmware repo and push it onto the bot with <code>./push [IP] --no-build</code>


sfoster1 (approved):
Noice

PR #619, opened 2023-03-23
feat(rear-panel): add light strip commands

fsinapi:
This PR adds three commands for interacting with the light strip on the rear panel:

- A command to add a step to the animation in the staging queue
- A command to clear the staging queue
- A command to start running the animation in the staging queue

An `AnimationHandler` is added to the light control task's message handler, and it is used as the source of the color values when the periodic update message is parsed. 

Tested on an OT-3 with https://github.com/Opentrons/opentrons/pull/12350 using the `usb_comm` script. Uploading a series of actions and then running an animation will, in fact, run that animation.

Other misc stuff
- Bumped up the stack size of the idle task and the timer task; I was getting overflows from them and this fixed it.
- Added a handler to enable using the FreeRTOS stack overflow features
- Added implementation of `send_light_control_queue` because I forgot to do that
- Added `nullptr` checking for all of the queue-send commands. The queues _should_ be set before they're ever accessed, but belt-and-suspenders isn't the worst thing...
- Added a bunch of linter-related changes for the AnimationHandler & friends because they weren't being compiled into the firmware before so the linter didn't check them.

### DIFF
--- a/include/rear-panel/core/messages.hpp
+++ b/include/rear-panel/core/messages.hpp
@@ -339,6 +339,70 @@ struct DoorSwitchStateInfo
     auto operator==(const DoorSwitchStateInfo& other) const -> bool = default;
 };
 
+struct AddLightActionRequest
+    : BinaryFormatMessage<rearpanel::ids::BinaryMessageId::add_light_action> {
+    static constexpr size_t LENGTH =
+        sizeof(uint16_t) + sizeof(rearpanel::ids::LightTransitionType) +
+        (4 * sizeof(uint8_t));
+    uint16_t length;
+    uint16_t transition_time_ms;
+    rearpanel::ids::LightTransitionType transition;
+    uint8_t red;
+    uint8_t green;
+    uint8_t blue;
+    uint8_t white;
+
+    template <bit_utils::ByteIterator Input, typename Limit>
+    static auto parse(Input body, Limit limit)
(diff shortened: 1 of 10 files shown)


include/rear-panel/core/messages.hpp
```diff
@@ -339,29 +339,158 @@ struct DoorSwitchStateInfo
...
+        uint16_t type = 0;
+        uint8_t transition_buf = 0;
+        auto ret = AddLightActionRequest{
+            .length = 0,
+            .transition_time_ms = 0,
+            .transition = rearpanel::ids::LightTransitionType::linear,
+            .red = 0,
+            .green = 0,
+            .blue = 0,
+            .white = 0};
```
caila-marashaj, line 368: What do these values for each color represent? It's not the PWM that corresponds with each color, right?

fsinapi: The Python definition of the message makes this more explicit, but these are uint8_t scaled from 0 to 255 to match standard RGB color definitions, which is how our design docs specify the colors to use. The max PWM is also scaled to 255 so we just keep the scale the same throughout the code.

fsinapi: Talked offline 👍

sfoster1 (approved):
Looks good to me!

PR #620, opened 2023-03-27
fix(pipettes): switch the i2c bus of the pressure sensor to match primary and secondary conventions

Laura-Danielle:
## Overview

We are using the convention of "primary" and "secondary" to refer to multiple sensors of the same type. The primary channel should generally refer to the "A1" version of the pipette. The eight channel has a flipped "primary" bus for the pressure sensor.


sfoster1 (approved):
Very Cool

PR #621, opened 2023-03-27
feat(head, gantry): enable the eeprom for the head and gripper boards

ryanthecoder:
In order to continue with RLAB-291 The gantry and head boards need to have the eeprom and i2c system setup.
This adds the i2c read/write tasks and the eeprom task so that further PR's can build on top of it

Tested reading and writing from eeprom on a DVT bot


sfoster1 (approved):
Looks good to me if it works!

PR #622, opened 2023-03-27
feat(rear-panel): add deck light messages

fsinapi:
Adds messages to control the deck lights, and adds handling in the host & light tasks to process them.

Also moved `rear_panel_parser` to be statically defined in a source file & updated tests to match.

Tested with [monorepo PR](https://github.com/Opentrons/opentrons/pull/12359) and was able to set the deck lights on & off, and get the status correctly.


Laura-Danielle (approved):
🛋️

PR #623, opened 2023-03-28
fix(pipettes): ensure that secondary hardware exists before trying to access the value

Laura-Danielle:
## Overview

I previously thought calling `.value()` on an optional parameter meant that it would either return a nullptr or the object it was supposed to store. Mais **non**, I was wrong and therefore we have to do this gross check to make sure we don't brick pipette firmware. If anyone has any elegant solutions, definitely let me know.

We should be OK to not include this in the internal release b/c the breaking change was merged in after we cut the v4 tag.

### DIFF
--- a/pipettes/firmware/main.cpp
+++ b/pipettes/firmware/main.cpp
@@ -128,7 +128,9 @@ static auto sensor_hardware_container =
 
 static auto data_ready_gpio_primary =
     pins_for_sensor.primary.data_ready.value();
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Looks good if tested, we can refactor this to be a bit nicer in the future

fsinapi (approved):
If you want to get a `std::nullopt` if there's no value, you could use [value_or](https://en.cppreference.com/w/cpp/utility/optional/value_or) (maybe... I think?). This works too!

PR #624, opened 2023-03-28
fix(bootloader): present pipette information

sfoster1:
The bootlaoder has no way to indicate what kind of pipette it is. This means that if a pipette is stuck in its bootloader, it cannot be automatically and safely updated.

To fix this, let's add a new byte to the device info and put the pipette type in it (for a pipette, anyway). The byte will be interpreted differently depending on the canbus node id that the device info response comes from, so there's no need for a ton of space here.

This is implemented in the bootloader with another attribute-weak hook that will only get compiled for the pipette itself. While I was here, I refactored the way we do node IDs to make things a little better.

Closes RQA-584

## Risks
This is a bootloader change! We need to test it on lots of kinds of devices.
We also need to add handling for this byte to the python side, but that's not that big a deal since we already are accepting of device info responses with more than we think in them.

## Testing
- Put it on pipettes of various types and check that the right byte comes back and that it works
- Put it on other stuff and check that the 0 byte comes back.

### DIFF
--- a/bootloader/core/CMakeLists.txt
+++ b/bootloader/core/CMakeLists.txt
@@ -8,7 +8,6 @@ add_library(bootloader-core
         util.c
         message_handler.c
         update_state.c
-        pipette_node_id.c
         )
 add_dependencies(bootloader-core generate_version)
 target_include_directories(bootloader-core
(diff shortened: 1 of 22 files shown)


Laura-Danielle (approved):
Changes look good. I can test on pipettes tomorrow if you need.

sfoster1:
Corresponding monorepo pr here: https://github.com/Opentrons/opentrons/pull/12383

PR #625, opened 2023-03-29
fix(bootloader): delay before app start on pipette

sfoster1:
On the pipettes, we don't reliably detect that we're on the right mount when we go from bootloader to application. This is because it's too fast. So let's slow it down. By reconfiguring the mount detect pin to tristate and adding a delay after that before going to the app, we make it reliable again.

## Testing
- [x] put it on a pipette on a right mount and go back and forth between application and bootloader a lot


Laura-Danielle (approved):
makes sense and lgtm

fsinapi (approved):
Makes sense. LGTM!

PR #626, opened 2023-03-29
fix(rear-panel): incorrect limiting of msg size

sfoster1:
Missed a test failure that had a real bug! The raw iterator manipulation does not correctly handle the case where we're out of space in our write buffer.


sfoster1:
Also I'm just gonna merge this when it passes

PR #628, opened 2023-03-30
fix(pipettes): mount id and bootloader

sfoster1:
The mount ID for the right mount wasn't working properly in all electrical cases:
- Booting the robot while pipette plugged in
- Bootloader->application

This fixes the issue by moving to a 100ms startup delay between configuring the mount ID as an input, and reading/writing it. That's in both the application and the bootloader, and works under EE testing.


fsinapi (approved):
🕙

caila-marashaj (approved):

PR #629, opened 2023-04-03
fix(gripper): update sensor id for sensor tasks

ahiuchingau:
We've switched the cap sensor IDs for the gripper but we forgot to also update the sensor id on the sensor task, so it's reporting the wrong sensor id.

### DIFF
--- a/gripper/core/tasks.cpp
+++ b/gripper/core/tasks.cpp
@@ -85,11 +85,11 @@ void gripper_tasks::start_tasks(
 
     auto& capacitive_sensor_task_front =
         capacitive_sensor_task_builder_front.start(
-            5, "capacitive sensor S1", i2c2_task_client, i2c2_poll_client,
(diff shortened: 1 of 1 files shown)


gripper/core/tasks.cpp
```diff
@@ -85,11 +85,11 @@ void gripper_tasks::start_tasks(
 
     auto& capacitive_sensor_task_front =
         capacitive_sensor_task_builder_front.start(
-            5, "capacitive sensor S0", i2c2_task_client, i2c2_poll_client,
+            5, "capacitive sensor S1", i2c2_task_client, i2c2_poll_client,
```
fsinapi, line 88: Not related to this PR, but we should change the name of these tasks to "cap sensor Sn" since our task name length is only 16 chars

fsinapi (approved):
🆔

Laura-Danielle (approved):
si

PR #630, opened 2023-04-03
refactor(include): modify the cap sensor task to take readings from different CINs

Laura-Danielle:
## Overview

The eight channel pipette is unique insofar as there is only a single cap sensor with two flex
cables connected to two different channels. Since we had to change up some of the register values, I
decided to go ahead and refactor the code to be more uniform with the pressure/environment sensor.
During the refactor I realized that the baseline sensor request does the same thing as auto zero
threshold...(duh).. and we should refactor the code to 1. remove the baseline sensor request message
and 2. ensure that all of the sensors have the same functionality in terms of auto-zeroing.

## TODO

- [x] Thoroughly test on robot
- [x] Add tests for "shared task" to check that it can receive and handle different ids and uses the correct modes in those scenarios

## Review Requests

Please check out the refactor and let me know if you think it's missing anything. There is definitely still some stuff we could probably change around, but I figured this was a good start to get the code a little more organized.


include/sensors/core/fdc1004.hpp
```diff
@@ -27,39 +27,467 @@
...
+};
+
+enum class MeasurementRate : uint8_t {
+    RESERVED = 0x0,
+    ONE_HUNDRED_SAMPLES_PER_SECOND = 0x1,
+    TWO_HUNDRED_SAMPLES_PER_SECOND = 0x2,
+    FOUR_HUNDRED_SAMPLES_PER_SECOND = 0x3,
+};
+
+enum class MeasureConfigMode : uint8_t { ONE, TWO, THREE, FOUR };
```
sfoster1, line 55: if this goes to the device we should define enumeration values explicitly

include/sensors/core/fdc1004.hpp
```diff
@@ -27,39 +27,467 @@
...
+    uint16_t padding_0 : 4 = 0;
+};
+
+struct __attribute__((packed, __may_alias__)) ConfMeasure2 {
+    static constexpr Registers address = Registers::CONF_MEAS2;
+    static constexpr bool readable = true;
+    static constexpr bool writable = true;
+    static constexpr uint32_t value_mask = (1 << 8) - 1;
+
+    // Notes:
```
sfoster1, line 262: i think it's ok to have this comment just say "see CONF_MEAS1"

include/sensors/core/fdc1004.hpp
```diff
@@ -27,39 +27,467 @@
...
+
+struct __attribute__((packed, __may_alias__)) ConfMeasure3 {
+    static constexpr Registers address = Registers::CONF_MEAS3;
+    static constexpr bool readable = true;
+    static constexpr bool writable = true;
+    static constexpr uint32_t value_mask = (1 << 8) - 1;
+
+    // Notes:
+
+    // You can store up to 4 measurement configurations on the cap sensor
```
sfoster1, line 292: "see CON_MEAS1"?

include/sensors/core/fdc1004.hpp
```diff
@@ -27,39 +27,467 @@
...
+
+struct __attribute__((packed, __may_alias__)) ConfMeasure4 {
+    static constexpr Registers address = Registers::CONF_MEAS4;
+    static constexpr bool readable = true;
+    static constexpr bool writable = true;
+    static constexpr uint32_t value_mask = (1 << 8) - 1;
+
+    // Notes:
+
+    // You can store up to 4 measurement configurations on the cap sensor
```
sfoster1, line 320: "see CONF_MEAS1"?

include/sensors/core/fdc1004.hpp
```diff
@@ -27,39 +27,467 @@
...
+
+struct __attribute__((packed, __may_alias__)) FDCConf {
+    static constexpr Registers address = Registers::FDC_CONF;
+    static constexpr bool readable = true;
+    static constexpr bool writable = true;
+    static constexpr uint32_t value_mask = (1 << 8) - 1;
+
+    uint8_t reset : 1 = 0;
+    uint8_t padding_0 : 2 = 0;
+    // Measurement rate Options:
```
sfoster1, line 348: maybe "measurement rate options should be values in MeasurementRate"?

include/sensors/core/fdc1004.hpp
```diff
@@ -27,39 +27,467 @@
...
+    uint16_t offset_decimal : 11 = 0;
+};
+
+struct __attribute__((packed, __may_alias__)) OffsetCalCIN2 {
+    static constexpr Registers address = Registers::OFFSET_CAL_CIN2;
+    static constexpr bool readable = false;
+    static constexpr bool writable = true;
+    static constexpr uint32_t value_mask = (1 << 8) - 1;
+
+    // -16pF -> 16pF range. Used to reduce parasitic capacitance due to
```
sfoster1, line 386: maybe "see OFFSET_CAL_CIN1"?

include/sensors/core/fdc1004.hpp
```diff
@@ -27,39 +27,467 @@
...
+    uint16_t offset_decimal : 11 = 0;
+};
+
+struct __attribute__((packed, __may_alias__)) OffsetCalCIN4 {
+    static constexpr Registers address = Registers::OFFSET_CAL_CIN4;
+    static constexpr bool readable = false;
+    static constexpr bool writable = true;
+    static constexpr uint32_t value_mask = (1 << 8) - 1;
+
+    // -16pF -> 16pF range. Used to reduce parasitic capacitance due to
```
sfoster1, line 410: maybe "see OFFSET_CAL_CIN1"?

include/sensors/core/fdc1004.hpp
```diff
@@ -27,39 +27,467 @@
...
+    uint16_t offset_decimal : 11 = 0;
+};
+
+struct __attribute__((packed, __may_alias__)) GainCalCIN1 {
+    static constexpr Registers address = Registers::GAIN_CAL_CIN1;
+    static constexpr bool readable = false;
+    static constexpr bool writable = true;
+    static constexpr uint32_t value_mask = (1 << 8) - 1;
+
+    uint8_t offset_integer : 5 = 0;
```
sfoster1, line 422: def want a comment here about purpose and format

sfoster1 (commented):
Looking good, IMO we can slightly improve some comment stuff which is pretty important for these register layouts

include/sensors/core/fdc1004.hpp
Laura-Danielle: It does not. It's only used in the `driver` portion of the code.

pipettes/core/sensor_tasks_g4.cpp
```diff
@@ -149,6 +152,23 @@ void sensor_tasks::start_tasks(
     queues.pressure_sensor_queue_front =
         &pressure_sensor_task_front.get_queue();
     queues.tip_notification_queue = &tip_notification_task.get_queue();
+
+    if (shared_cap_task) {
+        // There is only one cap sensor on the eight channel and so the "front"
+        // and "rear" nozzles are actually supported by 1 single task so the
+        // 'front'/'rear' should be set to the same queue/task.
+        tasks.capacitive_sensor_task_front = &capacitive_sensor_task_rear;
```
fsinapi, line 160: I like this a lot, really nice way to transparently share the task.

include/sensors/core/fdc1004.hpp
```diff
@@ -27,39 +27,448 @@
...
+
+template <typename Reg>
+concept ReadableRegister = requires() {
+    {Reg::readable};
+};
+
+struct __attribute__((packed, __may_alias__)) ManufactureId {
+    static constexpr Registers address = Registers::MANUFACTURER_ID;
+    static constexpr bool readable = true;
+    static constexpr bool writable = false;
```
fsinapi, line 144: The way that the `ReadableRegister` and `WritableRegister` concepts work is extremely unintuitive, but it actually evaluates whether this variable _exists_ rather than looking at the value of the constexpr (which isn't accessible in concept evaluation). So any of these registers with `readable = false` or `writable = false` should just have that respective struct member removed. Sorry!

fsinapi (approved):
This is some really nice refactoring. One comment on readable/writable register designation but otherwise LGTM.

caila-marashaj (approved):
Nice! Looks really clean!

PR #631, opened 2023-04-03
fix(eeprom): general filesystem accessor bug fixes

ryanthecoder:
We haven't used the general  filesystem handler yet for anything so this PR fixes a few bugs that cropped up during testing.

We needed to add another config value to the hw_iface for what the default value of eeprom memory was, and fetch that config value before doing anything else

added in some more test converge, and added lots of comments

also fixes a bug where integer promotion resulted in the wrong address being written to the lookup table.


sfoster1 (approved):
Looks good to me!

PR #633, opened 2023-04-07
feat(eeprom): add a startup task that ensures the data table is up to date

ryanthecoder:
The data table tail accessor is now created as a singleton that is shared across the various tasks that need to access the file system
this ensures that the various tasks don't try to edit the table simultaneously.

the new tasks is created at startup with a list of the table table revisions
the task then makes sure that any new data table updates are applied
when it's complete the task exits and signals to the eeprom filesystem that it's ready for use via the shared data table tail accessor

### DIFF
--- a/include/eeprom/core/dev_data.hpp
+++ b/include/eeprom/core/dev_data.hpp
@@ -61,10 +61,7 @@ class DevDataTailAccessor
 
     auto finish_data_rev() -> void { data_rev_finished = true; }
 
-    void read_complete(uint32_t message_index) final {
-        // we don't need message_index since this is an internal call
-        // and not initiated from a can message
-        std::ignore = message_index;
+    void read_complete(uint32_t) final {
         // test if data is set to default values at delivery
         auto delivery_state = std::vector<uint8_t>(
             addresses::lookup_table_tail_length, conf.default_byte_value);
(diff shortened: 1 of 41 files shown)


include/eeprom/core/dev_data.hpp
```diff
@@ -286,45 +333,24 @@ class DevDataAccessor
...
+        if (table_ready()) {
+            return calculate_table_entry_start(key) <
+                   tail_accessor.get_data_tail();
         }
         return false;
     }
 
-    auto ready() -> bool { return tail_updated && config_updated; }
+    auto table_ready() -> bool {
+        return config_updated && tail_accessor.get_tail_updated();
```
Laura-Danielle, line 345: can't this just be `get_tail_updated()` since we're checking whether `config_updated` is true here?

Laura-Danielle: Or, we should pull `config_updated` from `get_tail_updated()` and reserve that check for the `table_ready()` function.

Laura-Danielle (changes requested):
Had a few comments/change requests.

In general, I'm not sure whether we should be defining globals in header files. Was there a particular reason why we chose to do that? Could we maybe use `extern` and define them in a source file instead?

Laura-Danielle (approved):
lgtm! Thanks for the changes :)

PR #634, opened 2023-04-07
refactor(pipettes): remove second sensor task file

Laura-Danielle:
## Overview

Now that we're no longer supporting L5 boards, we do not need to separate the sensor task file into "g4" and "regular" because the "g4" sensor task file is the only one we're using now.


fsinapi (approved):
negative-line-diff pull requests >>>

PR #635, opened 2023-04-10
feat(motor-control): enable lifetime motor distance usage

ryanthecoder:
Whenever a move completes, the travel distance is added to the total distance moved which is stored in the eeprom


Laura-Danielle:
@ryanthecoder Could you point this PR on your other open branch to make reviewing a bit easier?

head/core/tasks_proto.cpp
```diff
@@ -70,6 +70,10 @@ static auto i2c3_poll_client =
     i2c::poller::Poller<freertos_message_queue::FreeRTOSMessageQueue>{};
 static auto eeprom_task_builder =
```
sfoster1, line 71: Let's disable this on non-DVT head boards for mfg error reasons

head/core/tasks_rev1.cpp
```diff
@@ -75,6 +75,10 @@ static auto eeprom_task_builder =
 
 static auto tail_accessor = eeprom::dev_data::DevDataTailAccessor{head_queues};
```
sfoster1, line 77: let's disable on non-dvt

include/motor-control/core/tasks/usage_storage_task.hpp
```diff
@@ -1,7 +1,182 @@
...
+        _ensure_part(m.key, distance_data_usage_len);
+        usage_data_accessor.get_data(m.key, 0);
+    }
+
+    void finish_handle(const usage_messages::IncreaseDistanceUsage& m) {
+        uint64_t old_value = 0;
+        std::ignore = bit_utils::bytes_to_int(
+            accessor_backing.begin(),
+            accessor_backing.begin() + distance_data_usage_len, old_value);
+        if (old_value == 0xFFFFFFFFFFFFFFFF) {
```
sfoster1, line 97: let's make this a constexpr so it's explicitly typed and we're not relying on type promotion

sfoster1 (changes requested):
I think this generally looks good, but after in-person discussion IMO
- usage keys should be in the message; the usage response should maybe house a list of TLVs
- those usage keys should therefore be well-known and generated from python (and don't necessarily have to be the same as the bits written to the eeprom)

head/core/tasks_proto.cpp
ryanthecoder: We don't have the issue on the A rev boards since they use a different eeprom chip than the Rev B boards

include/motor-control/core/motor_hardware_interface.hpp
```diff
@@ -1,13 +1,57 @@
...
+#include <vector>
 
 #include "motor-control/core/types.hpp"
 
 namespace motor_hardware {
 
+static constexpr auto max_requests_per_can_message = 5;
+class UsageEEpromConfig {
+  public:
+    UsageEEpromConfig(const std::vector<UsageRequestSet>& requests) {
```
sfoster1, line 17: Unfortunately `vector` will allocate internally and we can't use it. For something like this, let's use templated iterator pairs and I think that will do the job.

sfoster1 (changes requested):
Mostly looks good to me but unfortunately we really do have to get rid of all the usages of `std::vector` and replace with array or something so we don't allocate.

Laura-Danielle (commented):
small suggestions and one q about a clock enable removal

gantry/firmware/i2c_setup.c
```diff
@@ -16,7 +16,6 @@ void HAL_I2C_MspInit(I2C_HandleTypeDef* hi2c) {
 
     // PIN PC4 is SCL
     // PIN PA8 is SDA
-    __HAL_RCC_GPIOA_CLK_ENABLE();
```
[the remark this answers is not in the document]
ryanthecoder: Duplicate of line 14 above

sfoster1 (approved):
Nice, thanks for the changes! I think this looks good now.

PR #637, opened 2023-04-14
feat(gripper) use eeprom to store Jaw force mode time

ryanthecoder:
This adds another timer to the gripper that counts up when movement on the gripper stops,
if the gripper is in force application mode when the counter overflows it increments its usage in eeprom
if the gripper starts a new move while in force application mode it fetches the current timer count and stores that data
also expands the GetUsageDataResponse to include the force application time

<s>depends on #635 </s>
and https://github.com/Opentrons/opentrons/pull/12499


sfoster1 (approved):
Looks good to me!

PR #638, opened 2023-04-17
refactor(motor-control): re-enable stall detection

ahiuchingau:
This PR adds a new MoveStopCondition called `ignore_stalls`. This can be used together with other conditions and the firmware will ignore and not raise an error if the motor stalls during movement. 

There are 3 different scenarios where a stall is handled:
1. **MoveStopCondition::none**
  This is a regular linear move where we're not expecting to see a stall. Therefore an unrecoverable error will be raised and a stop message will be broadcasted. 
3. **MoveStopCondition::stall**
  This is a move where a stall is expected and will be handled differently (not implemented yet). A recoverable error will be raised and we do not need to send a stop message.
4. **MoveStopCondition::ignore_stalls** 
  This move type should only be used during hardware testing. When a stall occurs, it doesn't raise any error and the rest of the moves in the queue are executed per usual. Note this only works if all of the subsequent moves also have the stall ignored flag. Therefore, it makes sense to use a feature flag on Python side to disable stall detection, where we can simply add the `ignore_stalls` condition to every single linear move message.

### DIFF
--- a/include/bootloader/core/ids.h
+++ b/include/bootloader/core/ids.h
@@ -215,6 +215,7 @@ typedef enum {
     can_movestopcondition_encoder_position = 0x4,
     can_movestopcondition_gripper_force = 0x8,
     can_movestopcondition_stall = 0x10,
+    can_movestopcondition_ignore_stalls = 0x20,
 } CANMoveStopCondition;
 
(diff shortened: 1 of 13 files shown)


nusrat813 (approved):
Tested, LGTM

PR #639, opened 2023-04-19
fix(motor-control): allow StopRequests outside estop

sfoster1:
When we implemented StopRequest, we were really doing it as a way to make sure everything got the message about an estop activation. 

We now need to use it to clean up after cancelling protocols, and send it from the host outside an estop situation... and that didn't really work.

This PR decouples the logic of
- estop handling
- canceling and clearing the interrupt handler move queue
- getting StopRequests

It's now safe to get an arbitrary stop request during a move or not during a move which will halt motion and clean up all the move state and clear out the queues. This also simplifies some of the interrupt handler state logic.

## Testing
- [x] on a loaded machine, send a stop request during a running move
- [x] on a loaded machine, hit the estop
- [x] on a loaded machine, send a move while estop is active
- [x] on a loaded machine, deactivate estop then send a move
- [x] on a loaded machine, send a stoprequest while estop is asserted

## Review requests
- logic look good

## Todo
- Tests

### DIFF
--- a/include/bootloader/core/ids.h
+++ b/include/bootloader/core/ids.h
@@ -142,6 +142,7 @@ typedef enum {
     can_errorcode_labware_dropped = 0x9,
     can_errorcode_estop_released = 0xa,
     can_errorcode_motor_busy = 0xb,
+    can_errorcode_stop_requested = 0xc,
 } CANErrorCode;
 
 /** Tool types detected on Head. */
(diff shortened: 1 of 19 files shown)


fsinapi (approved):
🆒 LGTM

ryanthecoder (approved):
Looks good to me

PR #640, opened 2023-04-21
fix(pipettes): correct sense resistor values

sfoster1:
It turns out that the sense resistors for all the motors on the 96
pipette are .15ohm instead of .1ohm and we never noticed. Yes, that
includes older revisions.

We're going to rework the sense resistor specifically on the plunger
back to 0.1Ohm, so we can keep that value, but that's to have a larger
possible current so we bump the global scaler there. Meanwhile, the tip
motors will keep their 0.15Ohm resistors, so we should get the right
number written down.


Laura-Danielle (approved):
lgtm

PR #641, opened 2023-04-24
refactor(pipettes): change pipette microstep to 16

caila-marashaj:
When the plunger motors on the 96 channel pipette would stall with a microstep value of 64, the firmware would shut down. Decreasing the resolution to 16 seems to fix that problem and allow the pipette to keep operating after a stall.


Laura-Danielle (approved):
lgtm

PR #642, opened 2023-04-24
fix(gripper): stop the gripper error loop during estop

ryanthecoder:


sfoster1 (approved):
Yeah, this looks good to me!

PR #643, opened 2023-04-24
fix(motor-control): make _hardware_delay a template function

ryanthecoder:
didn't catch this before because it didn't throw an error on CI or my system, but clang host builds were failing


ahiuchingau (approved):
Fabulousioso

PR #644, opened 2023-04-24
fix(include): stop continuous polls when sync and echo flags are cleared

Laura-Danielle:
## Overview

We never actually disabled continuous polls when the sync and echo flags were cleared. We should explicitly disable the poller when these flags have changed.

I'm not sure if this will fix some of the reliability issues we were seeing, but I guess we can give it a go tomorrow.


sfoster1 (approved):
Oh, nice!

PR #645, opened 2023-04-25
fix(pipettes): completely disable NVIC for data ready across all 96 channels

Laura-Danielle:
For some reason, on DVT 96 channels the hardware interrupt still seems to cause the firmware to crash. We should completely disable it on the 96 channel to unblock testing until switching the pressure sensor to a software timer is completed.


sfoster1 (approved):
image

PR #646, opened 2023-04-26
fix(include): do not reset CAPDAC after every request to stop polling

Laura-Danielle:
## Overview

I was overzealous in when we should reset the CAPDAC register. We should only reset the capdac register if we're switching sensors.

Also, I broke some tests previously 😬


fsinapi (approved):
Makes sense!

sfoster1 (approved):
yay

PR #648, opened 2023-04-27
refactor(pipettes): do not accept update position estimation messages for the gear motors

Laura-Danielle:
## Overview

We don't care about encoder info for the tip motors because they do not have encoders. Since there aren't two separate node ids on the 96 channel pipettes, it's just easier to disable the update position estimation request for gear motors.


sfoster1 (approved):
Yeah that sounds like the right play!

sfoster1:
Once tests pass anyway

PR #649, opened 2023-04-28
chore: Only build state manager wheel

DerekMaggio:
Need to only build the wheel because when it builds both I get dependency resolution errors when doing pip install /ot3-firmware/state-manager/dist/*

Better to just not build it at all instead of filtering it out in the install.


sfoster1 (approved):
makes sense, wheels are better anyway

PR #650, opened 2023-04-28
fix(pipettes): did not use the correct bus for the front cap sensor on 96 channel

Laura-Danielle:
## Overview

Title says it all.


sfoster1 (approved):
ope

PR #652, opened 2023-05-03
fix(gantry): Fix the gantry emulators crashing

ryanthecoder:
The way I added the the eeprom to the gantry simulators accidently called the backing creation twice which ended up making it point to the wrong file


sfoster1 (approved):
looks good to me

PR #653, opened 2023-05-03
feat(pipettes): add acceleration to TipActionRequest message

caila-marashaj:
This code accompanies the monorepo changes in [this pr](https://github.com/Opentrons/opentrons/pull/12944)- it adds acceleration to `TipActionRequest` messages, and uses the `position_flags` parameter in `GearMotorAck`s after a move is complete


Laura-Danielle (changes requested):
You need to edit:

`include/motor-control/core/stepper-motor/motion_controller.hpp::PipetteMotionController`

in order for the acceleration to actually get used (and I believe the pip motion controller task too).

Let me know if you can't find where these are.

Laura-Danielle (approved):
yep looks good, but please don't merge in till the python pr is ready to be merged in.

sfoster1 (changes requested):
_just in case_ so we don't drive ourselves crazy if we flash the wrong thing, can we put the acceleration part at the end of the message and set it to 0 if we don't have a long enough message to include it

sfoster1:
Looks good otherwise though!

Laura-Danielle (approved):

include/can/core/messages.hpp
```diff
@@ -198,6 +198,9 @@ using MotorPositionRequest = Empty<MessageId::motor_position_request>;
 using UpdateMotorPositionEstimationRequest =
     Empty<MessageId::update_motor_position_estimation_request>;
 
+using UpdateGearMotorPositionEstimationRequest =
```
ahiuchingau, line 201: I don't understand what we're using to update gear motor position estimation? I don't think we have encoders on the tip motors, right?

sfoster1 (approved):
Looks good to me! Really great work.

PR #654, opened 2023-05-03
fix(motor-control): handling underflow values

ahiuchingau:
This PR handles negative encoder value properly.
 
When the encoder counter underflows once, the overflow counter becomes -1. The calculation for the encoder pulses interpreted this we've gone through -65535 pulse, but we actually haven't.

### DIFF
--- a/motor-control/firmware/stepper_motor/motor_hardware.cpp
+++ b/motor-control/firmware/stepper_motor/motor_hardware.cpp
@@ -64,7 +64,8 @@ int32_t MotorHardware::get_encoder_pulses() {
         return 0;
     }
     if (motor_encoder_overflow_count < 0) {
-        return motor_encoder_overflow_count * motor_hardware_encoder_pulse_count(enc_handle);
+        return ((motor_encoder_overflow_count + 1) << 16) -
+               motor_hardware_encoder_pulse_count(enc_handle);
     }
     return (motor_encoder_overflow_count << 16) +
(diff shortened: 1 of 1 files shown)


Laura-Danielle (approved):
nioce find

fsinapi (approved):
🆒

sfoster1 (changes requested):
From my reading of the g4 manual 28.3.25, I think this PR causes incorrect underflow handling and the previous code was correct.

Take the following behavior:
- `motor_encoder_overflow_count == 0`, `TIMx->CNT == 0`
- one negative-direction pulse detected on the encoder
- `OVF` fires, and `motor_encoder_overflow_count = -1`
- and, _to my reading_, `TIMx->CNT` gets set from `TIMx->ARR`, and now `TIMx->CNT=65535`
- Then, the proper thing to do is `(-1<<16) + 65535`; this is -2^16 + 2^16-1, or -1. If instead you do `0<<16 - 65535` you're doing 0 - 2^16 - 1, you get -(2^16-1) and end up with `-65535`.
This change would be correct, though, if the behavior of the timer was to count back up from 0 when the encoder went negative;

sfoster1:
A thing that could definitely be happening is that we miss an overflow though, that might cause a problem. Or count an extra one.

ryanthecoder:
This can be closed, was fixed in another PR

PR #655, opened 2023-05-04
feat(rear-panel) add support for missing gpio pin states

ryanthecoder:
Adds messages and handlers to fetch the state of Aux1/2 det, aux1/2 ID and estop 1/2 det pins over USB


ryanthecoder:
Yea they're separate gpio pins for aux vs estop detect.

PR #656, opened 2023-05-04
refactor(include): convert pressure driver to use a software timer rather than hardware interrupts

Laura-Danielle:
## Overview

We cannot rely on the data ready interrupt to tell us when data is ready for the pressure sensor. Instead, we will now use a software timer to poll the data from the pressure sensor. Thankfully all of the infrastructure was there to easily make this change, however, I also took the opportunity to refactor the pressure driver a little bit to (hopefully) make it a little cleaner/more clear.

It was also requested from EEs that we switch over to using the low pass filter built in on the sensor, so I made that change here as well.

## Tests

- [x] Verify liquid sensing still functions as expected
- [x] Verify values coming back from the pressure sensor match what we expect.

### DIFF
--- a/include/sensors/core/tasks/pressure_driver.hpp
+++ b/include/sensors/core/tasks/pressure_driver.hpp
@@ -278,10 +278,9 @@ class MMR920C04 {
                                                   m.read_buffer.cend(),
                                                   temporary_data_store));
 
-        int32_t signed_data_store =
-            static_cast<int32_t>(temporary_data_store >> 8);
+        uint32_t shifted_data_store = temporary_data_store >> 8;
 
-        save_pressure(signed_data_store);
+        save_pressure(shifted_data_store);
         auto pressure = mmr920C04::PressureResult::to_pressure(
             _registers.pressure_result.reading);
         if (bind_sync) {
@@ -319,10 +318,9 @@ class MMR920C04 {
                                                   m.read_buffer.cend(),
                                                   temporary_data_store));
 
-        int32_t signed_data_store =
-            static_cast<int32_t>(temporary_data_store >> 8);
+        uint32_t shifted_data_store = temporary_data_store >> 8;
 
-        save_temperature(signed_data_store);
+        save_temperature(shifted_data_store);
 
         if (echoing) {
             auto temperature = mmr920C04::TemperatureResult::to_temperature(
@@ -344,11 +342,10 @@ class MMR920C04 {
                                                   m.read_buffer.cend(),
                                                   temporary_data_store));
 
-        int32_t signed_data_store =
-            static_cast<int32_t>(temporary_data_store >> 8);
+        uint32_t shifted_data_store = temporary_data_store >> 8;
 
         auto pressure =
-            mmr920C04::PressureResult::to_pressure(signed_data_store);
+            mmr920C04::PressureResult::to_pressure(shifted_data_store);
         pressure_running_total += pressure;
 
         if (!m.id.is_completed_poll) {
@@ -392,11 +389,10 @@ class MMR920C04 {
                                                   m.read_buffer.cend(),
                                                   temporary_data_store));
 
-        int32_t signed_data_store =
-            static_cast<int32_t>(temporary_data_store >> 8);
+        uint32_t shifted_data_store = temporary_data_store >> 8;
 
         auto temperature =
-            mmr920C04::TemperatureResult::to_temperature(signed_data_store);
+            mmr920C04::TemperatureResult::to_temperature(shifted_data_store);
         temperature_running_total += temperature;
 
         if (!m.id.is_completed_poll) {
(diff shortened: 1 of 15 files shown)


pipettes/firmware/utility_gpio.c
```diff
@@ -59,35 +51,24 @@ static void tip_sense_gpio_init() {
...
         /* EXTI interrupt init block tip sense*/
         HAL_NVIC_SetPriority(block_2, 10, 0);
         HAL_NVIC_EnableIRQ(block_2);
 
-        /* EXTI interrupt init data ready*/
-        HAL_NVIC_SetPriority(block_3, 10, 0);
-        HAL_NVIC_EnableIRQ(block_3);
-    }
-    if (pipette_type == EIGHT_CHANNEL) {
-        IRQn_Type block_1 = get_interrupt_line(gpio_block_1);
```
sfoster1, line 84: do we not want the 8 chan tip sense?

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -251,155 +253,228 @@ class MMR920C04 {
...
-    auto handle_baseline_poll(uint32_t message_index, float data) -> void {
-        running_total += data;
-        if (readings_taken == total_baseline_reads) {
-            offset_average = running_total / total_baseline_reads;
-            send_baseline_response(message_index);
-            set_number_of_reads(0);
+    auto handle_ongoing_response(i2c::messages::TransactionResponse &m)
+        -> void {
+        auto reg_id = utils::reg_from_id<mmr920C04::Registers>(m.id.token);
+        std::cout << "echoing" << echoing << "\n";
```
sfoster1, line 270: not to future, make these `LOG`

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -251,155 +253,228 @@ class MMR920C04 {
...
+        }
+        if (temperature_read) {
+            save_temperature(signed_data_store);
+        }
+        if (bind_sync && pressure_read) {
+            auto pressure = mmr920C04::PressureResult::to_pressure(
+                _registers.pressure_result.reading);
+            auto signed_pressure =
+                mmr920C04::PressureResult::to_pressure(signed_data_store);
+            std::cout << "I'm in bind sync and pressure is" << pressure << "\n";
```
sfoster1, line 302: note to future, make these `LOG`

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -251,155 +253,218 @@ class MMR920C04 {
...
+
+        // Pressure is always a three-byte value
+        static_cast<void>(bit_utils::bytes_to_int(m.read_buffer.cbegin(),
+                                                  m.read_buffer.cend(),
+                                                  temporary_data_store));
+
+        int32_t signed_data_store =
+            static_cast<int32_t>(temporary_data_store >> 8);
+
+        auto pressure_read =
```
sfoster1, line 284: these can't both be true, right? can we factor this out into a function for each?

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -251,155 +253,218 @@ class MMR920C04 {
...
-            case mmr920C04::Registers::MEASURE_MODE_3:
-            case mmr920C04::Registers::MEASURE_MODE_4:
-            case mmr920C04::Registers::MACRAM_WRITE:
-                break;
+        auto reg_id = utils::reg_from_id<mmr920C04::Registers>(m.id.token);
+
+        int32_t signed_data_store =
+            static_cast<int32_t>(temporary_data_store >> 8);
+
+        auto pressure_read =
```
sfoster1, line 344: same thing here - can we factor this out into a pressure handler and a temperature handler?

include/sensors/core/tasks/pressure_sensor_task.hpp
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <iostream>
```
sfoster1, line 3:
```suggestion
#include <iostream>
```

include/sensors/core/tasks/pressure_sensor_task.hpp
```diff
@@ -47,28 +49,38 @@ class PressureMessageHandler {
...
+        if ((reg_id != mmr920C04::Registers::LOW_PASS_PRESSURE_READ) &&
+            (reg_id != mmr920C04::Registers::PRESSURE_READ) &&
+            (reg_id != mmr920C04::Registers::TEMPERATURE_READ) &&
+            (reg_id != mmr920C04::Registers::STATUS)) {
+            return;
+        }
+        // may not be routed to baseline function like we suspect
+        if (utils::tag_in_token(m.id.token,
+                                utils::ResponseTag::POLL_IS_CONTINUOUS)) {
+            std::cout << "handling ongoing response\n";
```
sfoster1, line 63: needs to be `LOG`

include/sensors/core/tasks/pressure_sensor_task.hpp
```diff
@@ -47,28 +49,38 @@ class PressureMessageHandler {
...
+            driver.handle_ongoing_response(m);
+            LOG("continuous transaction response");
+        } else {
+            std::cout << "handling baseline response\n";
```
sfoster1, line 67: needs to be `LOG`

sfoster1 (changes requested):
Looks generally good but can we factor out these pressure/temperature response handlers? also some iostream stuff still in there

sfoster1 (approved):
Yeah, this looks good to me. Hype to have it all back!

PR #657, opened 2023-05-04
fix(pipettes): fix pipette simulators crashing

fsinapi:
Fixes a few issues causing I2C on the pipette simulators to crash

- Set the EEPROM chip type differently for 96 channel target
- Allocate EEPROM to I2C1 or I2C3 based on whether this is a 96 channel
- Only include the front capsensor on 96 channel (shared on 8 channel)
- Add a second pressure sensor, with different hardware based on which target this is

Tested by running an ot3 emulator and watching the pipette simulator not crash.


sfoster1 (approved):
Nice

Laura-Danielle (approved):
make sense. Thanks for fixing this @fsinapi :)

PR #659, opened 2023-05-08
fix(include): use correct id to stop continuous polling

Laura-Danielle:
I'm not sure how this worked previously for the cap sensor, but we were having issues with the pressure sensor because we were using the incorrect token to stop the poll.


sfoster1 (approved):
Hwhoops

PR #660, opened 2023-05-09
fix(include): use the correct sampling rate for the pressure sensor

Laura-Danielle:
## Overview

This is technically still not right because we're ignoring the decimal place values of the sensor and the frequency of mode 1/2 will be indistinguishable. I made a [ticket](https://opentrons.atlassian.net/browse/RET-1358) to support decimal values for delays, but I think it will be a bit of a pain.


sfoster1 (approved):
Okay. IMO we won't get timing that close with our soft timers anyway.

PR #661, opened 2023-05-10
fix(motor-control): Fix encoder underflow errors that happen directly after homes

ryanthecoder:
On several boards we noticed an issue where the encoder would be -2^16 off directly after a home and during the course of the investigation we found that the overflow interrupt is running multiple times in rapid succession when the encoder value is reset to 0. 
this solution turns it off when it finishes a home, and before any move makes sure that it is enabled again.

### DIFF
--- a/include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
+++ b/include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
@@ -242,6 +242,7 @@ class MotorInterruptHandler {
                 hardware.get_encoder_pulses();
             hardware.reset_step_tracker();
             hardware.reset_encoder_pulses();
+            hardware.disable_encoder();
             stall_checker.reset_itr_counts(0);
             hardware.position_flags.set_flag(
                 can::ids::MotorPositionFlags::stepper_position_ok);
--- a/include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
+++ b/include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
@@ -305,6 +305,7 @@ class MotorInterruptHandler {
 
     void update_move() {
         has_active_move = move_queue.try_read_isr(&buffered_move);
+        hardware.enable_encoder();
         if (set_direction_pin()) {
             hardware.positive_direction();
         } else {
--- a/include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
+++ b/include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
@@ -305,7 +305,9 @@ class MotorInterruptHandler {
 
     void update_move() {
         has_active_move = move_queue.try_read_isr(&buffered_move);
-        hardware.enable_encoder();
+        if (has_active_move) {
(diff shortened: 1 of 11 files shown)


Laura-Danielle (approved):
lgtm once CI passes!

sfoster1 (changes requested):
I don't agree with this because this happens too early. The encoder is enabled or disabled when moves enter the execution queue, which can be a move before they're actually executed. If you enqueued a home and then another immediate move (which we currently don't do but is in no way illegal), then the encoder would be reenabled during the home; same vice versa.

Also, it sounds like the problem is that when the encoder is reading 0, if we're holding in position then small vibrations can very frequently bounce the value back and forth. That's something that can happen whenever the timer is at 0... which can happen in more situations than after a home, right? It could happen any time we stop around 0.

I think there's one and a half things we can do that fix this.

First, I think there's no getting away from needing some additional logic to patch in overflow counter changes if the encoder seems to have moved too fast - something like if it's changed its value by 2^8 or more in between ticks of the motor interrupt.

Second, and this is sort of hacky, like this PR says we do know that there's one place where we're guaranteed to have the encoder read 0 and be vulnerable to this: when we've homed, where we set the encoder to be 0. So what if we don't do that? I know this is kind of gross, but if we add a fixed-value logical offset to encoder values in the firmware then we move away from the problematic area. When we home, initialize the timer counter to like 64 or 32 or something - some value that's pretty small but more than a full step of the motor, since the motor shouldn't vibrate more than a full step. Then, whenever we read the encoder value, once we're in signed-int land subtract that offset. That way we remove the guaranteed problem point.

I think we need both, but I think the offset thing will fix the immediate problem so we could just do that for now.

andySigler:
just adding a comment here to say that I hit this error during moves, not homes, like @sfoster1 mentioned

sfoster1:
> just adding a comment here to say that I hit this error during moves, not homes, like @sfoster1 mentioned

during general moves, or during moves back to 0?

Laura-Danielle:
> > just adding a comment here to say that I hit this error during moves, not homes, like @sfoster1 mentioned
> 
> during general moves, or during moves back to 0?

I've only seen this issue on moves back to zero or moves with zero duration.

ryanthecoder:
sorry one more comment, I don't think this is related to vibrations while holding as I previously thought but instead due to one interrupt process trying to modify another interrupt process's flags.

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
```diff
@@ -304,6 +305,7 @@ class MotorInterruptHandler {
 
     void update_move() {
         has_active_move = move_queue.try_read_isr(&buffered_move);
+        hardware.enable_encoder();
```
sfoster1, line 308:
```suggestion
        if (has_active_move) {
             hardware.enable_encoder();
        }
```
right?

sfoster1 (approved):
Okay, I think this makes sense to fix that specific interrupt delay issue. I think it would be correct to merge this as is but I _think_ it's even more correct to only re-enable the encoder when we get another move, right?

PR #662, opened 2023-05-12
fix(pipettes): use correct id for tip motor eeprom tracking

Laura-Danielle:
Found a small lil typo whilst trying to figure out the 96 channel crashing stuff


vegano1 (approved):
nice

PR #664, opened 2023-05-12
fix(pipettes): use GPIO pin names rather than EXTI pin names for the interrupt

Laura-Danielle:
## Overview

Well apparently EXTI_LINE_X != the GPIO pin bitfield and it was crashing the 96 channel firmware every time the IT was called. This problem wasn't showing up on the single/eight channel because the EXTI_LINE_2 value happened to be correct for GPIO_PIN_2.


ryanthecoder (approved):
LGTM

fsinapi (approved):
👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

PR #665, opened 2023-05-15
fix(motor-control): Don't drop labware

ryanthecoder:
Don't drop labware when a stop requested or estop detected while a labware is being held


sfoster1 (approved):
Ah, good call. Nice.

caila-marashaj (approved):

ahiuchingau (approved):
Nice

PR #666, opened 2023-05-15
feat(rear-panel): Enable rear panel eeprom

ryanthecoder:
Add support for the eeprom chip in the rear-panel board.

compiles in the eeprom and i2c libraries into the eeprom binary
Adds Read/Write requests and Read Response messages to the usb messaging protocol

While we don't currently use the eeprom for anything we need it implemented in order to run verification tests during production


.github/workflows/rear_panel.yaml
```diff
@@ -52,7 +52,7 @@ jobs:
           path: "./stm32-tools"
           key: ${{ runner.os }}-${{ hashFiles('**/cmake/*') }}-${{ secrets.CACHE_VERSION }}
       - name: "Configure"
-        run: cmake --preset=cross . -DCMAKE_BUILD_TYPE=debug
```
Laura-Danielle, line 55: not sure why you made this change, but the history of specifying the build type as debug is that sometimes the optimized code will suppress some actual errors in code.

Laura-Danielle (approved):
lgtm

sfoster1 (approved):
This looks good and I'm ok with doing the release-mode tests. Please add more to the PR description before merging.

.github/workflows/rear_panel.yaml
ryanthecoder: We changed all of the boards to debug because the cross compiled debug builds no longer fit in the ROM of the chips.

PR #667, opened 2023-05-16
refactor(include): reduce the amount the active mode command is sent to the pressure driver

Laura-Danielle:
## Overview

The data sheet doesn't appear to specify what happens if you set the active mode and reset the sensor in quick succession, but reducing the amount of active modes seems to reduce the frequency of the sensor borking.


sfoster1 (approved):
nice nice

PR #668, opened 2023-05-16
chore(gantry): adjust X and Y motor configurations

fsinapi:
New values for the chopconfig registers based on David Gomez's testing last week. These values should, at the very least, be no _worse_ than the current settings.

Tested on a couple ABR robots and things seem good so far - there's less audible noise and none of the axes have skipped steps.


Laura-Danielle (approved):
ninja chop!

sfoster1 (approved):
image

PR #669, opened 2023-05-19
fix(can): modify TipActionResponse message order to match the expected order on python-side

Laura-Danielle:
## Overview

Unfortunately the ordering of the tip action response got messed up somewhere along the way. This caused the ack id to never return anyting other than the default state of 1 for the tip/gear motors.


sfoster1 (approved):
aha!

PR #670, opened 2023-05-19
chore: make firmware images contain refname

sfoster1:
Request from hardware to make it easier to distribute images to the factory. Now we'll get names like gantry-x-c2-v13.hex.

- [x] images have the refname in them 
- [x] images still end in .hex
- [x] applications are unmodified


andySigler (approved):

fsinapi (approved):
yes

PR #671, opened 2023-05-22
feat(rear-panel): add pin state requests/responses for aux tests

ryanthecoder:
Add messages to request estop and sync pin states,
also add a response that sends the sync line pin state


sfoster1 (approved):
Looks good to me

PR #672, opened 2023-05-23
feat(include): add checks for exceeding max sensor readings for the pressure and cap sensor

Laura-Danielle:
## Overview

This is only really relevant to the pressure sensor, but I decided to add support for the cap sensor as well. We want to make sure that the sensor does not ever exceed its max output to ensure that the sensor itself doesn't break.

There will be a python-side PR to add the new sensor output binding as well as the context manager to allow for continuous sensor monitoring.

### DIFF
--- a/include/bootloader/core/ids.h
+++ b/include/bootloader/core/ids.h
@@ -181,6 +181,7 @@ typedef enum {
     can_sensoroutputbinding_none = 0x0,
     can_sensoroutputbinding_sync = 0x1,
     can_sensoroutputbinding_report = 0x2,
+    can_sensoroutputbinding_max_threshold_sync = 0x3,
 } CANSensorOutputBinding;
 
 /** How a sensor's threshold should be interpreted. */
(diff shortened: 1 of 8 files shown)


include/bootloader/core/ids.h
```diff
@@ -181,6 +181,7 @@ typedef enum {
     can_sensoroutputbinding_none = 0x0,
     can_sensoroutputbinding_sync = 0x1,
     can_sensoroutputbinding_report = 0x2,
+    can_sensoroutputbinding_max_threshold_sync = 0x3,
```
sfoster1, line 184: this is a bitfield I think - should be `0x4`

sfoster1 (changes requested):
I think we want a different signalling mechanism than the sync line for this, because it could cause a conflict between the different things we're checking. We could use another gpio, or send a message or something.

include/bootloader/core/ids.h
Laura-Danielle: ah you're right

sfoster1 (approved):
ok i understand what we're going for here, we good

Laura-Danielle:
For other interested parties:

- Seth was worried about the state appearing to be the same (i.e. motors stopping) if we were checking overpressure versus an application threshold.
- In a follow-up PR we will be adding an error message that is sent when overpressure is detected because this is an unrecoverable state.

This means that there will be a noticeable difference the two states

PR #673, opened 2023-05-25
feat(include): add error code for overpressure trigger

Laura-Danielle:
## Overview

We should throw an error message for the use-case of an overpressure spike being detected. Potentially we'll want to do some filtering at some point, but I think the chances of us getting into this value range is very low and thus it's OK to not do any type of signal processing.


sfoster1 (approved):
Yeah that seems pretty good!

PR #675, opened 2023-05-30
feat(rear-panel): add the revb2 in anticipation of the new boards

ryanthecoder:
Adds the new revision which enables:
heartbeat LED
PGOOD input pin

### DIFF
--- a/include/rear-panel/core/tasks.hpp
+++ b/include/rear-panel/core/tasks.hpp
@@ -12,6 +12,7 @@
 #include "rear-panel/core/messages.hpp"
 #include "rear-panel/core/queues.hpp"
 #include "rear-panel/core/tasks/hardware_task.hpp"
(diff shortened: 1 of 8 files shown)


ryanthecoder:
Marked as draft until boards come in to test

sfoster1 (changes requested):
Looks OK overall and I'm glad to get ahead but for the heartbeat LED to be useful it can't be timer driven, because then even if the actual processing the CPU does isn't working the timer will still be doing. We should toggle it either in the idle task or in a dedicated low priority freertos task so that we can use its flickering as an indication that the device is working properly.

sfoster1 (approved):
Looks good! Do we need to set the rev in the bootloader too?

ryanthecoder:
There is no version in this bootloader, we us DNF for flashing and the "bootloader" just does the backup image copy and jumps to the application.

PR #676, opened 2023-05-31
feat(motor-control): limit switch backoff move for stepper motors

ahiuchingau:
Ideally, the origin point of each axis should be the exact same point each time we command the axis to home. With level switches, there's a range of operating distance where the contacts are tripped. To improve precision, every time we home an axis and trigger a limit switch, we can slowly back away from the switch until the first instance the contacts are released. This way we can ensure the position of the axis origin to be roughly the same every time. 

Main changes:
1. Add a new move stop condition `MoveStopCondition::limit_switch_backoff`
2. Instead of resetting the stepper and encoder tracker after a HomeRequest, we're going to reset everything after the subsequent `limit_switch_backoff` move instead

Changes to the home & backoff behavior from last review:
1. Still setting stepper position to a large positive number at the beginning of a home_request (otherwise we'd run into negative position which we don't support)
2. Clearing stepper position flag when we begin to home
3. Resetting stepper position to 0 but not reset the status flag when limit switch is triggered
4. Only when backoff is completed we are resetting everything: stepper position, encoder position, motor status flags 

Required monorepo changes: https://github.com/Opentrons/opentrons/pull/12824

### DIFF
--- a/include/bootloader/core/ids.h
+++ b/include/bootloader/core/ids.h
@@ -218,6 +218,7 @@ typedef enum {
     can_movestopcondition_gripper_force = 0x8,
     can_movestopcondition_stall = 0x10,
     can_movestopcondition_ignore_stalls = 0x20,
+    can_movestopcondition_limit_switch_backoff = 0x30,
 } CANMoveStopCondition;
 
 /** A bit field of the arbitration id parts. */
--- a/include/bootloader/core/ids.h
+++ b/include/bootloader/core/ids.h
@@ -218,7 +218,7 @@ typedef enum {
     can_movestopcondition_gripper_force = 0x8,
     can_movestopcondition_stall = 0x10,
     can_movestopcondition_ignore_stalls = 0x20,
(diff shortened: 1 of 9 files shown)


sfoster1 (approved):
Nice, big fan! Always forget you can do `or` in c++

include/bootloader/core/ids.h
```diff
@@ -216,6 +216,7 @@ typedef enum {
     can_movestopcondition_gripper_force = 0x8,
     can_movestopcondition_stall = 0x10,
     can_movestopcondition_ignore_stalls = 0x20,
+    can_movestopcondition_limit_switch_backoff = 0x30,
```
fsinapi, line 219: Can we change this to `0x40` so the enum can be treated as a bitmask?

ahiuchingau: yes yes it should all be 0x40, forgot to change this one

sfoster1 (approved):
Makes sense to me, nice!

PR #677, opened 2023-06-01
fix(pipettes): remove encoder handle from gear motor hardware

ryanthecoder:
remove the encoder handle from the gear motor hardware interfaces. they were resetting the plunger encoder when they did tip actions


ahiuchingau (approved):
Good 👁️

PR #678, opened 2023-06-02
refactor(motor-control): use um/tick/tick for acceleration to increase resolution

ahiuchingau:
This PR changes the acceleration params we use in linear move request to use `um/tick/tick` instead of `mm/tick/tick` so we can work with lower accelerations.

### DIFF
--- a/include/can/core/messages.hpp
+++ b/include/can/core/messages.hpp
@@ -20,7 +20,7 @@ using namespace can::ids;
 
 using stepper_timer_ticks = uint32_t;
 using brushed_timer_ticks = uint32_t;
-using um_per_tick = int32_t;
+using mm_per_tick = int32_t;
 using um_per_tick_sq = int32_t;
 
 /**
@@ -306,7 +306,7 @@ struct AddLinearMoveRequest : BaseMessage<MessageId::add_move_request> {
     uint8_t seq_id;
     stepper_timer_ticks duration;
     um_per_tick_sq acceleration;
-    um_per_tick velocity;
+    mm_per_tick velocity;
     uint8_t request_stop_condition;
 
     template <bit_utils::ByteIterator Input, typename Limit>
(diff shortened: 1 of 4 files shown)


sfoster1 (approved):
Love it

include/can/core/messages.hpp
```diff
@@ -20,7 +20,7 @@ using namespace can::ids;
 
 using stepper_timer_ticks = uint32_t;
 using brushed_timer_ticks = uint32_t;
-using um_per_tick = int32_t;
+using mm_per_tick = int32_t;
```
Laura-Danielle, line 23: wait shouldn't this stay as um_per_tick?

include/can/core/messages.hpp
```diff
@@ -536,8 +536,8 @@ using GetMotionConstraintsRequest =
 struct GetMotionConstraintsResponse
     : BaseMessage<MessageId::get_motion_constraints_response> {
     uint32_t message_index;
-    um_per_tick min_velocity;
-    um_per_tick max_velocity;
+    mm_per_tick min_velocity;
```
Laura-Danielle, line 539: Are we changing the resolution to mm_per_tick or um_per_tick?

Laura-Danielle (commented):
Had some questions about resolution.

include/can/core/messages.hpp
ahiuchingau: velocity should be mm_per_tick

include/can/core/messages.hpp
ahiuchingau: Yeah so we've been mistakenly using `um_per_tick` for velocity for a long time (not in calculation, just for naming the units), so this clears that/

So `mm_per_tick` for velocity and `um_per_tick_sq` for acceleration

PR #679, opened 2023-06-02
chore(motor-control): put motor-control functions in ccmram

fsinapi:
The motor interrupts are kind of slow. This PR is an attempt to speed them up, especially for the head controller.

* We add a section in the linker script to store function in the STM32 CCMRAM. 
  * This section is configured to grab every function in the `motor_handler` and `motor_hardware` namespaces
* We manually mark a few select functions to go in ccmram along with the specific motor controller stuff
* We add a section in the startup file to copy the ccmram functions from flash to ccmram, because the ccmram is volatile
* We also cut out the STM32 timer interrupt handler for the Head unit's motor control interrupt because it was taking up a lot of time with no real benefit

This seems to speed up the motors enough to let the head move both axes at the same time without crashing, which it couldn't do before. In the absence of any quantitative measurement of the interrupt latency, that feels like enough justification to make this change.

### DIFF
--- a/common/firmware/STM32G491RETx/STM32G491RETx_FLASH.ld
+++ b/common/firmware/STM32G491RETx/STM32G491RETx_FLASH.ld
@@ -73,6 +73,16 @@ SECTIONS
     _sccmram = .;       /* create a global symbol at ccmram start */
     *(.ccmram)
     *(.ccmram*)
+
+    /* NOTE ABOUT THE FOLLOWING SECTIONS
+    * 
+    * The CMAKE toolchain for cross-compilation MUST be configured
+    * with the option `-ffunction-sections` in order for this to work.
+    * This section matches the sections for the namespaces for motor
+    * functionality, and those namesapces are only broken out into
+    * individual linker sections if the function-sections option is enabled.
+    */
+
     *(.text.*motor_handler*) 
     *(.text.*motor_hardware*) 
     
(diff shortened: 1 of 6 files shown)


common/firmware/STM32G491RETx/STM32G491RETx_FLASH.ld
```diff
@@ -59,6 +59,27 @@ SECTIONS
...
+  * If initialized variables will be placed in this section,
+  * the startup code needs to be modified to copy the init-values.  
+  */
+  .ccmram :
+  {
+    . = ALIGN(4);
+    _sccmram = .;       /* create a global symbol at ccmram start */
+    *(.ccmram)
+    *(.ccmram*)
+    *(.text.*motor_handler*)
```
sfoster1, line 76: I think this relies on `-ffunction-sections` being on so let's drop a note wherever we're enabling this in cmake

sfoster1 (approved):
Looks good but also maybe make a note wherever we're enabling `-ffunction-sections` that it now controls ccmram usage

common/firmware/STM32G491RETx/STM32G491RETx_FLASH.ld
fsinapi: That's set in the `cmake` subtree, can either add it in a PR there or drop something in the linker file

Laura-Danielle (approved):
This looks great! Can we also make this change for the pipettes once we've sufficiently tested on the head board?

sfoster1:
> Can we also make this change for the pipettes once we've sufficiently tested on the head board?

This will apply to the pipettes automatically - it'll happen for everythign that uses the g491 linker script

sfoster1:
it probably won't happen for the tip motors on the 96 though but maybe that's fine

PR #681, opened 2023-06-06
fix(eeprom): replace all vectors with arrays

ryanthecoder:
Does not fix the crash but does get rid of vectors


Laura-Danielle (changes requested):
I think you need to handle the use-case of differently sized table_updater global objects.

sfoster1 (changes requested):
i think if we flip these so we define the `table_updater` elements before the messages, and then use`table_updater.size()` for the len we should be good

PR #682, opened 2023-06-07
fix(pipettes): don't use more RAM than allowed

fsinapi:
The linker file provided by STM is correct to say that the microcontroller has 112K RAM, but it _also_ separately defines the last 16K of that as the `CCMRAM` region. This means that, if the entire RAM is used by the program, variables will overlap into anything defined in ccmram and chaos will ensue.

The pipette firmware was all overflowing into ccmram, which the linker didn't complain about because it is unaware that the ccmram is aliased at the end of the normal SRAM address region.

The pipettes _also_ have a lot of tasks, and each of those tasks has a minimum stack size of 512 words (so 2048 bytes). This PR cuts all of the stack allocations in half.

Tested on a multichannel and a 96-channel. Able to calibrate, aspirate, dispense, pick up tips... seems good to go.

### DIFF
--- a/pipettes/core/peripheral_tasks.cpp
+++ b/pipettes/core/peripheral_tasks.cpp
@@ -11,17 +11,17 @@ static auto i2c3_task_client =
     i2c::writer::Writer<freertos_message_queue::FreeRTOSMessageQueue>();
 
 static auto i2c1_task_builder =
-    freertos_task::TaskStarter<512, i2c::tasks::I2CTask>{};
+    freertos_task::TaskStarter<256, i2c::tasks::I2CTask>{};
 static auto i2c3_task_builder =
-    freertos_task::TaskStarter<512, i2c::tasks::I2CTask>{};
+    freertos_task::TaskStarter<256, i2c::tasks::I2CTask>{};
 template <template <typename> typename QueueImpl>
 using PollerWithTimer =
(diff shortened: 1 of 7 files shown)


pipettes/core/peripheral_tasks.cpp
```diff
@@ -11,17 +11,17 @@ static auto i2c3_task_client =
...
 static auto i2c3_task_builder =
-    freertos_task::TaskStarter<512, i2c::tasks::I2CTask>{};
+    freertos_task::TaskStarter<256, i2c::tasks::I2CTask>{};
 template <template <typename> typename QueueImpl>
 using PollerWithTimer =
     i2c::tasks::I2CPollerTask<QueueImpl, freertos_timer::FreeRTOSTimer>;
 
 static auto i2c1_poll_task_builder =
-    freertos_task::TaskStarter<1024, PollerWithTimer>{};
+    freertos_task::TaskStarter<512, PollerWithTimer>{};
```
Laura-Danielle, line 22: hmm I think this was increased for a specific reason, but I can't remember why. We'll just need to do a lot of testing with the sensors

Laura-Danielle (approved):
Overall this lgtm.

We should do extensive testing with the sensors and motor movement.

pipettes/core/peripheral_tasks.cpp
fsinapi: I'm happy to preemptively bump this one back up, I just indiscriminately cut everything in half.

PR #684, opened 2023-06-07
fix(host): comment out .ccmram section for host builds

fsinapi:
If your host compiler is GCC, the .ccmram section attributes won't do anything because GCC silently ignores missing sections (whether that's good or bad is up for debate). If your host compiler is _not_ GCC, you might get compiler errors. This PR should eliminate those errors.


sfoster1 (approved):
yeah, that works for me!

PR #686, opened 2023-06-08
fix(pipettes): simulator environment var did not match command line variable

fsinapi:
`MOUNT` in the environment variables needed to be lowercase `mount` to map correctly to the command line variables.

Before this PR, the simulator would crash if you defined an environment variable `MOUNT`. Now it will work correctly + pick up the option.


sfoster1 (approved):
Whoops, nice find

PR #687, opened 2023-06-08
feat(motor-control): report stepper & encoder position when stall occurs

ahiuchingau:
Send a MoveComplete message, on top of the error message, to report current stepper & encoder position when a stall happens during an active move.


sfoster1 (approved):
Awesome, definitely meant to get to this at some point

PR #688, opened 2023-06-08
fix(head): enable the brake pins on boot

ryanthecoder:
96 channel drops on boot if brakes aren't on


sfoster1 (approved):
Looks good to me

PR #689, opened 2023-06-20
feat(include): Do not adjust capdac on initial startup readings

Laura-Danielle:
## Overview

Closes RLIQ-409, RLIQ-402 and RLIQ-305. Using a SMA windowed filter, do not set the capdac on readings until the readings themselves have stabilized.

I still kept the actual 'start-up' capdac which sets the value to 0 because I figured this was harmless. Let me know if this is not the case.

## Review Requests

Please check me to make sure that everything looks acceptable!


cpp-utils/include/ot_utils/core/filters/ema.hpp
```diff
@@ -0,0 +1,38 @@
...
+                prior_average, fixed_point_prior_weight, Radix);
+
+        IntegerValue filtered_data = current_weighted + prior_weighted;
+        prior_average = filtered_data;
+        return filtered_data;
+    }
+
+  private:
+    IntegerValue prior_average = 0;
+    float current_weight = static_cast<float>(1.0 / AlphaDivisor);
```
sfoster1, line 29: Let's make these `constexpr` functions or values since they're known by the template

cpp-utils/include/ot_utils/core/filters/sma.hpp
```diff
@@ -0,0 +1,68 @@
...
+// A finite impulse response (FIR) or 'windowed' filter
+
+// Use this filter when you have a finite window
+// of data that is particularly noisy.
+namespace ot_utils {
+
+namespace filters {
+
+template <class IntegerValue, int WindowSize = 10, int Radix = 8>
+class SimpleMovingAverage {
```
sfoster1, line 17: let's add a `ready() -> bool` method that a caller can use to check if values are ready. or make `compute` return an optional or something. right now, you can't really tell when the average isn't actually running without doing part of the average's job (deciding what the window is and how it works) and since this is going to be used to reduce noise, you're expecting it to, well, have less noise. But for the first 10 samples it's just exactly the input so you'd make bad decisions.

sfoster1 (changes requested):
Inline comments. Plus,
- Let's try and cut down on the explicit specialization of the fixed point stuff; I think we can do a pretty good job of that without doing full struct definitions for fixed point types
- you know `cpp-utils` is a subtree right? there's fun in your future

cpp-utils/include/ot_utils/core/filters/ema.hpp
```diff
@@ -0,0 +1,38 @@
+#pragma once
+
+#include "ot_utils/core/fixed_point.hpp"
+
+// Exponential Moving Average Algorithm
+// A infinite impulse response (IIR) or non-windowed
```
sfoster1, line 6: seems like an incomplete sentence?

cpp-utils/include/ot_utils/core/filters/ema.hpp
```diff
@@ -0,0 +1,38 @@
...
+                prior_average, fixed_point_prior_weight, Radix);
+
+        IntegerValue filtered_data = current_weighted + prior_weighted;
+        prior_average = filtered_data;
+        return filtered_data;
+    }
+
+  private:
+    IntegerValue prior_average = 0;
+    const float current_weight = static_cast<float>(1.0 / AlphaDivisor);
```
sfoster1, line 29: make these `constexpr` since they depend on template args

sfoster1 (approved):
Looks good to me if it works - will find someone to test

andy-g-calderon:
> Looks good to me if it works - will find someone to test

I was able to test and can confirm it resolved the issue that we were seeing.

sfoster1 (approved):
Yeah this looks good to me.

include/sensors/core/tasks/capacitive_driver.hpp
```diff
@@ -237,15 +240,24 @@ class FDC1004 {
             stop_continuous_polling(m.id.token);
             return;
         }
+
+        auto raw_capacitance = fdc1004_utils::convert_reads(polling_results[0],
+                                                            polling_results[1]);
+
+        if (data_unstable[int(sensor_id)]) {
+            raw_capacitance = filter.compute(raw_capacitance);
+            data_unstable.set(int(sensor_id), filter.stop_filter());
```
fsinapi, line 249: Is the logic correct here? My understanding is that `stop_filter` returns `true` if the capdac should stop filtering, aka if the data is stable. That means that, if the data is _not_ stable, `stop_filter` should return false. That will set the bit in `data_unstable` to 0, indicating that the data is stable.

fsinapi (changes requested):
Looks really nice but I had a question on the logic for `data_unstable`.

include/sensors/core/tasks/capacitive_driver.hpp
Laura-Danielle: Good catch!

fsinapi (approved):
Nice, LGTM!

PR #690, opened 2023-06-20
feat(gripper): Use ccram for gripper

ryanthecoder:
So before putting the gripper interrupts in ccram we were getting encoder overflow timing errors on the gripper jaw pretty frequently (~10 min of constant use) after doing the ccram stuff we drastically reduced the errors but it was still happening on avg every ~1.5 hours of constant use. So this throws the interrupts into ccram, and it also drops the priority of the motor usage timer to prevent it squashing over the interrupts.

As a final kind of guard against the encoder being -0xFFFF occasionally I threw in a check that the encoder is >=0 before reporting a stall

### DIFF
--- a/gripper/firmware/motor_hardware_shared.c
+++ b/gripper/firmware/motor_hardware_shared.c
@@ -177,6 +177,7 @@ void HAL_TIM_Encoder_MspDeInit(TIM_HandleTypeDef *htim) {
     }
 }
 
+__attribute__((section(".ccmram")))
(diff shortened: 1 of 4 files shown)


fsinapi (approved):
Nice, LGTM. If you wanna get really crazy we can probably throw `call_motor_handler` and `call_enc_handler` in ccmram too.

sfoster1 (approved):
Love it

PR #691, opened 2023-06-21
fix(include): pressure-sensor max-value

andySigler:
This PR sets the pressure sensor's max value to 8.2kPa, which is the highest value possibly returned by the pressure sensor. Anything higher cannot be measured and will return as 8.2kPa.


Laura-Danielle (approved):
Nice! Is this your first firmware PR for this repo??

PR #692, opened 2023-06-21
feat(gripper): add ccram header to the rest of the interrupts in the gripper

ryanthecoder:
Tested this overnight and it's working well so I think we should do this to further improve the gripper performance.


fsinapi (approved):
nice

sfoster1 (approved):
sweet sweet

PR #693, opened 2023-06-21
fix(pipettes): Convert back to EXTI for single and multi

Laura-Danielle:
## Overview

So, very confused and not sure if we also need to make this change for the 96 channel, BUT switching back only the clear flag to the EXTI value seems to fix the async notifications for tip sensing.

### DIFF
--- a/pipettes/firmware/stm32g4xx_it.c
+++ b/pipettes/firmware/stm32g4xx_it.c
@@ -119,14 +119,11 @@ void DebugMon_Handler(void) {}
 /******************************************************************************/
 
 void EXTI2_IRQHandler(void) {
-    __HAL_GPIO_EXTI_CLEAR_FLAG(EXTI_LINE_2);
     HAL_GPIO_EXTI_IRQHandler(GPIO_PIN_2);
 }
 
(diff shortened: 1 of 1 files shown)


fsinapi (commented):
If you look at `stm32g4xx_hal_gpio.c`, `HAL_GPIO_EXTI_IRQHandler` does its _own_ check on the flag (and will clear it) before it calls `HAL_GPIO_EXTI_Callback`. So I think the actual issue here is that we should either
1) call `HAL_GPIO_EXTI_Callback` directly from the IRQ instead of calling the IRQHandler that the HAL defines
2) not clear the flag ourselves, and instead just have the IRQ directly call the HAL's IRQHandler

Right now in `main`, the interrupt clears the flag and then calls the HAL handler that sees that the flag isn't set and therefore doesn't call our callback.

fsinapi (approved):
LGTM! I pushed this to a P1000 and toggled the sensor by hand while running  `journalctl -f | grep TipPresence` and saw a message every time.

sfoster1 (approved):
Ah this makes sense!

PR #694, opened 2023-06-26
fix(pipettes): increase sensor stack sizes

fsinapi:
I got a little too trigger happy with #682 with the sensor tasks. Some pipettes have been crashing when using the environmental or pressure sensors, and running one with a debugger + overflow detection enabled showed that the stack for the sensor task will overflow sometimes. The sensor tasks use their stack to maintain a parallel map of the register map of their respective sensor, so it kind of makes sense. We could (and maybe should) refactor the handlers to be statically allocated somehow to avoid throwing too much data on the stack, but for now we can just bump the stacks back up.


Laura-Danielle (approved):
makes sense.

PR #695, opened 2023-06-27
fix(motor-control): do not disable encoder after homing

fsinapi:
We really don't want the encoder to be disabled at any point because we want the motor controllers to be able to track any movements that happen while they're disabled. The encoder was originally disabled in order to deal with underflows after homing - this is seemingly due to the encoder being reset to 0 while there is some residual vibration on the motor shaft, resulting in the encoder ticking in the negative direction and underflowing the timer.

This PR changes the motor controllers to reset the encoder to a nonzero value, and does not disable the encoder after home movements.


sfoster1 (approved):
Yeah this is the right thing

ryanthecoder (approved):
agreed

vegano1 (approved):
lgtm

ahiuchingau (approved):
love it

PR #696, opened 2023-06-29
feat(i2c): interrupt-based I2C communication

fsinapi:
This PR changes the firmware's use of the HAL I2C API to be interrupt-based. The i2c task kicks off a read/write transmission and then releases the CPU until the I2C callback is called. This should release the CPU to run other FreeRTOS tasks during I2C transmissions, instead of spinning for the duration of the transmission. 

- When starting a new read/write, a handle associated with the HAL struct is loaded with a task-to-notify
- When the Tx/Rx complete callback is called, the task-to-notify is looked up via the HAL struct and, if it exists, that task is unblocked

This is the same implementation used on the Heater-Shaker and Thermocycler Gen2. 

Tested on FLEXY MCFLEXFACE by uploading firmware to the whole robot with a gripper, a 96 channel, and (after removing the 96) a single channel. Was able to read the serial number of all of the instruments & also calibrate them, and was able to dump the EEPROM from the other motor controllers without issues.


ahiuchingau (approved):
Sweeeeeeeeet

sfoster1 (approved):
This looks good to me, yeah. I wonder if it will make all the pipettes respond faster on attach

PR #698, opened 2023-07-06
fix(i2c): handle NACK condition properly

fsinapi:
When #696 switched from blocking to interrupt-based I2C, it did not change the handling of Nack conditions. Turns out that the old handling didn't work with the interrupt-based HAL functions, and instead we should catch those conditions based on whether the Complete Callback or the Error Callback gets invoked.

If the Error Callback gets called, simply retry the transmission.

Tested with a P50 Single Channel pipette. Before this PR, the pressure sensor always returned a pressure of -0.001. After the PR, it returns actual pressure data.

### DIFF
--- a/i2c/firmware/i2c.c
+++ b/i2c/firmware/i2c.c
@@ -11,6 +11,7 @@
 typedef struct {
     I2C_HandleTypeDef *i2c_handle;
     TaskHandle_t task_to_notify;
+    bool should_retry;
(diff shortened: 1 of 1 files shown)


andySigler (approved):
Tested on DVT single channel that previously was not working

sfoster1 (approved):
Okay, looks good!

PR #699, opened 2023-07-06
fix(motor-control): debounce stall detection

ryanthecoder:
Sometimes the motor interrupt and the encoder interrupt squash on each other for 1 cycle so this should fix that.


sfoster1:
@ryanthecoder I'm pretty sure #700 fixes this right?

PR #700, opened 2023-07-07
fix(motor-control): atomically account for encoder overflows

fsinapi:
## Overview

When one of the motor axes overflows its encoder, there's a not-uncommon race condition between the encoder timer's raw counter value and the variable capturing the number of timer overflows. If the motor interrupt controller tries to read the position between when the counter is updated but the overflows aren't updated, the encoder position will appear to be off by a considerable amount.

This PR enables the `UIFREMAP` bit for all of the encoders on the motor control boards. This bit effectively results in the MSb of the CNT register containing a flag for whether the encoder has a pending overflow. Instead of using the interrupt vector to determine when an overflow has occurred, we just check the UIFCPY bit in the CNT register. If an overflow has happened, the function getting the encoder pulses will clear the flag and return the direction in which the motor overflowed.

Because the encoder interrupts are disabled now, we have a new background timer that will periodically poll the encoder position. This is necessary because, if the position overflows twice while nothing is checking it, we would have no way to know how many times it overflowed.

## Testing

Tested by cycling the Z_L axis up and down such that the velocity reaches its peak value (this is important to overload the CPU with GPIO pulses) while it crosses the encoder overflow point. Without this change, there was a MTTF of maybe a dozen cycles. After the change I can run for 200 cycles with no issues.

Also disabled the gantry axes and dragged them across the deck and back, checking the encoder position from the REPL at a few points. The encoders were updating and the positions were consistent from edge to edge.

### DIFF
--- a/motor-control/firmware/brushed_motor/brushed_motor_hardware.cpp
+++ b/motor-control/firmware/brushed_motor/brushed_motor_hardware.cpp
@@ -62,10 +62,7 @@ int32_t BrushedMotorHardware::get_encoder_pulses() {
     if (!enc_handle) {
         return 0;
     }
-    int8_t overflows = 0;
-    uint32_t pulses = motor_hardware_encoder_pulse_count_with_overflow(
-        enc_handle, &overflows);
-    motor_encoder_overflow_count += overflows;
+    uint32_t pulses = motor_hardware_encoder_pulse_count(enc_handle);
     return (motor_encoder_overflow_count << 16) + (0xFF & pulses);
 }
 
--- a/motor-control/firmware/brushed_motor/brushed_motor_hardware.cpp
+++ b/motor-control/firmware/brushed_motor/brushed_motor_hardware.cpp
@@ -62,8 +62,11 @@ int32_t BrushedMotorHardware::get_encoder_pulses() {
     if (!enc_handle) {
         return 0;
     }
-    return (motor_encoder_overflow_count << 16) +
-           (0xFF & motor_hardware_encoder_pulse_count(enc_handle));
+    int8_t overflows = 0;
+    uint32_t pulses = motor_hardware_encoder_pulse_count_with_overflow(
+        enc_handle, &overflows);
+    motor_encoder_overflow_count += overflows;
+    return (motor_encoder_overflow_count << 16) + (0xFF & pulses);
 }
 
 uint16_t BrushedMotorHardware::get_stopwatch_pulses(bool clear) {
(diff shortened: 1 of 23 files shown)


motor-control/firmware/brushed_motor/brushed_motor_hardware.cpp
```diff
@@ -62,8 +62,8 @@ int32_t BrushedMotorHardware::get_encoder_pulses() {
     if (!enc_handle) {
         return 0;
     }
-    return (motor_encoder_overflow_count << 16) +
-           motor_hardware_encoder_pulse_count(enc_handle);
+    uint32_t pulses = motor_hardware_encoder_pulse_count(enc_handle);
+    return (motor_encoder_overflow_count << 16) + (0xFF & pulses);
```
sfoster1, line 66:
```suggestion
    return (motor_encoder_overflow_count << 16) + (0xFFFF & pulses);
```

motor-control/firmware/motor_control_hardware.c
```diff
@@ -94,15 +94,35 @@ bool motor_hardware_stop_pwm(void* htim, uint32_t channel) {
...
+}
+
+bool motor_hardware_encoder_is_counting_down(void *encoder_handle) {
+    if(encoder_handle == NULL) {
+        return true;
+    }
+    return __HAL_TIM_IS_TIM_COUNTING_DOWN((TIM_HandleTypeDef*)encoder_handle);
+}
+
+uint32_t motor_hardware_encoder_pulse_count_with_overflow(void* encoder_handle, int8_t *overflows) {
```
sfoster1, line 115:
```suggestion
uint16_t motor_hardware_encoder_pulse_count_with_overflow(void* encoder_handle, int8_t *overflows) {
```
Let's make this the same return type as the other one just for cleanliness

sfoster1 (changes requested):
Looks good overall and awesome that it fixes it, I think we need to widen the mask in the brushed DC handler though

motor-control/firmware/brushed_motor/brushed_motor_hardware.cpp
fsinapi: af28af4da7c0ae0df97e4aef333c752f79f4abf4

motor-control/firmware/motor_control_hardware.c
fsinapi: af28af4da7c0ae0df97e4aef333c752f79f4abf4

sfoster1 (approved):
Looks good to me when tests pass!

PR #701, opened 2023-07-10
chore: individual artifact files

sfoster1:
HW requested that releases get hex files for images uploaded directly so they can be deeplinked from PLM software.


sfoster1:
Example release: https://github.com/Opentrons/ot3-firmware/releases/tag/test-tag-8

fsinapi (approved):
Nice LGTM

PR #702, opened 2023-07-11
fix(motor-control): clear encoder status flag when home fails

ahiuchingau:
If the limit switch is not triggered within a `HomeRequest`, this means that we can no longer trust the encoder value and we should just clear the flag. This will force the axis to do a proper home when it receives a `HomeRequest` again.

### DIFF
--- a/include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
+++ b/include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
@@ -283,6 +283,10 @@ class MotorInterruptHandler {
         if (limit_switch_triggered()) {
             position_tracker = 0;
             hardware.reset_step_tracker();
+            if (stall_checker.has_encoder()) {
+                hardware.position_flags.set_flag(
+                    can::ids::MotorPositionFlags::encoder_position_ok);
+            }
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
Excellent, think this will solve that issue!

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
```diff
@@ -283,6 +283,10 @@ class MotorInterruptHandler {
         if (limit_switch_triggered()) {
             position_tracker = 0;
             hardware.reset_step_tracker();
+            if (stall_checker.has_encoder()) {
```
fsinapi, line 286: I don't think we should be setting the encoder flag here, and we should also probably get rid of the `hardware.reset_step_tracker()` line above it. Since we define the end of our homing sequence to be when the motors are backed off of the switches, we should only reset all of the trackers after backing off of the switch.

(If this is here because the hardware-controller expects the encoder flag to be asserted then we should change in that in Python imo)

fsinapi (commented):
I think we need to keep all of the position-flag stuff contained to when the switch-backoff step finishes, but otherwise this LGTM.

fsinapi (approved):
💯

PR #703, opened 2023-07-13
fix(sensors): mitigate cap sensor configuration issues

fsinapi:
The capacitive sensor seems to be having issues that stem from the initial register configuration. One issue seems to be that we are enabling a single channel at a time, but doing it twice in a row on startup. We may also be losing I2C messages in the flurry of bus traffic on startup, which may be exacerbated by the I2C tasks now releasing the CPU during transactions and therefore allowing other tasks to try to write more messages.

This PR does a couple of things which, in conjunction, seem to fix the issue of bad capacitive sensor data:
- Instead of directly polling the data registers, check the Reading Ready flags in the FDC register first.
- Always enable readings from both Reading 1 and Reading 2, and let them just run continuously forever
- Only configure one channel's readings at a time
- Add some `vTaskDelay` calls during startup to both 1) avoid I2C queue overflow 2) maybe give the sensor enough time to initialize?
- Removed the SMA filter, which had a few problems with fixed point conversions and also seems unnecessary with these fixes

Todo:

- [x] Test on single channel
- [x] Test on 96 channel
- [x] Test on gripper
- [x] Fix the capacitive sensor tests

### DIFF
--- a/include/sensors/core/tasks/capacitive_driver.hpp
+++ b/include/sensors/core/tasks/capacitive_driver.hpp
@@ -8,10 +8,12 @@
 // the vTaskDelay function and hopefully sometime in the near future I
 // can refactor this file with a nice templated sleep function.
 #include "FreeRTOS.h"
+// NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
 #define HACKY_TASK_SLEEP(___timeout___) vTaskDelay(___timeout___)
 
 #else
 
+// NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
 #define HACKY_TASK_SLEEP(___timeout___) (void)(___timeout___)
 #endif
 
@@ -204,7 +206,7 @@ class FDC1004 {
     }
 
     void handle_fdc_response(i2c::messages::TransactionResponse &m) {
-        uint16_t reg_int;
+        uint16_t reg_int = 0;
         static_cast<void>(bit_utils::bytes_to_int(
             m.read_buffer.cbegin(), m.read_buffer.cend(), reg_int));
         auto fdc = read_register<fdc1004::FDCConf>(reg_int).value();
--- a/include/sensors/core/tasks/capacitive_driver.hpp
+++ b/include/sensors/core/tasks/capacitive_driver.hpp
@@ -8,6 +8,11 @@
 // the vTaskDelay function and hopefully sometime in the near future I
 // can refactor this file with a nice templated sleep function.
 #include "FreeRTOS.h"
+#define HACKY_TASK_SLEEP(___timeout___) vTaskDelay(___timeout___)
+
+#else
+
+#define HACKY_TASK_SLEEP(___timeout___) (void)(___timeout___)
 #endif
 
 #include "can/core/can_writer_task.hpp"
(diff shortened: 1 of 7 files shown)


include/sensors/core/tasks/capacitive_driver.hpp
```diff
@@ -1,7 +1,14 @@
...
 
 #include <array>
-#include <bitset>
+
+#ifdef ENABLE_CROSS_ONLY_HEADERS
+// TODO(fps 7/12/2023): This is super hacky and I hate throwing #ifdefs
+// in our nicely host-independent code but for now we really just need
+// the vTaskDelay function and hopefully sometime in the near future I
+// can refactor this file with a nice templated sleep function.
+#include "FreeRTOS.h"
```
sfoster1, line 10: Lol yuck. Let's at least make this like
```
#ifdef ENABLE_CROSS_ONLY_HEADERS
#include "FreeRTOS.h"
#define HACKY_TASK_SLEEP( ___timeout___ ) vTaskDelay(___timeout___)
#else
#define HACKY_TASK_SLEEP(___timeout___) (void)(___timeout___)
#endif
```
so we keep the ifdefs to just one place and can just do `HACKY_TASK_SLEEP(100);` later

include/sensors/core/tasks/capacitive_driver.hpp
```diff
@@ -38,19 +44,27 @@ class FDC1004 {
...
-        // We should send a message that the sensor is in a ready state,
-        // not sure if we should have a separate can message to do that
-        // holding off for this PR.
-        if (shared_sensor) {
-            measure_mode = fdc1004::MeasureConfigMode::TWO;
+        if (!_initialized) {
+            // FIXME we should grab device id and compare it to the static
+            // device id in code.
+
+            // We should send a message that the sensor is in a ready state,
```
sfoster1, line 51: we'd want a query-response cycle for this rather than just a notification so we can handle if the pipette was up before the server as is common when you boot the machine with the pipettes on it.

include/sensors/core/tasks/capacitive_driver.hpp
```diff
@@ -371,10 +359,46 @@ class FDC1004 {
...
-    std::bitset<2> data_stable{"00"};
     std::array<uint16_t, 2> baseline_results{};
     std::array<uint16_t, 2> polling_results{};
 
+    auto send_followup(i2c::messages::TransactionResponse &m) -> void {
+        auto address = utils::reg_from_id<uint8_t>(m.id.token);
+        uint8_t followup_address = 0;
+        auto tags = utils::tags_from_token(m.id.token);
+        i2c::messages::TransactionIdentifier id = m.id;
+        if (address == static_cast<uint8_t>(fdc1004::Registers::FDC_CONF)) {
```
sfoster1, line 370: kind of a nit but i like this kind of thing a bit more as a `switch` because it lines up nicer

sfoster1 (changes requested):
Couple aesthetic things but if it works it works

sfoster1 (approved):
Implementation looks good, will test!

include/sensors/core/tasks/capacitive_driver.hpp
```diff
@@ -164,66 +176,51 @@ class FDC1004 {
...
-                utils::build_id(fdc1004::ADDRESS, converted_msb_addr, tags));
-        } else {
-            poller.multi_register_poll(
-                fdc1004::ADDRESS, converted_msb_addr, 2, converted_lsb_addr, 2,
-                number_reads, DELAY, own_queue,
-                utils::build_id(fdc1004::ADDRESS, converted_msb_addr, tags));
-        }
+        poller.single_register_poll(
+            fdc1004::ADDRESS,
+            static_cast<uint8_t>(fdc1004::Registers::FDC_CONF), 2, number_reads,
```
Laura-Danielle, line 181: we can just trigger a ready from the FDC_CONf??

sfoster1:
Tested on 96, testing on gripper

sfoster1:
Tested on gripper

fsinapi:
Tested on single, still works fine

PR #704, opened 2023-07-13
fix(motor-control): update position flag inside encoder background timer

ahiuchingau:
We should update the position flag whenever the encoder value is off from where we expect the motor to be, not just during an active move. We can do this in the encoder background timer so that we are re-evaluating the position flag every time we read the encoder pulses.

### DIFF
--- a/include/motor-control/core/stepper_motor/motor_encoder_background_timer.hpp
+++ b/include/motor-control/core/stepper_motor/motor_encoder_background_timer.hpp
@@ -32,6 +32,8 @@ class BackgroundTimer {
         if (!_interrupt_handler.has_active_move()) {
             // Refresh the overflow counter if nothing else is doing it
             std::ignore = _motor_hardware.get_encoder_pulses();
+            // Update position flag if needed
+            std::ignore = _interrupt_handler.check_for_stall();
         }
     }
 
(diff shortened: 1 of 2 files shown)


include/motor-control/core/stepper_motor/motor_encoder_background_timer.hpp
```diff
@@ -32,6 +32,8 @@ class BackgroundTimer {
         if (!_interrupt_handler.has_active_move()) {
             // Refresh the overflow counter if nothing else is doing it
             std::ignore = _motor_hardware.get_encoder_pulses();
```
fsinapi, line 34: Because `stall_detected` already checks the encoder pulses, we can just get rid of this line

fsinapi (approved):
Really nice! The interrupt handler's `stall_detected` function already reads from the encoder though, so we should be able to get rid of the original `get_encoder_pulses` call in the timer.

sfoster1 (approved):
Oh wow this really is the final link in the chain huh

include/motor-control/core/stepper_motor/motor_encoder_background_timer.hpp
ahiuchingau: So true!

fsinapi (approved):
👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

PR #707, opened 2023-07-18
chore: add revisions for PVT boards

sfoster1:
Add revision handling for the variants:
- Gantry C2
- Rear-panel C1 (B2 did not actually exist, keeping it just in case)
- LT pipette D1
- Gripper C2

Head got a bump from C2.0 to C2.1, which we don't track; SOM got a bump but we don't track that; 96 hasn't gotten its bump yet.

These changes are to enter PVT.

## Risk
Medium I guess. These are going to be flashed to real boards and used in real ways, but there's no code changes before and after this, so I don't know.

## Testing
- [x] Put it on each and every one of these boards and boot it
- [x]  check that it reports correctly from bootloader
- [x] check that appropriate entries in the firmware manifest cause automated updates
- [x] check that the updated firmware preserves the revision

### DIFF
--- a/bootloader/firmware/stm32G4/CMakeLists.txt
+++ b/bootloader/firmware/stm32G4/CMakeLists.txt
@@ -93,8 +93,8 @@ endmacro()
 foreach_revision(
     PROJECT_NAME bootloader-gripper
     CALL_FOREACH_REV gripper_bootloader_loop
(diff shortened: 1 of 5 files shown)


Laura-Danielle (approved):
nioce

PR #708, opened 2023-07-19
fix(sensors): reconfigure cap sensor if necessary

fsinapi:
We were having some trouble with the 96 channel attach + calibration workflow. If the 96 channel is newly attached and then the robot server tries to calibrate it, sometimes it will not respond to any capacitive sensor requests. This does not seem to apply to when the pipette is restarted electronically (e.g. with a debugger without de-attaching it), or if it was attached since the robot turned on.

One recent change (#703) made it so that, once the cap sensor is initialized for the first time on startup, the measurement config register (FDC Conf) will never be updated again. Because the sensor task will only return data once the Ready flag is asserted for a channel, the reading register _needs_ to be configured. If the capacitive sensor IC somehow restarted without the MCU restarting, no readings will ever come back. Maybe the MCU is more tolerant of rail fluctuations than the cap IC? Or maybe something else unrelated is causing the configuration reg to be wiped out.

I just tested this by going through the 96 Unnattach flow and then going through the Attach + Calibrate flow. I was not able to recreate the issue after 4 tries.

### DIFF
--- a/include/sensors/core/tasks/capacitive_driver.hpp
+++ b/include/sensors/core/tasks/capacitive_driver.hpp
@@ -216,6 +216,12 @@ class FDC1004 {
             // Start a single-shot, two-register transaction for the data.
             send_followup(m);
         } else {
+            // Double check that the configuration reg seems okay. If not, we
+            // should reconfigure it before re-requesting a reading.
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Looks good to me!

PR #710, opened 2023-07-27
fix(motor-control): do not disengage motor when a StopRequest is received

ahiuchingau:
We should just let the DisengageMotorRequest handle that instead.


fsinapi (approved):
yes

Laura-Danielle (approved):
Nice. Can you also remove this from the `PipetteMotionController` class for the gear/tip/whatever you want to call them motors?

PR #711, opened 2023-08-01
feat(gripper): add stay engaged param to grip message

ahiuchingau:
We want to be able to differentiate between an actual labware grip and a grip that closes the jaw to minimize the footprint of the gripper during gantry movement (we call this the idle position from the api). 

This will inform the firmware whether or not the jaw motor needs to stay engaged when we receive a stop request and prevent unexpected dropping of labware. And if the gripper was in the idle position, the jaw can simply disengage since it does not matter because it is not gripping anything.

The next step is to get gripper to report the real-time jaw state to the hardware controller API (so we don't have to rely soely on the gripper handler instance variable to keep track of the jaw state, which gets lost when the robot server stops).

### DIFF
--- a/include/can/core/messages.hpp
+++ b/include/can/core/messages.hpp
@@ -1005,6 +1005,8 @@ struct GripperGripRequest : BaseMessage<MessageId::gripper_grip_request> {
     uint8_t seq_id;
     brushed_timer_ticks duration;
     uint32_t duty_cycle;
+    int32_t encoder_position_um;
+    uint8_t stay_engaged;
 
     template <bit_utils::ByteIterator Input, typename Limit>
     static auto parse(Input body, Limit limit) -> GripperGripRequest {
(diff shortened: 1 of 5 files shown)


include/can/core/messages.hpp
```diff
@@ -1005,26 +1005,34 @@ struct GripperGripRequest : BaseMessage<MessageId::gripper_grip_request> {
...
+    int32_t encoder_position_um;
+    uint8_t stay_engaged;
 
     template <bit_utils::ByteIterator Input, typename Limit>
     static auto parse(Input body, Limit limit) -> GripperGripRequest {
         uint8_t group_id = 0;
         uint8_t seq_id = 0;
         brushed_timer_ticks duration = 0;
         uint32_t duty_cycle = 0;
+        int32_t encoder_position_um = 0;
```
sfoster1, line 1017: why are we adding this here?

include/can/core/messages.hpp
```diff
@@ -1005,26 +1005,34 @@ struct GripperGripRequest : BaseMessage<MessageId::gripper_grip_request> {
     uint8_t seq_id;
     brushed_timer_ticks duration;
     uint32_t duty_cycle;
+    int32_t encoder_position_um;
```
sfoster1, line 1008: why are we adding this in the _request_ from host?

sfoster1 (changes requested):
Accidental inclusion of gripper position in the request?

include/can/core/messages.hpp
ahiuchingau: It's always part of the payload of the generic gripper move messages, but we never defined it because it is always 0 for the gripper grip request

sfoster1 (approved):
Okay, the position thing makes sense now.

PR #714, opened 2023-08-14
feat(motor-control): report brushed motor state (jaw state)

ahiuchingau:
This PR:
1. adds CAN messages to request and report gripper jaw state from the brushed motor hardware interface
2. simplifies the gripper jaw states to `unhomed`, `force controlling home`, `force controlling` and `position controlling`
3. includes a small refactor to the brushed motor interrupt handler


sfoster1 (approved):
This looks ok to me. I did like the whole precisely one enum for state but i understand why we might want to get rid of it. Do we need this in 7?

ahiuchingau:
Yea this should be in 7.0.0 too

PR #715, opened 2023-08-29
fix(motor-control): critical section for encoders

sfoster1:
When we do the encoder polling timer we really want the motor interrupt to not be able to start in the middle of the timer code, so let's wrap it in a critical section.

## Testing
We saw this in a protocol that was moving the z up and down by 5mm over and over so let's run a protocol that just does that and see if it happens I guess? This is a bug that seems to have occurred once in hundreds or thousands of machine hours of testing so it's going to be tough to say for certain.

It did pass though.


fsinapi (approved):
This looks good to me! My biggest concern would be with this interfering with normal motor operations, so as long as max-speed movements don't seem affected I'm happy with it!

PR #716, opened 2023-08-29
feat(motion-control): handle motor driver errors

pmoegenburg:
# Overview

This PR enables Flex motor drivers (tmc2130/2160) to report when they are in error state due to over-temperature, short circuit, or open circuit. A driver's diag0 pin will now go active when the driver is in error state, and will trigger an interrupt for the specific motion subsystem that will:
    - produce a MotionControllerTask message that will stop the subsystem motion controller
    - set a CancelRequest for the motor interrupt handler, which will produce an ErrorMessage on the CAN bus
    - spawn a SPI message to retrieve the specific error from the motor driver, which will surface as a ReadMotorDriverErrorStatusResponse message on the CAN bus

This PR also produces an ErrorMessage when a motor driver is in error state and an EnableMotorRequest, 
AddLinearMoveRequest, or HomeRequest message is sent to that motion subsystem.

# Test Plan

* [x] Verify pin setup correctly to output low when error state entered
  * Heat driver IC to >150*C (default over-temperature threshold) via heat gun
* [x] Verify interrupt triggered when pin outputs low
* [x] Verify subsystem motion controller enters error state when interrupt triggers
* [x] Verify SPI communication to driver to gather specific error occurs when interrupt triggers
* [x] Verify CAN message reports this specific error
* [x] Verify all of this during an actual move

# Changelog

# Review requests

# Risk assessment

This PR introduces the detection of motor driver errors for each motion subsystem. It will effect each motion subsystem in the ways described above, specifically stopping the subsystem's motion controller and motor interrupt handler and producing an ErrorMessage and ReadMotorDriverErrorStatusResponse message when an error is detected.


include/motor-control/core/tasks/motion_controller_task.hpp
```diff
@@ -140,9 +144,20 @@ class MotionControllerMessageHandler {
         controller.send_usage_data(m.message_index, usage_client);
     }
 
+    // set error flag here (C++ atomic)?! Disable subsystem (enable brakes)?!
+    // cancel_and_clear_moves(can::ids::ErrorCode::motor_driver_error)?
+    void handle(const can::messages::MotorDriverErrorEncountered& m) {
+        controller.stop(); // also request_cancel()?
```
sfoster1, line 150: definitely request_cancel if `stop` doesn't do it

include/gantry/core/interfaces_rev1.hpp
```diff
@@ -9,10 +9,13 @@
 
 namespace interfaces {
 
+extern "C" using diag0_handler =
+    void(*)();  // typedef void (*diag0_handler)(void)
```
sfoster1, line 13: unnecessary comment?

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
```diff
@@ -422,6 +431,11 @@ class MotorInterruptHandler {
             can::messages::ErrorMessage{.message_index = message_index,
                                         .severity = severity,
                                         .error_code = err_code});
+        if (err_code == can::ids::ErrorCode::hardware) {
```
sfoster1, line 434: I think we should make an error code that's specifically "motor driver error detected" if we're going to key "query the motor driver status" off it.

include/motor-control/core/tasks/gear_tmc2160_motor_driver_task.hpp
```diff
@@ -82,11 +82,23 @@ class MotorDriverMessageHandler {
...
 
+    void handle(const can::messages::ReadMotorDriverErrorStatus& m) {
+        LOG("Received read motor driver error register request");
+        uint32_t data = 0;
+        auto converted_addr =
+            static_cast<uint8_t>(tmc2160::registers::Registers::DRVSTATUS);
+        std::array tags{spi::utils::ResponseTag::IS_ERROR_RESPONSE};
+        uint32_t token = spi::utils::build_token(
+            converted_addr, spi::utils::byte_from_tags(tags));
+        driver.read(token, data, m.message_index);
```
sfoster1, line 99: here, for instance, let's make `driver.read()` take a byte of tags or something in addition to the register address

include/motor-control/core/tasks/motion_controller_task.hpp
```diff
@@ -140,9 +144,28 @@ class MotionControllerMessageHandler {
         controller.send_usage_data(m.message_index, usage_client);
     }
 
+    void handle(const can::messages::MotorDriverErrorEncountered& m) {
+        // check if gpio is low before making controller calls
+        controller.stop(can::ids::ErrorSeverity::unrecoverable);
+        if (!controller.is_timer_interrupt_running()) {
```
sfoster1, line 150: probably want to switch the sense of this, i.e.
```
if (controller.is_timer_interrupt_running() {
    controller.stop()
} else {
    can_client.send_can_message...
}
```

include/motor-control/core/tasks/motion_controller_task.hpp
```diff
@@ -187,8 +212,22 @@ class MotionControllerTask {
...
+    void run_diag0_interrupt() {
+        if (!driver_error_handled()) {
+            static_cast<void>(
+                queue.try_write_isr(can::messages::MotorDriverErrorEncountered{
+                    .message_index = 0}));
+        }
+    }
+
+    auto driver_error_handled() -> bool {
+        return driver_error_handled_flag.exchange(true);
```
sfoster1, line 225: what resets this flag? as coded, it looks like a given microcontroller will only ever respond to a driver error once until it's rebooted. we probably need a way to reset this when the error goes away.

include/motor-control/core/tasks/tmc2130_motor_driver_task.hpp
```diff
@@ -81,11 +81,23 @@ class MotorDriverMessageHandler {
         LOG("Received read motor driver request: addr=%d", m.reg_address);
         uint32_t data = 0;
         if (tmc2130::registers::is_valid_address(m.reg_address)) {
-            driver.read(tmc2130::registers::Registers(m.reg_address), data,
-                        m.message_index);
+            auto converted_addr = static_cast<uint8_t>(m.reg_address);
+            uint32_t token = spi::utils::build_token(converted_addr);
+            driver.read(token, data, m.message_index);
```
sfoster1, line 86: same as the 2160 with the `read()` call - this should still take a register address and also take a byte of tags or something

include/motor-control/core/tasks/tmc2160_motor_driver_task.hpp
```diff
@@ -80,11 +94,23 @@ class MotorDriverMessageHandler {
...
 
+    void handle(const can::messages::ReadMotorDriverErrorStatus& m) {
+        LOG("Received read motor driver error register request");
+        uint32_t data = 0;
+        auto converted_addr =
+            static_cast<uint8_t>(tmc2160::registers::Registers::DRVSTATUS);
+        std::array tags{spi::utils::ResponseTag::IS_ERROR_RESPONSE};
+        uint32_t token = spi::utils::build_token(
+            converted_addr, spi::utils::byte_from_tags(tags));
+        driver.read(token, data, m.message_index);
```
sfoster1, line 111: same as the other motor driver stuff - `read()` should take a register address and a byte of tags

gantry/firmware/stm32g4xx_it.c
```diff
@@ -147,6 +147,7 @@ void FDCAN1_IT0_IRQHandler(void) {
  */
 void TIM7_IRQHandler(void) { HAL_TIM_IRQHandler(&htim7); }
 void TIM2_IRQHandler(void) { HAL_TIM_IRQHandler(&htim2); }
+void EXTI9_5_IRQHandler(void) { HAL_GPIO_EXTI_IRQHandler(GPIO_PIN_5); } // source this from somewhere?
```
sfoster1, line 150: what's the comment about?

include/motor-control/core/motor_hardware_interface.hpp
```diff
@@ -72,6 +72,12 @@ class UsageEEpromConfig {
     size_t num_keys = 0;
 };
 
+// std::optional usage? See HardwareConfig struct
+struct __attribute__((packed)) CancelRequest {
+    uint8_t severity;
```
sfoster1, line 77: investigate if we can use the actual enum types here

include/motor-control/core/brushed_motor/brushed_motor_interrupt_handler.hpp
```diff
@@ -201,7 +201,7 @@ class BrushedMotorInterruptHandler {
         } else if (estop_triggered()) {
             in_estop = true;
             cancel_and_clear_moves(can::ids::ErrorCode::estop_detected);
-        } else if (hardware.has_cancel_request()) {
+        } else if (hardware.has_cancel_request().code != 0U) {
```
sfoster1, line 204: this code should definitely be comparing enum values; ideally the `CancelRequest` struct will have the enum values in it but if not let's do this compare in a static cast

include/motor-control/firmware/stepper_motor/motor_hardware.hpp
```diff
@@ -53,13 +54,25 @@ class MotorHardware : public StepperMotorHardwareIface {
...
     void reset_encoder_pulses() final;
-    auto has_cancel_request() -> bool final {
-        return cancel_request.exchange(false);
-    }
     void disable_encoder() final;
     void enable_encoder() final;
 
-    void request_cancel() final { cancel_request.store(true); }
+    // need any std::optional usage?
+    auto has_cancel_request() -> CancelRequest final {
```
sfoster1, line 61: IMO here this one wants to be called `get_cancel_request`, `has_cancel_request` really makes me think it returns a bool

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
```diff
@@ -202,9 +207,12 @@ class MotorInterruptHandler {
         } else if (estop_triggered()) {
             cancel_and_clear_moves(can::ids::ErrorCode::estop_detected);
             in_estop = true;
-        } else if (hardware.has_cancel_request()) {
-            cancel_and_clear_moves(can::ids::ErrorCode::stop_requested,
-                                   can::ids::ErrorSeverity::warning);
+        } else if (cancel_request.code !=
+                   0U) {  // is this correct? Should be zero-initialized. Check,
```
sfoster1, line 211: should be checking for enum value

include/motor-control/core/tasks/tmc2130_motor_driver_task.hpp
```diff
@@ -86,6 +86,15 @@ class MotorDriverMessageHandler {
         }
     }
 
+    void handle(const can::messages::ReadMotorDriverErrorStatus& m) {
```
sfoster1, line 89: this class's spi response handler needs to look at the tag like in the 2160 driver right

head/firmware/main_rev1.cpp
```diff
@@ -432,11 +451,12 @@ auto main() -> int {
 
     utility_gpio_init();
     can_bus_1.start(can_bit_timings);
-    head_tasks::start_tasks(can_bus_1, motor_left.motion_controller,
-                            motor_right.motion_controller, psd, spi_comms2,
-                            spi_comms3, motor_driver_configs_left,
-                            motor_driver_configs_right, rmh_tsk, lmh_tsk,
-                            i2c_comms3, eeprom_hw_iface);
+    std::tie(call_diag0_z_handler, call_diag0_a_handler) =
```
sfoster1, line 454: i love `std::tie` so much. the most c++ answer to destructuring.

pipettes/firmware/hardware_config.c
```diff
@@ -140,7 +146,7 @@ static uint16_t get_spi_pins_lt(GPIO_TypeDef* for_handle) {
     }
 }
 
-static uint16_t get_motor_driver_pins_ht(GPIO_TypeDef* for_handle) {
+static uint16_t get_motor_driver_pins_ht(GPIO_TypeDef* for_handle, bool for_diag0) {
```
sfoster1, line 149: I'm always uncomfortable with having a bool in a function argument that makes it return a completely different thing. Let's either have this return a quick struct by-value or have a second function that's like `get_diag0_pin_{ht,lt}`.

include/pipettes/core/tasks/motion_controller_task.hpp
```diff
@@ -117,9 +154,52 @@ class MotionControllerMessageHandler {
...
+                            .message_index = m.message_index,
+                            .severity = can::ids::ErrorSeverity::unrecoverable,
+                            .error_code = can::ids::ErrorCode::
+                                motor_driver_error_detected});
+                    driver_client.send_motor_driver_queue(
+                        can::messages::ReadMotorDriverErrorStatusRequest{
+                            .message_index = m.message_index});
+                }
+            } else {
+                controller.clear_cancel_request();
```
sfoster1, line 177: Do we need this? If we're here, then this is the first time that we've interacted with this message, and therefore we haven't set a cancel request yet. It kind of seems like this would either (a) do nothing because there's no cancel request or (b) clear a cancel request from somewhere else which we definitely don't want.

include/motor-control/core/tasks/motion_controller_task.hpp
```diff
@@ -136,13 +183,55 @@ class MotionControllerMessageHandler {
...
         }
     }
+
     void handle(const can::messages::GetMotorUsageRequest& m) {
         controller.send_usage_data(m.message_index, usage_client);
     }
 
+    void handle(
+        const motor_control_task_messages::RouteMotorDriverInterrupt& m) {
+        if (m.debounce_count > 9) {
```
sfoster1, line 193: this `9` value and the delay value below should probably be constexprs defined somewhere prominent at the top of the file - they're key values for some behavior and it would be good to find it in a consistent place

include/motor-control/core/tasks/motion_controller_task.hpp
```diff
@@ -136,13 +183,55 @@ class MotionControllerMessageHandler {
...
+                            .message_index = m.message_index,
+                            .severity = can::ids::ErrorSeverity::unrecoverable,
+                            .error_code = can::ids::ErrorCode::
+                                motor_driver_error_detected});
+                    driver_client.send_motor_driver_queue(
+                        can::messages::ReadMotorDriverErrorStatusRequest{
+                            .message_index = m.message_index});
+                }
+            } else {
+                controller.clear_cancel_request();
```
sfoster1, line 211: similar thing with the `clear_cancel_request`

include/motor-control/core/tasks/gear_tmc2160_motor_driver_task.hpp
```diff
@@ -87,6 +87,15 @@ class MotorDriverMessageHandler {
         }
     }
 
+    void handle(const can::messages::ReadMotorDriverErrorStatusRequest& m) {
+        LOG("Received read motor driver error register request");
+        uint32_t data = 0;
+        std::array tags{spi::utils::ResponseTag::IS_ERROR_RESPONSE};
```
sfoster1, line 93: doesn't look like the `IS_ERROR_RESPONSE` tag is handled in the `TransactResponse` message handler above

include/motor-control/core/stepper_motor/tmc2160_driver.hpp
```diff
@@ -295,9 +315,9 @@ class TMC2160 {
      * be read, so this function gets it from the actual device.
      */
     auto update_glob_scaler(uint32_t data) -> void {
-        auto ret = read_register<GConfig>(data);
+        auto ret = read_register<GConfig>(data);  // should be GlobalScalar
```
sfoster1, line 318: yeah it really should be huh. @ahiuchingau what do you think of this? Might impact some current-control stuff. @pmoegenburg would you mind opening a second quick PR to fix this and removing the comments here because otherwise they'll cause merge conflicts? Great find.

sfoster1 (changes requested):
This is looking really good, just a couple small changes I think we should see. This is gonna be a huge step forward!

include/pipettes/core/tasks/motion_controller_task.hpp
pmoegenburg: Good call!

include/motor-control/core/tasks/gear_tmc2160_motor_driver_task.hpp
pmoegenburg: Great catch!

PR #717, opened 2023-09-01
feat(all): allow for up to 12 steps in each move group

fsinapi:
In https://github.com/Opentrons/opentrons/pull/13451, we increase the maximum number of steps that may be generated in a move group. We need to be able to accept all of those steps on the firmware side.

The only real downside here is RAM usage. I was able to build everything locally, but we obviously need to make sure CI is able to link everything as well. If there are any problems I think the best thing to do is cut down on the total move groups, since in practice we only use a few of the allotted 6.


sfoster1 (approved):
Looks good to me, wouldn't even really mind preemptively reducing move groups to 3.

PR #720, opened 2023-09-14
fix(gripper): do not set pwm to 0

ahiuchingau:
This currently doesn't affect the gripper behavior because we don't hold jaw width at all at the moment. 

So we don't have to merge this just yet and we should test this to make sure it's fine!


ahiuchingau:
Closing this for now since we haven't been using the position controlled moves with the gripper.

PR #721, opened 2023-09-14
fix(motor-control): do not enable motor when receives a zero-length move

ahiuchingau:
We shouldn't need to to engage the motors when we're not planning on moving them. 

This might cause a lot of weird motor position errors so... gonna keep this as a draft as i do more testing and add more logic on the python side to handle this better.

### DIFF
--- a/include/motor-control/core/stepper_motor/motion_controller.hpp
+++ b/include/motor-control/core/stepper_motor/motion_controller.hpp
@@ -73,7 +73,7 @@ class MotionController {
             .seq_id = can_msg.seq_id,
             .stop_condition = can_msg.request_stop_condition,
             .usage_key = hardware.get_usage_eeprom_config().get_distance_key()};
-        if (!enabled) {
+        if (!enabled && velocity_steps != 0) {
             enable_motor();
         }
         queue.try_write(msg);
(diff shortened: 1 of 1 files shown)


include/motor-control/core/stepper_motor/motion_controller.hpp
```diff
@@ -73,7 +73,7 @@ class MotionController {
             .seq_id = can_msg.seq_id,
             .stop_condition = can_msg.request_stop_condition,
             .usage_key = hardware.get_usage_eeprom_config().get_distance_key()};
-        if (!enabled) {
+        if (!enabled && velocity_steps != 0) {
```
fsinapi, line 76: Would also want to check `acceleration_steps != 0`, otherwise the motor won't get enabled for a move step that accelerates from 0

ahiuchingau: good call!

sfoster1:
I don't know that I agree with this. I think the common case is that we _do_ want these motors enabled so that if the machine rocks around or something they won't move.

We do actually have a good way to keep these axes disabled: don't send the moves to them. I think if there's cases where we want the motors to not enable if they're not moving, python should trim out the `AddLinearMoveRequests` from them.

PR #722, opened 2023-09-19
feat(gripper): report jaw encoder pulses when something goes wrong

ahiuchingau:
We're using the UpdatePositionEstimation message to report the jaw encoder position right when the move fails, this message will be ignored by the host but we can observe it on the can bus and will make debugging easier.


sfoster1 (approved):
Looks good to me!

PR #723, opened 2023-09-19
feat(pipettes): require overpressure for 500ms before error

fsinapi:
The intent of the overpressure warning is to avoid damage to the pressure sensor. The sensor will only give us a reading up to ~8kPa, but the actual pressure it can withstand is far higher. In an effort to avoid overeager protocol-ending errors, we are adding a requirement that the pressure remains over the threshold for at least 500MS before throwing an error.

This PR also adds reporting of any readings over the threshold.

Tested on Flexy by blocking the pipette nozzle while aspirating/dispensing. You can see in the can logs that you get a bunch of Pressure Reading responses before they start including error messages. Short aspiration/dispense distances don't have enough time to trigger the error.

### DIFF
--- a/include/sensors/core/tasks/pressure_driver.hpp
+++ b/include/sensors/core/tasks/pressure_driver.hpp
@@ -147,9 +147,13 @@ class MMR920C04 {
     }
 
     auto poll_continuous_pressure(uint8_t tags) -> void {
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Excellent, thank you!

PR #724, opened 2023-09-19
feat(push): add option to specify ssh key when pushing firmware to ot3

fsinapi:
🔑


vegano1:
uff, good call

sfoster1 (approved):
keys open doors

PR #725, opened 2023-09-19
feat(gripper): allow the encoder value to debounce if deviated from expected val

ahiuchingau:
There's potential noise on the gripper jaw encoder line that causes the gripper to falsely report a move has failed. We're now confirming the encoder value has deviated from the expected val twice consecutively when executing an idle move before reporting it has failed.


sfoster1 (approved):
This makes sense, looks good to me once tested!

PR #727, opened 2023-09-21
fix(sensors): Reduce overpressure time to 200ms

sfoster1:
This is a little safer on the 1000ul which is otherwise capable of driving enough pressure changes to cause a problem.


Laura-Danielle (approved):
numby

PR #728, opened 2023-10-03
fix(pipettes): changes to enable the correct sync line ports for the 96-channel

ahiuchingau:
The GPIO ports were wrong unfortunately.


caila-marashaj (approved):

sfoster1 (approved):
Looks good once we fix tests!

PR #730, opened 2023-10-13
only autmomatically notify about low throughput tip changes

caila-marashaj:
## Overview
On the 96 channel pipette, readings of the tip presence sensor are only meaningful when the gear motors are moved down far enough to drop the shroud below the optical sensor. If tips are accidentally dropped during a protocol, we wouldn't actually know until we check by commanding the gear motors to move. Therefore, we don't need an interrupt for changes in the 96 channel's sensor status.

## Changelog
- Initialize the tip presence sensor for the 96 channel as a regular gpio input, and keep interrupt initialization for the single and 8 channel pipettes
- Remove the check in the interrupt's callback for whether the gpio pin is the `primary` or `secondary` sensor, as this only applies to the 96 channel
- Invert the active setting of the gpio pin only for the 96 channel, so the 96 channel's response logic matches the single and 8 channel's

### DIFF
--- a/pipettes/firmware/main.cpp
+++ b/pipettes/firmware/main.cpp
@@ -138,11 +138,12 @@ static auto tail_accessor =
     eeprom::dev_data::DevDataTailAccessor{sensor_queue_client};
 
 extern "C" void HAL_GPIO_EXTI_Callback(uint16_t GPIO_Pin) {
-    if (GPIO_Pin == tip_sense_gpio_primary.pin &&
-        sensor_queue_client.tip_notification_queue_rear != nullptr) {
-        static_cast<void>(
-            sensor_queue_client.tip_notification_queue_rear->try_write_isr(
-                sensors::tip_presence::TipStatusChangeDetected{}));
+    if (GPIO_Pin == tip_sense_gpio_primary.pin) {
+        if (sensor_queue_client.tip_notification_queue_rear != nullptr) {
+            static_cast<void>(
+                sensor_queue_client.tip_notification_queue_rear->try_write_isr(
+                    sensors::tip_presence::TipStatusChangeDetected{}));
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


pipettes/firmware/utility_gpio.c
```diff
@@ -52,24 +42,10 @@ static void tip_sense_gpio_init() {
...
-        HAL_NVIC_SetPriority(block_15_10, 10, 0);
-        HAL_NVIC_EnableIRQ(block_15_10);
-    } else {
-        IRQn_Type block_2 = get_interrupt_line(gpio_block_2);
-        /* EXTI interrupt init block tip sense*/
-        HAL_NVIC_SetPriority(block_2, 10, 0);
-        HAL_NVIC_EnableIRQ(block_2);
-
-    }
+    IRQn_Type block_2 = get_interrupt_line(gpio_block_2);
```
ahiuchingau, line 45: We should only enable these interrupt when it's not a 96

pipettes/firmware/main.cpp
```diff
@@ -141,20 +141,12 @@ static auto tail_accessor =
     eeprom::dev_data::DevDataTailAccessor{sensor_queue_client};
 
 extern "C" void HAL_GPIO_EXTI_Callback(uint16_t GPIO_Pin) {
-    if (GPIO_Pin == tip_sense_gpio_primary.pin) {
+    if (PIPETTE_TYPE != NINETY_SIX_CHANNEL) {
```
ahiuchingau, line 144:
```suggestion
    if (PIPETTE_TYPE != NINETY_SIX_CHANNEL && GPIO_Pin == tip_sense_gpio_primary.pin)) {
```

fsinapi (approved):
nice this LGTM but we should also not enable the IRQ for the tip sense interrupt on the 96ch (in `nvic_priority_enable_init`)

sfoster1 (approved):
Looks good to me if tested and works!

PR #731, opened 2023-10-19
fix(motor-control): monitor encoder during empty moves

fsinapi:
RQA-1812, and possibly some other ABR failures, seems like it may be a result of an axis encoder overflowing multiple times during an empty move. We have an encoder background timer that runs when the interrupt handler doesn't have an active move, but that doesn't account for the fact that we often set up motor axes with empty moves (no acceleration & no velocity) matching the duration of the move groups for active axes. 

To fix this edge case, we should also check the velocity of the buffered move when the encoder timer is determining whether it should read the encoder. This PR adds a new function in the interrupt handler to return whether 1) there is a buffered move 2) the buffered move's velocity is not 0.

We only access the function in a critical section so I don't think it matters that the velocity isn't fully atomic.

Uploaded to Flexy. Threw the gantry around with the Estop on and was able to rehome fine, tried backdriving various axes while they were idle during other movements, everything seemed to get detected fine (rehoming would fast-home to the right place). Pretty confident this isn't introducing any _new_ issues, but I think the only way to really test if this is a fix is to put it on the ABR bots that have had this problem.

### DIFF
--- a/include/motor-control/core/stepper_motor/motor_encoder_background_timer.hpp
+++ b/include/motor-control/core/stepper_motor/motor_encoder_background_timer.hpp
@@ -32,7 +32,7 @@ class BackgroundTimer {
     auto callback() -> void {
         auto critical_section =
             freertos_synchronization::FreeRTOSCriticalSectionRAII();
-        if (!_interrupt_handler.has_active_move()) {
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
Yeah this looks good to me, it's the fix we talked about.

PR #732, opened 2023-11-07
fix(head): engage ebrake at boot

ahiuchingau:
The ebrakes shall be turned on at boot.


sfoster1 (approved):
Yes please. I'll test this.

sfoster1 (approved):
Definitely does not make things worse!

PR #733, opened 2023-11-08
fix(motor-control): z stage falls after being disengaged

ahiuchingau:
This worked somehow? Need way more testing to verify this fix


caila-marashaj (approved):
riveting

sfoster1 (approved):
Looks good to me!

PR #734, opened 2023-11-27
fix(motor-control): only add a delay before disengaging motor if ebrake exists

ahiuchingau:
We shouldn't have added the delay if the ebrake doesn't exist. 

Also fixing an old commit message to actually show what it does.


ahiuchingau:
Yes will be squash-merging this guy once i get it tested.

PR #735, opened 2023-12-08
feat(gripper): add can message to handle updating gripper jaw idle value

ahiuchingau:
This PR lets us update the gripper jaw holdoff tick value, which affects how long the gripper jaw encode value has to remain stable before reporting the gripper motion is completed. 

We currently hardcoded this value to be 32 ticks in the firmware (which is exactly 1 ms), but we want the ability to modify this value when necessary. Increasing this value, in theory, should let the gripper hold on to the labware better.


sfoster1 (changes requested):
can we make the message use ms?

sfoster1 (approved):
Nice, I really like folding this into the error config

PR #737, opened 2024-01-02
fix(motor): estop release causes z to drop

ahiuchingau:
We never disengage the motors when the estop is activated; so when the estop is released, the ebrake also releases which means now the motor control is switched back to relying on the driver's hold current. Sometimes the z stage would fall because hold current is not enough to hold it in place. 

This PR disables the z motor (also engage the ebrake) whenever the estop pin is pulled high. This will ensure the ebrake stays engaged when the estop is released and prevents the z stages from falling.

### DIFF
--- a/head/firmware/utility_hardware.c
+++ b/head/firmware/utility_hardware.c
@@ -62,9 +62,12 @@ void estop_input_gpio_init() {
     /*Configure GPIO pin EStopin : PB4 */
     GPIO_InitTypeDef GPIO_InitStruct = {0};
     GPIO_InitStruct.Pin = GPIO_PIN_4;
-    GPIO_InitStruct.Mode = GPIO_MODE_INPUT;
+    GPIO_InitStruct.Mode = GPIO_MODE_IT_RISING;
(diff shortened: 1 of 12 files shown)


head/firmware/utility_hardware.c
```diff
@@ -62,30 +62,15 @@ void estop_input_gpio_init() {
...
+    GPIO_InitStruct.Mode = GPIO_MODE_IT_RISING;
     GPIO_InitStruct.Pull = GPIO_NOPULL;
     HAL_GPIO_Init(GPIOB, &GPIO_InitStruct);
-}
 
-void ebrake_gpio_init() {
-    __HAL_RCC_GPIOB_CLK_ENABLE();
-    // left brake PB0
-    // right brake PB5
-    GPIO_InitTypeDef GPIO_InitStruct = {0};
```
sfoster1, line 74: wait why get rid of this stuff?

sfoster1 (changes requested):
Why get rid of the `ebrake_gpio_init`?

sfoster1 (approved):
offline explanation, looks good to me once tested!

PR #738, opened 2024-01-04
refactor(pipettes): add new scaling and max pressure for mmr920c10

caila-marashaj:
## Overview
The conversion factor as well as the new max pressure we should allow for the new pressure sensor(mmr920c10) are double that of the mmr920c04. This just updates the pressure sensor's header file to reflect that.


sfoster1 (changes requested):
Two things:
1. Let's get ahead of the game since we probably won't actually have this be a revision pending discussion and make the sensor type determination a runtime check that happens during the execution of a task, so it'll be easier to switch it to something based on reading an eeprom or whatever
2. Let's consider how, given that it probably won't be a major revision, we can communicate to the python side that this is the new sensor capability and range. A new message of some kind?

sfoster1 (approved):
ok, let's keep the revision thing for now and we can change it later.

include/sensors/core/mmr920.hpp
```diff
@@ -26,12 +26,16 @@
...
  */
 namespace sensors {
-namespace mmr920C04 {
+namespace mmr920 {
 constexpr uint16_t ADDRESS = 0x67 << 1;
 
-// Pressure cannot be measured beyond +/-8226.4F
+// Pressure cannot be measured beyond +/-8226.4F on the old sensors
+// New sensors have double the threshold but half the fidelity
+#if PCBA_PRIMARY_REVISION == 'e'
```
sfoster1, line 34: Regardless of whether we keep the revision handling or some other method, we should move this ifdef to somewhere closer to an application root, and have this file parametrized on what kind of sensor it should represent.

sfoster1 (changes requested):
We should move the #if stuff into somewhere closer to an application root, and have the header be parametrized somehow. The _sensor_ shouldn't care what revision of pipette it's on - rather, the pipette application should care, and request the appropriate revision from the sensor.

sfoster1 (approved):
Excellent, looks much better, thank you!

PR #739, opened 2024-01-05
chore(state-manager): update state manager python version & update dependencies

ahiuchingau:
Our OT3 is running on Python 3.10 now, this means the state manager environment should also reflect that. 

### Changelog
* add `main-setup` Github action that installs the required version of gcc and python for reusability
* replace old workflows with:
  * `cross-compile-all`: build/format/lint all firwmare executable targets in parallel
  *  `build-simulators`: build simulators for targets and upload them as artifacts
  * `build-tests`: build-and-test

This hopefully makes it easier for us to add firmware targets/update dependencies in the future.


sfoster1 (approved):
noice ty

.github/workflows/build-simulators.yaml
```diff
@@ -0,0 +1,74 @@
...
+        run: cmake --build ./build-host --target state-manager-setup
+
+      - name: Build Simulator
+        run: cmake --build ./build-host --target ${{ matrix.target }}-simulator
+
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.target }}-simulator-${{ github.ref_name }}
+          path: |
```
sfoster1, line 67: could be `./ot3-firmware/build-host/*/simulator/*-simulator` or something I think

sfoster1 (approved):
Looks good to me! I think the upload path for the simulators can just be done with shell globbing, might be a bit simpler, but that's a small thing

PR #743, opened 2024-01-18
fix(hepa-uv): add hepa-uv to firmware manifest json file.

vegano1:
This is required to add the hepa-uv entry to the `opentrons-firmware.json` manifest file used by the Flex to perform firmware updates.


sfoster1 (approved):
Looks good to me!

PR #744, opened 2024-01-18
feat(hepa-uv): Add a task to handle the UV Ballast state based on external input

vegano1:
## Overview

The HEPA/UV has a UV Ballast that needs to turn on/off based on external input, this pr adds a a task that actuates the UV ballast based on these external inputs. It also adds a timer that turns off the UV ballast after 15m.

Closes: [RET-1402](https://opentrons.atlassian.net/browse/RET-1402)

## Changes

- Add a UVTask that listens for irq interrupts and turns on/off the UV Ballast accordingly
- Turn off the HEPA fan and UV Ballast on startup

## Review Requests

- Does this approach make sense?
- I'll have to go back in at some point to add a better interface instead of using gpio::set/reset directly
- Feedback is appreciated!

[RET-1402]: https://opentrons.atlassian.net/browse/RET-1402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/include/hepa-uv/core/uv_task.hpp
+++ b/include/hepa-uv/core/uv_task.hpp
@@ -1,0 +1,90 @@
+#pragma once
+
+#include "can/core/ids.hpp"
+#include "common/core/bit_utils.hpp"
+#include "common/core/logging.h"
+#include "common/core/message_queue.hpp"
+#include "hepa-uv/firmware/gpio_drive_hardware.hpp"
+#include "messages.hpp"
+
+namespace uv_task {
+
+using TaskMessage = interrupt_task_messages::TaskMessage;
+
+class UVMessageHandler {
+  public:
+    explicit UVMessageHandler(gpio_drive_hardware::GpioDrivePins &drive_pins)
+        : drive_pins{drive_pins} {}
+    UVMessageHandler(const UVMessageHandler &) = delete;
+    UVMessageHandler(const UVMessageHandler &&) = delete;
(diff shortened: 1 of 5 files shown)


include/hepa-uv/core/uv_task.hpp
```diff
@@ -0,0 +1,132 @@
...
+            reed_switch_set = gpio::is_set(drive_pins.reed_switch);
+        }
+
+        // reset push button state if the door is opened or the reed switch is
+        // not set
+        if (!door_closed || !reed_switch_set) uv_push_button = false;
+
+        // set the UV Ballast
+        if (door_closed && reed_switch_set && uv_push_button) {
+            gpio::set(drive_pins.uv_on_off);
```
sfoster1, line 74: what happens if i push the button twice?

sfoster1 (changes requested):
I think we need to track whether the ballast is currently active because otherwise we can start the timer multiple times with multiple presses

include/hepa-uv/core/uv_task.hpp
vegano1: We toggle the uv_push_button every time it's pressed on line 61.
Assuming the ballast is on and the timer is running, pressing the button again would turn it off and cancel the timer.

sfoster1 (approved):
Ah, didn't see the toggle. Looks good to me.

PR #745, opened 2024-01-20
feat(hepa-uv): Configure Push button led timers and add led hardware to drive leds.

vegano1:
## Overview

We need to set the push button LEDs based on the system state or when the user interacts with the HEPA/UV module, this PR adds that functionality.

Closes: [RET-1403](https://opentrons.atlassian.net/browse/RET-1403)

##  Change log

- Added configuration for hardware timers required to drive the push button LEDs
- Added led_hardware to drive the push button LEDs by setting the duty cycle
- Added LEDControlTask to manage push button state change requests
- Added LEDControlTask client (queue) instance to the HEPA and UV tasks so we can change the push button leds
- Set the push button led states based on the current state of the system

## Review Requests

- Is anything not making sense or should it be different?

[RET-1403]: https://opentrons.atlassian.net/browse/RET-1403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ahiuchingau:
> Definitely looks like it will work and good to merge as is (and honestly, if the LEDs are only ever gonna turn on and off based on the button - especially for the fan - I don't think there's necessarily a reason to make them their own task. IMO we only need the task if there's multiple sources of input, or if we want to do animations that need their own source of timing). I think you could condense the hardware initialization somewhat, though.

I'm trying to understand the hepa-uv system - what's initiating the LED state change request by sending the LED task message?

vegano1:
> > Definitely looks like it will work and good to merge as is (and honestly, if the LEDs are only ever gonna turn on and off based on the button - especially for the fan - I don't think there's necessarily a reason to make them their own task. IMO we only need the task if there's multiple sources of input, or if we want to do animations that need their own source of timing). I think you could condense the hardware initialization somewhat, though.
> 
> I'm trying to understand the hepa-uv system - what's initiating the LED state change request by sending the LED task message?

Good callout, I was calling the led_hardware directly to change the LEDs and not using the queues.
I just added that.

sfoster1 (approved):
nice, looks good to me!

PR #746, opened 2024-01-23
fix(hepa-uv): remove hepa-uv simulators so they don't block CI.

vegano1:
The functionality for the HEPA/UV module is not complete yet, so let's remove the simulator build so it does not stop other simulators from building. This will be re-added in a future PR.


ahiuchingau (approved):
Great! Thanks

PR #747, opened 2024-01-23
feat(hepa-uv): Add the ability to set the duty cycle for the HEPA fan.

vegano1:
## Overview

We needed to set the duty cycle that will let the HEPA fan spin, this PR does that by adding an interface that lets us set the PWM of the timer corresponding to the HEPA_PWM pin.

Closes: [RET-1409](https://opentrons.atlassian.net/browse/RET-1409)

## Change log

- Add the timer configuration for the HEPA_PWM timer
- Add a function to set the PWM for the HEPA fan
- Set the duty cycle of the hepa fan from the HEPATask when the user presses the push button

## Review Requests

Is anything looking weird?
- I wanted to consolidate the hardware initialization and driving functions into one file, so I renamed led_hardware.h/cpp to hepauv_hardware.h/cpp, is this too broad? it deals with initializing the timers for the LEDs and hepa fan, as well as setting the LEDs and HEPA fan duty cycle.

[RET-1409]: https://opentrons.atlassian.net/browse/RET-1409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hepa-uv/firmware/hepauv_hardware.c
```diff
@@ -162,13 +196,18 @@ static void MX_TIM_Init(TIM_TypeDef* tim) {
...
-     */
-    htim->Init.Prescaler =
-        calc_prescaler(SystemCoreClock, LED_TIMER_FREQ);
     htim->Init.CounterMode = TIM_COUNTERMODE_UP;
-    htim->Init.Period = LED_PWM_WIDTH - 1;
+
+    // Set prescaler and period for hepa_pwm timer (TIM3)
+    if (tim == TIM3) {
+        htim->Init.Prescaler = 64 - 1;
+        htim->Init.Period = 40 - 1;
```
sfoster1, line 204: this is not a lot of control authority for this PWM, I think. You'll only be able to have 40 levels. Maybe it's fine but something to be aware of

hepa-uv/firmware/hepauv_hardware.c
```diff
@@ -1,16 +1,18 @@
-#include "hepa-uv/firmware/led_hardware.h"
+#include "hepa-uv/firmware/hepauv_hardware.h"
```
sfoster1, line 1: i think if you want to merge things like this you should instead restructure into timer_hardware and button_hardware or something

hepa-uv/firmware/hepauv_hardware.c
```diff
@@ -273,25 +312,35 @@ void set_button_led_pwm(PUSH_BUTTON_TYPE button, uint32_t red, uint32_t green, u
     button_led_hw_update_pwm(white, WHITE_LED, button);
 }
 
-void button_led_hw_initialize_leds() {
+void set_hepa_fan_pwm(uint32_t duty_cycle) {
+    // update hepa fan speed
+    htim3.Instance->CCR1 = duty_cycle;
```
sfoster1, line 317: if you have this done this way, giving this function a value over the timer's period register will result in the output always being set which I think is not what you want. It will also only represent a duty cycle if the period is exactly 100. I'd recommend doing something like `htim3.Instance->CCR1 = max(duty_cycle, 100)*100 / htim3.Instance->Period` or something, actually calculate it

sfoster1 (changes requested):
I think some little stuff needs to be tweaked around the pwm timer config

sfoster1 (approved):
Looks good to me!

PR #749, opened 2024-01-29
feat(pipettes): sync moves with pressure sensor reading

caila-marashaj:
## Overview
This branch allows us to start and stop the pressure sensor in sync with the execution of a pipette move group. 

It seems pretty involved for what it does currently, but I did it this way because I think it will make it easier in the near future to build code for collecting and packaging large amounts of sensor data while the move is happening.

It also is pretty messy with `#ifdef` calls at the moment, but it has potential to be used for any sensor and could be nice to have for testing in the future, so any suggestions about making that a bit cleaner are much appreciated.

## Changelog
- creates a derived `PipetteMotorInterruptHandler` class, which should allow us to send `BindSensorOutputRequest` messages directly to the pressure sensor task


sfoster1 (approved):
Let's get a Ci task that builds in the new configuration too to make sure nobody accidentally breaks it, and maybe see if we can add a bit in the device info?

PR #751, opened 2024-02-01
fix(motor-control): check for encoder overflow more often

ahiuchingau:
Currently we only check for encoder overflow when the motor is actively stepping. And if the motor velocity is very slow, we could potentially miss an overflow or two because of the encoder jitters. The fix here is to check for it more often. This will now allow our axes to move much more slowly.


sfoster1 (approved):
Well heck that's a good one

PR #753, opened 2024-02-08
feat(hepa-uv): Add CAN messages to control the Hepa/UV filter.

vegano1:
# Overview

We need to control the Hepa/UV filter from the Flex for testing and eventual system integration, this PR implements the CAN messages required to make this happen. It also adds a new method to query the time remaining before a software timer expires. This PR coincides with [Opentrons/opentrons#14452](https://github.com/Opentrons/opentrons/pull/14452) PR.

Closes: [RET-1422](https://opentrons.atlassian.net/browse/RET-1422) [RET-1423](https://opentrons.atlassian.net/browse/RET-1423)

# Test Plan

- [x] Make sure the Hepa fan still turns on/off with the push button
- [x] Make sure the UV Ballast still turns on/off with the push button when the Flex door is closed and reed switch is set.
- [x] Make sure that we can set/get the Hepa fan state and duty cycle with the SetHepaFanStateRequest and GetHepaFanStateRequest messages.
- [x] Make sure that we can set/get the UV light state and timeout with the SetHepaUVStateRequest and GetHepaUVStateRequest messages.
- [x] Make sure that the UV light ONLY turns on via CAN message when the door is closed and reed switch is set.
- [x] Make sure that the UV light turns OFF when the door is opened
- [x] After turning ON the UV light via a CAN message, make sure we can turn off the light when the push button is pressed.

# Change Log

- Added implementation for CAN messages to set/get Hepa fan state/duty cycle
- Added implementation for CAN messages to set/get UV Light state/timeout in seconds
- Added a get_remaining_time method to the FreeRTOSTimer class to know the time in ms before the timer expires.
- Integrated the new CAN messages into their corresponding Hepa Task and UV Task
- Refactored the Hepa Task logic so it takes into account the CAN messages so we aren't duplicating code.
- Refactored the UV Task logic so it takes into account the CAN messages so we aren't duplicating code.

# Review Requests

- Does anything not make sense with the can messages?

# Risk Assessment

Low, unreleased

[RET-1422]: https://opentrons.atlassian.net/browse/RET-1422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RET-1423]: https://opentrons.atlassian.net/browse/RET-1423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cpp-utils/include/ot_utils/freertos/freertos_timer.hpp
+++ b/cpp-utils/include/ot_utils/freertos/freertos_timer.hpp
@@ -6,8 +6,13 @@
 #include "task.h"
 #include "timers.h"
 
-#define ROUNDED_DIV(A, B) (((A) + ((B) / 2)) / (B))
-#define TICKS_TO_MS(TICKS) (uint32_t)ROUNDED_DIV((TICKS) * 1000, configTICK_RATE_HZ)
+constexpr auto ROUNDED_DIV(uint32_t A, uint32_t B) noexcept -> uint32_t {
+    return (((A) + ((B) / 2)) / (B));
+}
+
+constexpr auto TICKS_TO_MS(uint32_t ticks) noexcept -> uint32_t {
+    return (uint32_t)ROUNDED_DIV((ticks) * 1000, (uint32_t)configTICK_RATE_HZ);
+}
 
 namespace ot_utils {
 namespace freertos_timer {
@@ -57,11 +62,11 @@ class FreeRTOSTimer {
         }
     }
 
-    uint32_t get_remaining_time() {
+    auto get_remaining_time() -> uint32_t {
         /*
         The time in ms remaining before this timer expires and the callback is executed.
         */
-        if (!is_running()) return uint32_t(0);
+        if (!is_running()) { return uint32_t(0); }
         return TICKS_TO_MS(xTimerGetExpiryTime( timer ) - xTaskGetTickCount());
     }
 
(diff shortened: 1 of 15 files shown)


cpp-utils/include/ot_utils/freertos/freertos_timer.hpp
```diff
@@ -6,6 +6,9 @@
 #include "task.h"
 #include "timers.h"
 
+#define ROUNDED_DIV(A, B) (((A) + ((B) / 2)) / (B))
+#define PASSED_MS(TICKS) (uint32_t)ROUNDED_DIV((TICKS) * 1000, configTICK_RATE_HZ)
```
sfoster1, line 10: can we name this like `TICKS_TO_MS` or something, it's a little unclear just from the name what it does.

sfoster1 (approved):
Looks good, but let's change the name of that macro. Also, I think this functionality is getting to the point where it should really have tests.

vegano1:
> Looks good, but let's change the name of that macro. Also, I think this functionality is getting to the point where it should really have tests.

agreed, I will add a test suite in up coming PR.

caila-marashaj (approved):
Awesome!

PR #754, opened 2024-02-09
feat(motor-control): use motor status message to get motor enabled

ahiuchingau:
Make use of our CAN message `get_status_request` to get whether or not the motor is enabled.


sfoster1 (approved):
Okay, sure!

PR #755, opened 2024-02-21
fix 2160 motor driver register usage

pmoegenburg:
This corrects the 2160 motor driver register we use with the `update_glob_scaler` method.


pmoegenburg:
Confirmed this method isn't being used anywhere.

PR #757, opened 2024-03-04
fix(z-motors): should disegange Z motors at boot when motor controller task first starts

ahiuchingau:
We've been mistakenly engaging the z motors during start up (this supposedly disables the ebrake and uses the start the motor interrupt handler, which is the exact opposite of what we did in hardware gpio init).

This PR makes sure the z motors stay disengaged until the host tells it to move later on.

Closes RQA-2301, RQA-2429


sfoster1 (approved):
Looks good to me!

PR #758, opened 2024-03-18
feat(hepa-uv): add error code for door & reed state change during uv task

ahiuchingau:
This PR does two things:
1. Add error code for door & reed state change so the hepa uv can notify the server when the uv task gets interrupted by the hardware states
2. Only stops turns off the uv light if the hardware state changes if the uv task is running


vegano1 (approved):
This makes sense to me,
thank you.

vegano1 (approved):

PR #760, opened 2024-03-19
fix(sensors): improvments to compile time defines for pressure move

ryanthecoder:
Some of the files weren't getting the compile flag at the right time, and then use a new per-pipette define to handle different static array sizes


sfoster1 (approved):
Looks good, it might be nice to define it as 0 or 1 and that way it can always be defined but this works

PR #762, opened 2024-03-20
feat(hepa-uv): add current sensing for the uv lights.

vegano1:
# Overview

There is a sensor to measure the voltage of the UV light, this PR sets up the GPIO/ADC and lets us query the voltage on the UV_SNS_MCU pin, which we then use to calculate the current at the Rsns resister (DC Ballast). This PR works in conjunction with this [Opentrons/opentrons#14710](https://github.com/Opentrons/opentrons/pull/14710) pull request.

Closes: [PLAT-16](https://opentrons.atlassian.net/browse/PLAT-16)

# Test Plan

- [x] Make sure we get the correct current when the UV lights are ON
- [x] Make sure we get the correct current when the UV lights are ON
- [x] Make sure we get the correct current when one or more light bulbs are damaged/missing.

# Change Log

- Add plumbing to enable querying the UV current
- Adds a `uv_current_ma` Uint16 field to the `GetHepaUVStateResponse`

# Review Requests

# Risk Assessment

low, unreleased

[PLAT-16]: https://opentrons.atlassian.net/browse/PLAT-16?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ahiuchingau (approved):
Makes sense to me

PR #764, opened 2024-03-28
feat(hepa-uv): Ret 1406 add hepa tachometer

vegano1:
## Overview

The HEPA fan has a frequency generator that serves as a tachometer allowing us to read the rpm and determine if the fan is spinning. This PR adds the plumbing and exposes the tachometer rpm so the HEPA task can use it.

Closes: [PLAT-17](https://opentrons.atlassian.net/browse/PLAT-17)

## Change log

- Adds the timer for the HEPA_FG tachometer and puts it in input capture mode
- Calculate the rpm based on the tachometer period which we get every 4 rising pulses.
- Exposes the rpm to the HEPA task so it can be used in the future

# Test Plan

- [ ] Make sure we get the correct RPM when the Hepa fan is ON
- [ ] Make sure the RPM is 0 when the Hepa fan is OFF

# Review requests

# Risk Assessment

Low, unreleased


ahiuchingau (approved):
LGTM!

PR #765, opened 2024-04-04
fix(can): hook up the tertiary flag through the system message handler task

ryanthecoder:
The tertiary flag was unused before but now we need it to differentiate special pipettes that are compiled to use the on-device buffer for sensor data. The constructors did not pass this information through to the SystemMessageHandler before,

specially compiled pipettes now show up like this

`2024-04-04 16:37:55.195538 device_info_response (pipette_left->host)
        message_index: UInt32Field(value=1)
        version: UInt32Field(value=47)
        flags: VersionFlagsField(value=)
        shortsha: FirmwareShortSHADataField(value=b'75017537')
        revision: OptionalRevisionField(primary=d, secondary=1, tertiary=1)
        subidentifier: UInt8Field(value=1)`


sfoster1 (approved):
🐢

PR #766, opened 2024-04-08
chore(cmake): Add sensor variant firmware ci

ryanthecoder:
This PR adds a new configure preset and build preset that builds the firmware with the "USE_PRESSURE_MOVE" define
This enables the special firmware that allows the system to buffer sensor data on the pipettes, and sync sensor tasks with movements.

The workflow file builds and uploads just the applications so that internal users can just pull the .zip file of them from CI instead of building them


pmoegenburg (commented):
lgtm!

caila-marashaj (approved):
Nice!

PR #767, opened 2024-04-08
fix(hepa): fix a linter bug that makes CI complain

ryanthecoder:
Noticed that my CI builds were failing from this, and I want those sweet sweet ✅


vegano1 (approved):
Not sure, how I missed this.
Thank you!

PR #768, opened 2024-04-10
feat(sensors): enable capacitive sensor logging

pmoegenburg:
This PR, along with [../opentrons/ #14785](https://github.com/Opentrons/opentrons/pull/14785), adds csv file logging capability to the capacitive_probe process for the pipettes and the gripper. This work utilizes the work that adds the same capability to the liquid_probe process for the pipettes. The AddSensorLinearMoveRequest and SendAccumulatedSensorDataRequest messages were modified to include a sensor_type data field. The USE_PRESSURE_MOVE flag was modified to USE_SENSOR_MOVE.

 - [x] Ran script without csv logging and detected saltwater liquid height
 - [x] Ran script with csv loffing and detected saltwater liquid height


gripper/core/tasks.cpp
```diff
@@ -155,13 +164,27 @@ void gripper_tasks::QueueClient::send_capacitive_sensor_queue_rear(
...
+
 // gripper does not have environment nor pressure sensor
 void gripper_tasks::QueueClient::send_environment_sensor_queue(
     const sensors::utils::TaskMessage&) {}
 void gripper_tasks::QueueClient::send_pressure_sensor_queue_front(
     const sensors::utils::TaskMessage&) {}
 void gripper_tasks::QueueClient::send_pressure_sensor_queue_rear(
     const sensors::utils::TaskMessage&) {}
+void gripper_tasks::QueueClient::send_pressure_sensor_queue_front_isr(
+    const sensors::utils::TaskMessage&) {}
```
ryanthecoder, line 185: There is no pressure sensor on the gripper so if we need this then we have made a mistake somewhere else

gripper/core/CMakeLists.txt
```diff
@@ -1,4 +1,9 @@
 function(target_gripper_core TARGET)
+    if(${USE_SENSOR_MOVE})
+    target_compile_definitions(${TARGET} PUBLIC USE_SENSOR_MOVE)
+    target_compile_definitions(${TARGET} PUBLIC P_BUFF_SIZE=1700)
```
ryanthecoder, line 4: lets not do this and remove all of the instances of USE_SENSOR_MOVE from the gripper code, if we never use it then we don't need to add the compile time changes for it

include/common/core/sensor_buffer.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#if defined(USE_SENSOR_MOVE)
```
ryanthecoder, line 3: this should be #ifdef for consistancy

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
```diff
@@ -651,18 +728,30 @@ class MotorInterruptHandler {
         hardware.set_step_tracker(
             static_cast<uint32_t>(position_tracker >> 31));
     }
-#ifdef USE_PRESSURE_MOVE
+#ifdef USE_SENSOR_MOVE
     void send_to_pressure_sensor_queue_rear(
         can::messages::BindSensorOutputRequest& m) {
-        std::ignore = sensor_tasks::get_queues()
-                          .pressure_sensor_queue_rear->try_write_isr(m);
+        SensorClientHelper<SensorClient>::send_to_pressure_sensor_queue_rear(
```
ryanthecoder, line 734: Instead of doing all of this templated stuff we should just use the same get queues() method, it makes the code much simpler

gripper/core/CMakeLists.txt
pmoegenburg: Seth said it was ok to keep gripper capacitive data logging ability in firmware for future use, hidden behind the USE_SENSOR_MOVE flag

gripper/core/tasks.cpp
pmoegenburg: Pre-existing/known issue, see line 177 comment

include/motor-control/core/stepper_motor/motor_interrupt_handler.hpp
pmoegenburg: Necessary because not every subsystem that requires MotorInterruptHandler has a SensorClient

ryanthecoder: We don't need to since they're all behind compiler defines. I made a quick commit that rips it out

PR #770, opened 2024-04-17
fix(pipettes): Assign front capacitive sensor task to s1

caila-marashaj:
In `sensor_tasks`, both the front and rear capacitive sensor task builders were assigned to sensor id S0. This fixes it so the rear one will be s0, and the front s1.


sfoster1 (approved):
LGTM!

PR #772, opened 2024-04-22
fix(motion-control): motor driver error surfacing

pmoegenburg:
Fixes (motor driver false positive) error seen with 96-channel pipette. To test, successfully calibrated 96-channel pipette as well as gripper and then homed.


ahiuchingau (approved):
NOICE

PR #773, opened 2024-04-23
fix(motor-control): move enable motor motion controller task

ahiuchingau:
This seems to have fixed the motor enable/ebake not disengaging issue now that we have call the enable motor on the task level...


sfoster1 (approved):
Looks good to me once tested, though I wonder if we should have a `force` argument or something since right now there's no way to enable the motor that avoids the `enabled` check

ahiuchingau:
This issue is fixed by other PRs.

PR #774, opened 2024-04-26
fix(push): fix the push script after breaking it with CI changes

ryanthecoder:
I broke the changes to the push script when I updated some of the things for the CI build variant


sfoster1 (approved):

vegano1 (approved):
nice, I was about to fix this
Thank you!

PR #775, opened 2024-05-03
feat(push): add an option to specify targets

caila-marashaj:
## Overview
These changes to push allow us to specify targets in the form of presets, and only those targets will be built and copied onto the robot. To allow this, a new `update_shortsha` function grabs the existing manifest file from the robot, and updates the sha of the specified targets.

You can specify just one or multiple targets, for example:
`./push 10.10.10.10 --targets pipettes-single pipettes-multi gantry-x`.
`./push 10.10.10.10 --targets pipettes-single`

If this argument isn't used, the script will run the same as it did before, meaning it will build and update all firmware presets.

Logically, it behaves independently of the `sensors` argument, meaning you can use the `firmware-g4-sensors` build preset, and either send the full build or just the hex files for whichever nodes.

## Changelog

- add targets parameter
- add a function to grab and update the robot's manifest file with new sha
- only build and send desired firmware if given targets

## Test Plan

For all nodes:
  - [x] Make sure sha only changes for specified targets using `can_control`
  - [x] Check that the manifest file is updated as expected
  - [x] Make sure `sensors` argument still works with and without targets
  - [x] Make sure everything works the same if `targets` isn't used

## Review Requests

- Is there any particular order where I should call `update_shortsha` different from where I am?
- Are there any ways this might break anything that I should test?

### DIFF
--- a/push
+++ b/push
@@ -23,6 +23,21 @@ _DEFAULT_EXTRAS = {'stdout': sys.stdout, 'stderr': sys.stderr}
 _SSH_EXTRA_OPTS = ['-o', 'StrictHostKeyChecking=no',
                    '-o', 'UserKnownHostsFile=/dev/null']
 _ROBOT_MANIFEST_FILE_PATH = "/usr/lib/firmware/opentrons-firmware.json"
+TARGETS = [
+    "pipettes",
+    "pipettes-rev1",
+    "pipettes-single",
+    "pipettes-multi",
+    "pipettes-96",
+    "gripper",
+    "hepa-uv",
+    "gantry",
+    "gantry-x",
+    "gantry-y",
+    "head",
+    "rear-panel",
+    "bootloader",
+]
 
 class CantFindUtilityException(RuntimeError):
     def __init__(self, which_util):
@@ -145,12 +160,21 @@ def _find_utils():
         raise CantFindUtilityException('cmake')
     return ssh, scp, cmake
 
+def _check_targets(targets):
+    for t in targets:
+        if t not in TARGETS:
+            print(f"preset {t} is not in target options, ignoring")
+            targets.remove(t)
+    return targets
+
 def _restart_robot(host, ssh):
     _ssh(ssh, host, 'nohup systemctl restart opentrons-robot-server &')
 
 def _do_push(host, repo_path, build, restart, sensors, targets):
 
     ssh, scp, cmake = _find_utils()
+    if targets:
+        targets = _check_targets(targets)
     if build:
         _prep_firmware(repo_path, cmake, sensors, targets)
     with _prep_robot(host, ssh):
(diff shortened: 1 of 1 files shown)


push
```diff
@@ -169,6 +216,10 @@ def _arg_parser(parent=None):
         action='store_true',
         help='Private SSH key to use'
     )
+    parser.add_argument(
+        '--targets',
+        nargs='+'
```
sfoster1, line 221: this means you _have to_ provide targets, right? maybe `*` is better?

sfoster1 (changes requested):
Looks good but I do want to make sure that you can still use this without these args and get the old behavior

push
```diff
@@ -54,31 +64,63 @@ def _controlled_tempdir():
...
 
-def _build_fw(zip_path, apps_path):
-    with ZipFile(zip_path, 'w') as zf:
-        for fname in os.listdir(apps_path):
-            zf.write(os.path.join(apps_path, fname), fname)
-
-def _transfer_firmware(host, repo_path, scp, ssh, sensors):
+def _build_fw(zip_path, apps_path, targets):
+    if targets:
+        regex_list = [re.compile(f"{target}" + r"(.*).hex") for target in targets]
```
vegano1, line 69: The rear panel produces `.bin` files, you'll need to include those as well.

push
```diff
@@ -40,6 +43,13 @@ def _scp_to_robot(scp_util, host, local, remote, **extras):
         **extras
     )
 
+def _scp_from_robot(scp_util, host, local, remote, **extras):
+    _cmd(
+        [scp_util]
+        + _SSH_EXTRA_OPTS
+        + ['root@{host}:{remote}'.format(host=host, remote=remote), local],
```
vegano1, line 50: could also use f-strings

f"root@{host}:{remote}"

push
vegano1: We should define a list of targets and validate them against the user input, something like this.

```
TARGETS = [
    "head",
    "gantry-x",
    "gantry-y",
    "gripper",
    "pipettes-single",
    "pipettes-multi",
    "pipettes-96",
    "hepa-uv",
    "rear-panel",
]

for target in args.targets:
    if target not in TARGETS:
        print(f"Unknown target {target}", file=sys.stderr)
        continue

    # do work here
```

We actually do something similar in the `scripts/subsystem_versions.py` script

push
```diff
@@ -54,31 +64,63 @@ def _controlled_tempdir():
...
-    with ZipFile(zip_path, 'w') as zf:
-        for fname in os.listdir(apps_path):
-            zf.write(os.path.join(apps_path, fname), fname)
-
-def _transfer_firmware(host, repo_path, scp, ssh, sensors):
+def _build_fw(zip_path, apps_path, targets):
+    if targets:
+        regex_list = [re.compile(f"{target}" + r"(.*).hex") for target in targets]
+        with ZipFile(zip_path, 'w') as zf:
+            for fname in os.listdir(apps_path):
```
vegano1, line 71: Same as the above comment, the `scripts/subsystem_versions.py` script validates the filenames, it could be useful to check that.

vegano1 (approved):
Changes look good to me, I left a few comments.
We should validate the `--target` args against a known list of targets.
Other than that the change works, I tested it locally!

PR #776, opened 2024-05-07
fix(sensors): don't send errors when queue is full

ryanthecoder:
When we send a bunch of errors because the queue is full it can cause the can bus to choke up with stop requests which are annoying and make the logs hard to read.

The new LLD script that uses this custom firmware now limits motion based on factors that include the buffer size of the pipettes so we can just get rid of this.


sfoster1 (approved):
yes

PR #777, opened 2024-05-10
fix(motor-control): Revert motor driver chip errors

ryanthecoder:
Something in here is stall/collision detection errors and/or pipette firmware crashes in release testing


sfoster1 (approved):
so it goes

vegano1:
oof

PR #778, opened 2024-05-10
fix(sensors): revert measure mode rate change

ryanthecoder:
we increased the measure mode rate change for the LLD testing but it seems to cause false overpressure errors so this PR gates using the different rates behind the compile time flag that we use to build the LLD variant firmware


sfoster1 (approved):
cool cool

PR #780, opened 2024-05-21
fix(motor control): Use FreeRTOS delay instead of hardware delay

caila-marashaj:
We recently started adding a delay between engaging a head motor and starting the execution of a move. We were using a HAL delay to do this, and it turns out that this blocks FreeRTOS's task handling, and in certain edge cases causes a firmware crash. 

We should instead use FreeRTOS's built-in `vTaskDelay`


sfoster1 (approved):
excellent

SyntaxColoring:
Whoa, nice find.

How did this cause a firmware crash? Some kind of watchdog timeout?

caila-marashaj:
> Whoa, nice find.
> 
> How did this cause a firmware crash? Some kind of watchdog timeout?

It seems like a memory fault after the delay blocked the motor interrupt a bunch of times.

PR #781, opened 2024-05-21
fix(gripper): capacitive sensor task should use correct sensor ID

ahiuchingau:
The S1 sensor has been falsely using the wrong ID while responding to CAN messages. 

This is because the sensor ID for the driver is defaulted to S0, and we rely on `set_sensor_id()` to update the sensor. We've made these changes for the pipettes (https://github.com/Opentrons/ot3-firmware/pull/630). And since `shared_sensor` is always false for the gripper, we've never set the sensor ID for S1.


ryanthecoder (approved):
lgtm

caila-marashaj (approved):
Nice

PR #783, opened 2024-06-03
fix(sensors): send the pressure sensor response with the baseline applied

ryanthecoder:
Right now we send the raw sensor value instead of the corrected one, it will be better for lld development and just in general to get the corrected pressure data instead


sfoster1 (approved):
Okay, cool

pmoegenburg (approved):

PR #784, opened 2024-06-12
feat(pipettes): send most recent sensor data after probe

caila-marashaj:
## Overview
After a liquid probe, we want to send the 500 most recent pressure data readings in a buffer over CAN to determine on the Python side if we've successfully hit liquid, hit the bottom of a piece of labware, or not found liquid at all. 

This pr creates a circular buffer of pressure sensor data, meaning that after every i2c transaction, we'll write the pressure reading to the buffer. When we reach the end of the buffer, we'll circle back to the front of the buffer and continue overwriting the oldest data until the probing is finished. Then, after receiving a `SendAccumulatedSensorDataRequest`, we send the buffer in order starting from the oldest data.

## Changelog

- create a sensor buffer regardless of compile target
- change `p_buff` in all sensor firmware to `sensor_buffer`, since the capacitive sensor now uses this too
- change the buffer size for all pipettes to 500
- create a new function in `pressure_driver` to manage the buffer
- change the way we send the data buffer to accommodate the new shape of the data

## Test Plan
for single, multi, and 96 channel pipettes:

- [x] run liquid probe and make sure that still works the same
- [x] afterward, using `can_control`, send a `SendAccumulatedSensorDataRequest`
- [x] make sure 500 `read_sensor_response` messages are sent over CAN
- [x] make sure that the readings in those messages align with the sensor responses sent over CAN during the probe

## Todo
I plan on adding a test for this in a follow-up pr


include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -280,26 +289,24 @@ class MMR920 {
...
 
     void send_accumulated_sensor_data(uint32_t message_index) {
-#ifdef USE_SENSOR_MOVE
-        for (int i = 0; i < sensor_buffer_index; i++) {
-            // send over buffer adn then clear buffer values
+        for (size_t i = 0; i < SENSOR_BUFFER_SIZE; i++) {
+            // send over buffer and then clear buffer values
+            int current_index = (i + sensor_buffer_index) % SENSOR_BUFFER_SIZE;
+
             can_client.send_can_message(
```
ryanthecoder, line 296: lets keep the #ifdef around the send_can_message part of this for now and just move the buffer storage out. we don't want to flood the can bus while we're probing and we will probably remove sending this can message in real time in the future.

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -323,6 +330,7 @@ class MMR920 {
             _registers.pressure_result.reading, sensor_version);
 
         if (max_pressure_sync) {
+            sensor_buffer_log(pressure);
```
ryanthecoder, line 333: This doesn't go inside this if block, it should be the `if(echo_this_time) ` block otherwise we are double logging the data
and we can remove the ifdef from the` if(echo_this_time)` block

sfoster1 (approved):
Looks good once tested, though I'm curious about the use of `(*buff).at()` - why not either `buff->at()` or `(*buff)[]`? the latter is probably faster since it doesn't range check; this would normally be bad but we can't handle exceptions anyway.

PR #785, opened 2024-06-18
circle buffer rough fix

caila-marashaj:


include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -371,6 +395,8 @@ class MMR920 {
         if (echo_this_time) {
             auto response_pressure = pressure - current_pressure_baseline_pa;
             // do we want pressure or response pressure
+            if (pressure_transations < 500) { pressure_transactions++; }
```
ryanthecoder, line 398: 500 should be SENSOR_BUFFER_SIZE instead

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -290,25 +291,48 @@ class MMR920 {
     }
 
     void send_accumulated_sensor_data(uint32_t message_index) {
-        for (int i = 0; i < static_cast<int>(SENSOR_BUFFER_SIZE); i++) {
-            // send over buffer and then clear buffer values
-            // NOLINTNEXTLINE(div-by-zero)
-            int current_index = (i + sensor_buffer_index) %
-                                static_cast<int>(SENSOR_BUFFER_SIZE);
+        if (pressure_transations >= 500) {
```
ryanthecoder, line 294: 500 should be SENSOR_BUFFER_SIZE

sfoster1 (changes requested):
This logic is kind of weird and fiddly. We should make it something that can be more easily tested. I'd suggest putting the indexing logic in its own file and class (like `DoubleBuffer`) with some methods to do various things and like a function called `get_read_index()` for getting data out (really, I'd suggest using a custom iterator, but that's kind of a pain). Then, we need to test the hell out of it.

PR #786, opened 2024-06-21
feat(sensors): Self baselining during sensor moves

ryanthecoder:
We baseline the sensor reading during moves so that we can compensate better for different tips having a different baseline pressure differential when the plunger is moving

### DIFF
--- a/include/sensors/core/tasks/pressure_driver.hpp
+++ b/include/sensors/core/tasks/pressure_driver.hpp
@@ -378,8 +378,8 @@ class MMR920 {
 
         if (echo_this_time) {
             auto response_pressure = pressure - current_pressure_baseline_pa;
-            // do we want pressure or response pressure
-            sensor_buffer_log(pressure);
+
+            sensor_buffer_log(response_pressure);
 
             if (sensor_buffer_index == 10) {
                 current_pressure_baseline_pa =
--- a/include/sensors/core/tasks/pressure_driver.hpp
+++ b/include/sensors/core/tasks/pressure_driver.hpp
@@ -364,6 +364,8 @@ class MMR920 {
             if (std::fabs(pressure - current_pressure_baseline_pa) >
                 threshold_pascals) {
                 hardware.set_sync();
+                // force this to stay set long enough to debounce on other nodes
+                vTaskDelay(10);
             } else {
                 hardware.reset_sync();
             }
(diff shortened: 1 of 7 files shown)


include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -290,10 +293,21 @@ class MMR920 {
     }
 
     void send_accumulated_sensor_data(uint32_t message_index) {
```
sfoster1, line 295: does this overlap with #785 ?

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -355,31 +372,42 @@ class MMR920 {
...
-                can::messages::ReadFromSensorResponse{
-                    .message_index = m.message_index,
-                    .sensor = can::ids::SensorType::pressure,
-                    .sensor_id = sensor_id,
-                    .sensor_data =
-                        mmr920::reading_to_fixed_point(response_pressure)});
+
+            sensor_buffer_log(response_pressure);
+
+            if (sensor_buffer_index == 10 && !crossed_buffer_index) {
```
sfoster1, line 399: this 10 is setting the baselining window size, right? let's drop that in a define or a constexpr or something

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -355,31 +372,42 @@ class MMR920 {
...
+
+                current_pressure_baseline_pa =
```
sfoster1, line 401: why are we incrementing instead of setting this? won't this make it run away? also we could put `10*current_pressure_baseline_pa` in the init argument to `accumulate` if we do want the accumulation

sfoster1 (changes requested):
Would love to see some constant definitions, and I'm a bit confused about the 10 sample windowing

include/sensors/core/tasks/pressure_driver.hpp
ryanthecoder: I'll add a comment that explains why this won't run away

ryanthecoder: re-worked it to make it more clear

sfoster1 (approved):
Okay, looks good to me. We may want to gate autobaselining behind a separate command, but we can do that in a different PR if you want to merge this one.

PR #787, opened 2024-06-25
refactor(sensors): rename shared-sensor to using-both-sensors

caila-marashaj:
## Overview
There's a class variable for the capacitive sensor driver called `shared_sensor`, that is used to keep track of when both of the sensors are in use on an 8 or 96 channel pipette. I don't think the name alone communicates what it's for, so here's my suggestion for renaming it.


sfoster1 (approved):
Looks good, but you'll have to run a format

PR #789, opened 2024-07-01
feat(sensors): Enable dual sensor logic for 8/96 channel

ryanthecoder:
Hardware testing wants the ability to do all of the following logic commands for LLD on the 8/96:
A only
B only
A or B
A and B

This lets the python side decide which of these to use.


sfoster1 (approved):
Looks good to me!

PR #790, opened 2024-07-02
refactor(sensors): remove USE_SENSOR_FIRMWARE compile target

caila-marashaj:
## Overview
We want to keep the `SENSOR_MOVE` option, but get rid of sensor firmware build variants. We also want to make what was previously the `USE_SENSOR_MOVE` if-define in the sensor drivers the default option, so we can get rid of _some_ of the `ifdef`s for now.
     
For the pressure sensor, `reporting` will no longer happen continuously over the CAN bus as long as the `report` binding is set. It will use the circular buffer which has a default size of 500, and send over all of the data that buffer has only in response to a `SendAccumulatedSensorDataRequest`.

The capacitive sensor, for now, will keep the old `reporting` functionality, which is reporting continuously as long as the `report` bit is set, as well as have a circular buffer and populate that if `report` is true.

In the future, we'll want to probably have more parity between the sensors' data repoting  also pull out the "circle buffer" functionality into its own class.

## Changelog

- get rid of `-sensor` firmware variants
- remove the `sensors` option from the `push` script
- remove sensor-preset related github workflows
- make the capacitive sensor's logging a little more like the pressure sensor's to temporarily make up for these other changes


sfoster1 (approved):
cool cool. nice work!

pipettes/core/CMakeLists.txt
```diff
@@ -15,6 +13,7 @@ function(target_pipettes_core_common TARGET REVISION)
 endfunction()
 
 function(target_pipettes_core_single TARGET REVISION)
+    target_compile_definitions(${TARGET} PUBLIC USE_SENSOR_MOVE)
```
ryanthecoder, line 16: I don't think we need this here, I think the def in target_pipettes_core_common will be sufficient

PR #791, opened 2024-07-10
fix(can): parse the binding flag and pass to sensor move message

ryanthecoder:
We added it on the python side without doing all the parsing correctly in the c++ code


sfoster1 (approved):
Cool, good fix

PR #793, opened 2024-07-12
fix(sensors): correctly correct the past sensor readings

ryanthecoder:
Silly mistake, was essentially doing nothing before


sfoster1 (approved):
lol

PR #794, opened 2024-07-17
chore(sensors): Reduce sensor buffer

ryanthecoder:
Buffer was too large and was causing annoying delays when receiving over can.
Also the deluge of data would sometime choke the can bus


sfoster1 (approved):
Nice!

PR #797, opened 2024-07-23
feat(sensors): support new pipette sensor board

ryanthecoder:
The new sensor board adds a GPIO expander so that we can read the version of the board and adjust how we read the pressure sensor.

The way this works is that there is a SensorHardwareVersionSingleton that is in charge of holding the version of the board, and it defaults to the version before we had the ability to see the sensor board rev. This singleton is then shared via the SensorHardwareBase instance(s) and a new ReadBoardRevTask

The new task attempts to query the i2c bus at the address of our expander, if it receives a response it compares the bitmap to known versions and sets the version of the singleton.

Now any sensor that gets updated, can change their behavior by asking their SensorHardware instance what rev to use instead of relying on compiler defines.

### DIFF
--- a/include/sensors/core/tasks/read_sensor_board_rev_task.hpp
+++ b/include/sensors/core/tasks/read_sensor_board_rev_task.hpp
@@ -12,11 +12,14 @@ using TaskMessage =
 template <class I2CQueueWriter, class OwnQueue>
 class ReadSenorBoardHandler {
   public:
-    explicit ReadSenorBoardHandler(I2CQueueWriter &i2c_writer, OwnQueue &own_queue)
+    explicit ReadSenorBoardHandler(I2CQueueWriter &i2c_writer,
+                                   OwnQueue &own_queue)
         : i2c_writer{i2c_writer}, task_queue{own_queue} {
         // request gpio expander status register
-        i2c_writer->read(pie4ioe4::ADDRESS, static_cast<uint8_t>(pie4ioe4::registers::input_status), 1,
-                        task_queue);
+        i2c_writer->read(
+            pie4ioe4::ADDRESS,
+            static_cast<uint8_t>(pie4ioe4::registers::input_status), 1,
+            task_queue);
     }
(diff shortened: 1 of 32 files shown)


include/sensors/core/tasks/read_sensor_board_rev_task.hpp
```diff
@@ -0,0 +1,93 @@
...
+    auto operator=(const ReadSenorBoardTask &&c) = delete;
+    ~ReadSenorBoardTask() = default;
+
+    /**
+     * Task entry point.
+     */
+    void operator()(
+        i2c::writer::Writer<QueueImpl> *writer,
+        sensors::hardware::SensorHardwareVersionSingleton *version_wrapper) {
+        auto handler = ReadSenorBoardHandler(writer, get_queue());
```
sfoster1, line 76:
```suggestion
        auto handler = ReadSensorBoardHandler(writer, get_queue());
```

include/sensors/core/tasks/read_sensor_board_rev_task.hpp
```diff
@@ -0,0 +1,93 @@
...
+    I2CQueueWriter &i2c_writer;
+    OwnQueue &task_queue;
+};
+
+/**
+ * The task type.
+ */
+template <template <class> class QueueImpl>
+requires MessageQueue<QueueImpl<TaskMessage>, TaskMessage>
+class ReadSenorBoardTask {
```
sfoster1, line 59:
```suggestion
class ReadSensorBoardTask {
```
etc

sfoster1 (changes requested):
It looks good but there's a typo in the new class... it'll bother me forever if we don't change it.

Also can we add some way for the pipette to tell the host which sensor board it's got?

ryanthecoder:
Tested Side by side with an old pipette and performing as expected

sfoster1 (approved):
Nice!

PR #799, opened 2024-08-02
fix(sensors): Share sync line control bitmasks

ryanthecoder:
I had been mistaken about the sensor hardware being shared instances but I see now that they were not, so I broke out the bitmask communication between sensor tasks into another singleton that the sensor hardware tasks can share.  
I also updated the tests so that when we test with multiple sensors they actually use two different hardware instances.

### DIFF
--- a/include/sensors/core/sensor_hardware_interface.hpp
+++ b/include/sensors/core/sensor_hardware_interface.hpp
@@ -59,8 +59,6 @@ class SensorHardwareSyncControlSingleton {
     [[nodiscard]] auto mask_satisfied() const -> bool {
         if (set_sync_required_mask !=
             static_cast<uint8_t>(SensorIdBitMask::UNUSED)) {
-            printf("%x is required\n", set_sync_required_mask);
-            printf("%x sync_state_mask\n", sync_state_mask);
             // if anything is "required" only sync when they are all triggered
             return (sync_state_mask & set_sync_required_mask) ==
                    set_sync_required_mask;
@@ -70,7 +68,6 @@ class SensorHardwareSyncControlSingleton {
 
     auto set_sync(can::ids::SensorId sensor) -> void {
         // force the bit for this sensor to 1
-        printf("setting sync on %x sensor\n", get_mask_from_id(sensor));
         sync_state_mask |= get_mask_from_id(sensor);
     }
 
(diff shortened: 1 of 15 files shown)


include/sensors/core/sensor_hardware_interface.hpp
```diff
@@ -43,6 +43,70 @@ static auto get_mask_from_id(can::ids::SensorId sensor) -> uint8_t {
...
+        -> SensorHardwareSyncControlSingleton& = default;
+    SensorHardwareSyncControlSingleton(SensorHardwareSyncControlSingleton&&) =
+        default;
+    auto operator=(SensorHardwareSyncControlSingleton&&)
+        -> SensorHardwareSyncControlSingleton& = default;
+
+    [[nodiscard]] auto mask_satisfied() const -> bool {
+        if (set_sync_required_mask !=
+            static_cast<uint8_t>(SensorIdBitMask::UNUSED)) {
+            printf("%x is required\n", set_sync_required_mask);
```
sfoster1, line 62: mmm don't like these even they compile out

include/sensors/core/sensor_hardware_interface.hpp
```diff
@@ -43,6 +43,70 @@ static auto get_mask_from_id(can::ids::SensorId sensor) -> uint8_t {
...
+
+    auto set_sync(can::ids::SensorId sensor) -> void {
+        // force the bit for this sensor to 1
+        printf("setting sync on %x sensor\n", get_mask_from_id(sensor));
+        sync_state_mask |= get_mask_from_id(sensor);
+    }
+
+    auto reset_sync(can::ids::SensorId sensor) -> void {
+        // force the bit for this sensor to 0
+        sync_state_mask &= 0xFF ^ get_mask_from_id(sensor);
```
sfoster1, line 79:
```suggestion
        sync_state_mask &= ~get_mask_from_id(sensor);
```
?

sfoster1: it's the more common way I've seen, anyway

sfoster1 (changes requested):
Definitely want to get these prints out. Also wait why `&= 0xff ^ thing` rather than `&= ~thing`?

sfoster1 (approved):
Nice, thanks!

PR #802, opened 2024-08-20
fix(hepa-uv): fix issue when turning on UV after re-seating the hepa/uv on the flex.

vegano1:
# Overview

When the door is closed, the reed switch is set, and the UV push button is pressed, the UV light turns on as expected.
However, if you then lift the back of the Hepa/UV and unset the reed switch, then set it back down setting the reed switch again. When you press the UV push button the safety relay pin will NOT be active even though the door is closed, and reed is set. The only way to activate the safety relay pin again is to open, and close the Flex door, and THEN press the UV push button again. This is by design per the hardware team as a way to make sure the user understands the intention when turning ON the UV light again.

Closes: [RQA-2872](https://opentrons.atlassian.net/browse/RQA-2872)

# Changelog

- Check safety relay is not set before attempting to change the UV light state.

# Test Plan

- [x] Make sure that the UV light turns on when the door is closed, reed switch is set, and uv push button is pressed
- [x] Make sure that the UV light turns OFF and the push button LED is set to blue when the door is opened.
- [x] Make sure that the UV light turns OFF and the push button LED is set to blue when the reed switch is not set.
- [x] Make sure that after re-setting the reed switch that the UV light does not turn on and UV push button turns blue.
- [x] Make sure that after re-setting the reed switch that the UV light ONLY turns ON AFTER opening/closing the door.

[RQA-2872]: https://opentrons.atlassian.net/browse/RQA-2872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/include/hepa-uv/core/uv_task.hpp
+++ b/include/hepa-uv/core/uv_task.hpp
@@ -91,9 +91,9 @@ class UVMessageHandler {
         debounce_timer.start();
 
         // update states
+        update_safety_relay_state();
         door_closed = gpio::is_set(drive_pins.door_open);
         reed_switch_set = gpio::is_set(drive_pins.reed_switch);
-        update_safety_relay_state();
         if (m.pin == drive_pins.uv_push_button.pin)
             uv_push_button = !uv_push_button;
     }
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Bit convoluted, but okay!

ahiuchingau (approved):
Makes sense!

CaseyBatten (approved):
Change to the relay activation case makes sense, and this seems fairly non-invasive.

PR #804, opened 2024-09-23
fix(motor-control): delay after enabling motor and before disengaging brake

ahiuchingau:
The z-stages still drop sometimes when the motor first gets enabled (see [ticket](https://opentrons.atlassian.net/jira/servicedesk/projects/RESC/issues/RESC-328?jql=project%20%3D%20%22RESC%22%20ORDER%20BY%20created%20DESC)). I suspect the hold current hasn't had a chance to ramp up before the brake gets disengaged. This should fix most of the issues.

If motors are still dropping after this fix, we might want to consider temporarily increasing the hold current while the motor is being enabled to hold the motor in place.


sfoster1 (approved):
That looks good to me. I like the hold-while-enabled thing, too

PR #805, opened 2024-09-27
feat(sensors): add a new can message to batch read sensor data

ryanthecoder:
In a first step to streamline tool_sensors.py we want to batch read sensor data, this will speed up the process and let us stream a lot better

This PR removes much of the complexity from tool sensors and puts creates a new log specifically for sensor data.

### DIFF
--- a/include/sensors/core/tasks/pressure_driver.hpp
+++ b/include/sensors/core/tasks/pressure_driver.hpp
@@ -320,7 +320,7 @@ class MMR920 {
     }
 
     auto try_send_next_chunk(uint32_t message_index) -> void {
-        std::array<uint32_t, can::messages::BATCH_SENSOR_MAX_LEN> data{};
+        std::array<int32_t, can::messages::BATCH_SENSOR_MAX_LEN> data{};
         auto count = get_buffer_count();
         auto data_len = std::min(uint8_t(count),
                                  uint8_t(can::messages::BATCH_SENSOR_MAX_LEN));
--- a/include/sensors/core/tasks/pressure_driver.hpp
+++ b/include/sensors/core/tasks/pressure_driver.hpp
@@ -329,7 +329,8 @@ class MMR920 {
         }
         for (uint8_t i = 0; i < data_len; i++) {
             data.at(i) = mmr920::reading_to_fixed_point(
-                (*sensor_buffer).at(sensor_buffer_index_start + i));
+                (*sensor_buffer)
+                    .at((sensor_buffer_index_start + i) % SENSOR_BUFFER_SIZE));
         }
         auto response = can::messages::BatchReadFromSensorResponse{
             .message_index = message_index,
--- a/include/sensors/core/tasks/pressure_driver.hpp
+++ b/include/sensors/core/tasks/pressure_driver.hpp
@@ -71,7 +71,8 @@ class MMR920 {
     void set_echoing(bool should_echo) {
         echoing = should_echo;
         if (should_echo) {
-            sensor_buffer_index = 0;  // reset buffer index
+            sensor_buffer_index_start = 0;  // reset buffer index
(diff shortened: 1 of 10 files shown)


caila-marashaj (approved):
logic looks good to me!

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -315,24 +315,32 @@ class MMR920 {
...
             }
+
+            auto response = can::messages::BatchReadFromSensorResponse{
+                .message_index = message_index,
+                .sensor = can::ids::SensorType::pressure,
+                .sensor_id = sensor_id,
+                .data_length = j,
+                .sensor_data = data,
+            };
+            can_client.send_can_message(can::ids::NodeId::host, response);
```
sfoster1, line 342: We haven't faced it before as much, but I think we can improve this a lot by changing (or adding a new version of) `send_can_message` that will either (a) block (by sleeping) or (b) return `false` when the local buffer is full. That will let us dynamically slow down data output when the buffer is full while not adding a delay all the time.

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -315,24 +315,32 @@ class MMR920 {
...
-                    .message_index = message_index,
-                    .sensor = can::ids::SensorType::pressure,
-                    .sensor_id = sensor_id,
-                    .sensor_data = mmr920::reading_to_fixed_point(
-                        (*sensor_buffer).at(current_index))});
-            if (i % 10 == 0) {
-                // slow it down so the can buffer doesn't choke
-                vtask_hardware_delay(20);
+            uint8_t j = 0;
+            while (j < 14 and i < count) {
```
sfoster1, line 326: We absolutely need some tests for the fiddly index math.

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -315,24 +315,32 @@ class MMR920 {
         can_client.send_can_message(
             can::ids::NodeId::host,
             can::messages::Acknowledgment{.message_index = count});
-        for (int i = 0; i < count; i++) {
+        std::array<uint32_t, 14> data{};
+        int i = 0;
+        while (i < count) {
```
sfoster1, line 320: could make this `for(size_t i = 0; i < count; ) { }` if we wanted. in general, seems like we could use a size here instead of a signed value.

include/sensors/core/tasks/pressure_driver.hpp
```diff
@@ -315,24 +315,32 @@ class MMR920 {
         can_client.send_can_message(
             can::ids::NodeId::host,
             can::messages::Acknowledgment{.message_index = count});
-        for (int i = 0; i < count; i++) {
+        std::array<uint32_t, 14> data{};
```
sfoster1, line 318: would love to see this 14 derived from somewhere

include/sensors/core/tasks/pressure_driver.hpp
sfoster1: also can use `j < data.size()` or even take this to a range-for that stops early a la
```c++
for (auto& copy_to : data) {
    if (i >= count) {
        break;
    }
    copy_to = mmr920::reading_to_fixed_point(sensor_buffer->at(current_index));
    i++;
    current_index = (i + start) % static_cast<int>(SENSOR_BUFFER_SIZE);
}
```

sfoster1 (changes requested):
Couple small readability and robustness changes and also definitely want some tests for the index math and logic

ryanthecoder:
Marking as draft since Caila didn't get a chance to use this before her trip. I'll clean it up and make it better styled

sfoster1 (approved):
Nice, looks good to me!

PR #806, opened 2024-09-30
fix(hepa-uv): fix issue when turning on UV after re-seating the hepa

y3rsh:
# Merge back of release to main

Make sure any isolated changes, history, and tags that happened in the chore_release-8.0.0 branch are in main.


vegano1 (approved):
🚀

PR #810, opened 2024-10-23
fix(motor-control): catch falling 96 channel 

ryanthecoder:
So from the various logs I have found 2 different cases where the 96 channel can fall down to the deck

1. During the first home after a boot up, the 96 sometimes falls, it's a little unclear on why this happens, there could be some static friction condition that happens when things are powered off.

2. If there is a Stop Request sent while the 96 is moving faster than the discontinuity speed in the downward direction, the steps immediately stop without acceleration and the momentum of the 96 overcomes the head motor, creates a resonate condition and falls.

This PR makes it a bit better by "catching" the pipette when this happens. First of all when a "stop request" is sent we will automatically call disable motor from the motion controller. This will activate the brake if there is one on that axis.

Second, specifically to address the first cause, this PR adds a new way of detecting when a fall is happening during a home by seeing if the axis starts moving in the wrong direction and triggering a collision detected. Since the python side already spits out a stop request immediately upon receipt of a collision detected this will trigger the brake.

### DIFF
--- a/include/motor-control/core/stepper_motor/motion_controller.hpp
+++ b/include/motor-control/core/stepper_motor/motion_controller.hpp
@@ -195,6 +195,7 @@ class MotionController {
         if (hardware.is_timer_interrupt_running()) {
             hardware.set_cancel_request(error_severity, error_code);
         }
+        disable_motor();
     }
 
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
Ah, that's really nice. Nice work!

PR #811, opened 2024-10-24
fix(motor-control): Reimplement motor driver chip errors as poller

ryanthecoder:
The additional interrupts were causing too much of a slowdown on some nodes so this re-implements this as a polling based detection system instead of a interrupt based one.
RQA-3405


sfoster1 (approved):
Lots easier, nice!

vegano1 (approved):
negative diff, lets go!

PR #813, opened 2024-11-04
fix(motion-control): Don't clear stepper-ok in stop request

ryanthecoder:
We added the disable motor in the stop request because that was the cause of some falling pipettes and we needed to turn the brake on. however disable clears the stepper ok flag and this isn't needed in this case.

So the new behavior is that it calls hardware disable_motor to turn the brake on, then it calls activate motor after to turn it back off and leaves the motor controller's 'enabled' flag true so the motion controller task has the option to call motor position update request.


caila-marashaj (approved):

sfoster1 (approved):
Ah, that looks good to me!

ryanthecoder:
Tested, still catches the 96 and does not block update motor position estimation

PR #814, opened 2024-11-05
fix(motion-control): Don't clear stepper-ok in stop request (#813)

sfoster1:
* We don't need to clear the stepper ok field when we get a stop message.

* Ok i was wrong about how this worked. We don't want to leave the motor enable false we just want to turn the brake on for a bit then turn it back off to leave the motor enabled.

### DIFF
--- a/include/motor-control/core/stepper_motor/motion_controller.hpp
+++ b/include/motor-control/core/stepper_motor/motion_controller.hpp
@@ -189,7 +189,8 @@ class MotionController {
         if (hardware.is_timer_interrupt_running()) {
             hardware.request_cancel();
         }
-        disable_motor();
+        hardware.deactivate_motor();
+        hardware.activate_motor();
     }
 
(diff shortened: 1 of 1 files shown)


include/motor-control/core/stepper_motor/motion_controller.hpp
```diff
@@ -189,7 +189,8 @@ class MotionController {
         if (hardware.is_timer_interrupt_running()) {
             hardware.request_cancel();
         }
-        disable_motor();
+        hardware.deactivate_motor();
```
vegano1, line 192: do we need a delay here?

sfoster1: no, this is about setting logical state. also it's already been tested and merged into `main`, this is a cherry-pick basically

vegano1 (approved):

PR #815, opened 2024-11-07
fix how the batch message is seralized on the firmware

ryanthecoder:
Forgot to serialize "data_length" and then the python side was complaining when the end of the message wasn't padded because it was expecting to have a static length


sfoster1 (approved):
ah, rip. nice catch

PR #818, opened 2024-12-12
fix(motor-control): improve stop request handling

ryanthecoder:
So when a stop request is broadcast it was previously sending out 2-3 responses per motor axis. 1 from the move group controller, 1 from the motor controller and 1 from the motor interrupt handler if the motor has been previously enabled (which is always true except right after a firmware update or power cycle.)

This PR removes the response from the move group controller, since this message always responded with a message index 0 which is unhelpful and will reduce the number of messages by 1/3 and prevents many of the "Error 1008 CAN_BUS_ERROR: error frame encountered" instances


sfoster1 (approved):
Oh that's so much better

PR #821, opened 2025-03-18
fix(motor-control): improve brake delay time

ryanthecoder:
So we haven't seen much problems with our old 20ms delay, however the spec on the brakes is 50ms.

Recently there have been some bad batches from the supplier that take longer to engage/disengage so we are going to delay 100ms now to make sure we can accommodate any brake that's in-spec to slightly over spec,


sfoster1 (approved):
Okay, this is a good baseline but we may need to increase it further

PR #822, opened 2025-03-20
feat(eeprom): Store overpressure count

ryanthecoder:
Store lifetime overpressures in the eeprom, this will now report out when a GetMotorUsageRequest is sent to the plunger node.


sfoster1 (approved):
tested ok

PR #823, opened 2025-03-21
feat(eeprom): Add resin-tip dispense tracking

ryanthecoder:
Adds a new can message that the pipette uses to increment it's resin tip dispensing tracker in eeprom.

The new value is can be retrieved with a standard usage request.


sfoster1 (approved):
works now!

PR #824, opened 2025-05-08
chore(rear-panel): add revision D1 for stacker support

vegano1:
The rear panel D1 revision only has electrical/hardware changes that don't require firmware changes. However, we need new binaries to flash at the factory so we can tell which board is which.


sfoster1 (approved):
Sounds good to me though you might also want to make it default

PR #837, opened 2026-03-04
feat(wear leveling): Boundary Setting Mechanism

munanwugo-opentrons:
# Overview
The decision has been made to exclude all of the locations on the EEPROM that had been used before wear-leveling had been implemented. This PR includes the logic to find that edge and write it to a new "boundary" location in the EEPROM header. 

### `update_data_rev_handler.hpp`
There are now 2 new messages that update_data_rev handler can handle: MigrateDataMessage, OTLibraryUpdateMessage.
The latter is to be implemented at a later date. 

The MigrateDataMessage handler will now pass a key-value pair into the `find_data_end()` method in DevDataAccessor. It then collects that result, and writes it to the EEPROM using the `set_data_boundary()` method in the addresses::DataAddressWrapper singleton.

### `dev_data.hpp`
Inside the DevDataAccessor class, a new `find_data_end(key, length)` method is created. This function simply finds the end of a data entry.

### `addresses.hpp`
A new DataAddressWrapper singleton class was created for the start and end addresses of the OT-Library. to allow for one write in any given runtime. There are a number of reasons for this
1. We don't want anyone accidentally modifying the addresses multiple times
2. We need a way for all classes to be able to see the addresses stored in this class
3. By constraining enforcement scope to runtime, we are allowing for multiple migration events and OT-Library updates

### `types.hpp`
a new 8-bit integer "book_address" type was introduced to prepare for future accessing implementations.

I also removed a bunch of semicolons after the curly braces that signify the end of a namespace since clang kept getting upset.

## Testing
New unit testing file `test_data_update_rev.cpp` included to test that data is properly cutoff at page boundary and that data is properly locked post-cutoff.

## Questions
since the EEPROM writes addresses from back to front, I'm assuming that writes begin at the most significant address  and aend at the least significant. If that's the case, should I change the final data boundary calculation to use a minus instead of a plus to compensate?

## Risk Assessment
Medium. Touching critical machine hardware, but doesn't do anything too extreme just yet.

### DIFF
--- a/eeprom/tests/test_update_data_rev_task.cpp
+++ b/eeprom/tests/test_update_data_rev_task.cpp
@@ -1,0 +1,44 @@
+#include <cstring>
+#include "../../include/eeprom/core/dev_data.hpp"
+
+extern "C" {
+    void vTaskDelay(const int x) {std::ignore = x;}
+    void vTaskDelete(void* x) {std::ignore = x;}
+}
+
+#include "../../include/eeprom/core/types.hpp"
+#include "../../include/eeprom/core/update_data_rev_task.hpp"
+#include "../../include/eeprom/tests/mock_eeprom_task_client.hpp"
+#include "../../stm32-tools/catch2/src/single_include/catch2/catch.hpp"
+
+using namespace eeprom;
+
+SCENARIO("Sending migrate data message") {
+    // migrate data message
+    const uint16_t data_revision = 1;
+    std::vector<std::pair<types::address, types::data_length>> data = {
+        {0, 8},
+        {1, 8},
+        {2, 8},
+    };
+
+    const data_rev_task::MigrateDataMessage mock_data_message{
+        .data_rev = data_revision,
+        .data_table = data
+    };
+
+    // Test handler
+    auto eeprom_client = MockEEPromTaskClient{};
+    auto tail_accessor = eeprom::dev_data::DevDataTailAccessor{eeprom_client};
+    auto data_rev_handler = data_rev_task::UpdateDataRevHandler{
+    eeprom_client, tail_accessor};
+
+    GIVEN("Finding boundary of old data") {
+        data_rev_handler.handle_message(mock_data_message);
+
+        THEN("address should have updated to reflect the new location") {
+            REQUIRE(addresses::ot_library_begin == addresses::data_address_begin);
+            REQUIRE(addresses::ot_library_end == 64);
+        }
+    }
+}
--- a/eeprom/tests/test_update_data_rev_task.cpp
+++ b/eeprom/tests/test_update_data_rev_task.cpp
@@ -1,6 +1,9 @@
 #include <cstdint>
 #include <cstring>
-#include <tuple>
+#include <iostream>
+
+#include "eeprom/core/dev_data.hpp"
+#include "eeprom/core/hardware_iface.hpp"
 
 extern "C" {
 void vTaskDelay(const int x) { std::ignore = x; }
@@ -8,8 +11,6 @@ void vTaskDelete(void* x) { std::ignore = x; }
 }
 
 #include "catch2/catch.hpp"
-#include "eeprom/core/dev_data.hpp"
-#include "eeprom/core/hardware_iface.hpp"
 #include "eeprom/core/types.hpp"
 #include "eeprom/core/update_data_rev_task.hpp"
 
@@ -65,15 +66,12 @@ SCENARIO("Sending migrate data message") {
 
             data_rev_handler.handle_message(mock_data_message);
 
-            // The end address is set to this value because the program should
-       successfully exclude the
-            // final page of data (64 bytes)
             types::address end_address = eeprom_length - 64;
 
             THEN("address should have updated to reflect the new location") {
                 REQUIRE(addresses::DataAddressWrapper::get_ot_library_begin() ==
-                        192); // 192 to reflect the 3 pages now reserved for the
-       lookup table and header REQUIRE(
+                        192);
+                REQUIRE(
                     eeprom::addresses::DataAddressWrapper::get_ot_library_end()
        == end_address);
             }
@@ -92,5 +90,6 @@ SCENARIO("Sending migrate data message") {
             // the value
             THEN("The data address lock should hold") {
                 REQUIRE(addresses::ot_library_end == end_address);
-            }*/
+            }
+}*/
 }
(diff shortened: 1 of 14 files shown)


eeprom/tests/test_update_data_rev_task.cpp
```diff
@@ -0,0 +1,70 @@
+#include <cstring>
+#include "../../include/eeprom/core/dev_data.hpp"
+
+extern "C" {
+    void vTaskDelay(const int x) {std::ignore = x;}
+    void vTaskDelete(void* x) {std::ignore = x;}
+}
+
+#include "../../include/eeprom/core/types.hpp"
```
caila-marashaj: the paths here should be absolute, it should look like `#include “eeprom/core/types.hpp”` for all these

eeprom/tests/test_update_data_rev_task.cpp
```diff
@@ -0,0 +1,70 @@
...
+    auto eeprom_client = MockEEPromTaskClient{};
+    auto tail_accessor = eeprom::dev_data::DevDataTailAccessor{eeprom_client};
+    auto data_rev_handler = data_rev_task::UpdateDataRevHandler{
+    eeprom_client, tail_accessor};
+
+    GIVEN("Finding boundary of old data") {
+        data_rev_handler.handle_message(mock_data_message);
+
+        THEN("address should have updated to reflect the new location") {
+            REQUIRE(addresses::ot_library_begin == addresses::data_address_begin);
```
caila-marashaj: On top of testing functionality, these tests should sort of be able to serve as a guide on what the feature is doing and why. Here it isn’t quite clear why updating the data rev should make `ot_library_begin` be set to `data_address_begin`, or why the `library_end` location should be 64. You could use just setting 64 to a global variable name to explain that.

include/common/core/message_utils.hpp
```diff
@@ -73,4 +73,4 @@ struct VariantCat<std::variant<VariantAContents...>,
     using type = std::variant<VariantAContents..., VariantBContents...>;
 };
 
-};  // namespace utils
+}  // namespace utils
```
caila-marashaj, line 76: not sure if this was intentional

include/eeprom/core/addresses.hpp
```diff
@@ -1,103 +1,167 @@
...
+ * When a client sends a get_data request it will provide [key]. As with the
+ * write it should also overload with [key, length] to read the first ‘length’
+ * bytes and [key, length, offset] to read ‘length’ bytes offset by ‘offset’
+ * bytes from the beginning of the value
+ *
+ * When a read or write operation is performed, first the driver will read the
+ * value memory address at  lookup_table[2*sizeof(address)*key] and the data
+ * length from lookup_table[(2*sizeof(address)*key) + sizeof(address)]. If the
+ * operation attempts to read a uninitialized key ie when
+ * ((2*sizeof(address)*key) > data_table_tail) it will call the callback with an
```
caila-marashaj, line 72: great docstring!

munanwugo-opentrons: I don't know why it said I edited that, that's been there the whole time lol

include/eeprom/core/addresses.hpp
```diff
@@ -1,103 +1,167 @@
...
+        // Basicalsy _boundary_address = value
+        //            boundary_address = header location that contains _boundary_address
+        static void set_data_boundary(types::address _boundary_address,
+            auto& eeprom_client) {
+            // if either these have been written to, don't do anything
+            if (_ot_library_end || _ot_library_begin) {
+                return;
+            }
+
+            if (_boundary_address % 64 != 0) {
```
caila-marashaj: lint might yell at you here for having a ‘magic number’. Probably just assign this to a global variable at the top

eeprom/tests/test_update_data_rev_task.cpp
munanwugo-opentrons: Would you say that I should do this in the code itself (like using the actual C++), or more in the in-line documentation in comments and stuff?

include/eeprom/core/addresses.hpp
```diff
@@ -1,103 +1,167 @@
...
+
+        //sets the ot_library boundary with old data
+        // FYI: this _boundary_address is the actual address that will be stored at the location of the
+        // above boundary_address
+        // Basicalsy _boundary_address = value
+        //            boundary_address = header location that contains _boundary_address
+        static void set_data_boundary(types::address _boundary_address,
+            auto& eeprom_client) {
+            // if either these have been written to, don't do anything
+            if (_ot_library_end || _ot_library_begin) {
```
caila-marashaj: you may need to explicitly check whether `ot_libraary_end == 0` or `ot_library_begin == 0`, otherwise you might accidentally always evaluate `true` even if `ot_library_begin` and `ot_library_end` are 0 and not just `null`

include/eeprom/core/addresses.hpp
```diff
@@ -1,103 +1,167 @@
...
+            message::WriteEepromMessage write;
+            write.memory_address = boundary_address_begin;
+            write.length = boundary_address_length;
+
+
+            auto* write_iter = write.data.begin();
+            write_iter = bit_utils::int_to_bytes(
+                _boundary_address, write_iter, write_iter + write.length);
+
+            eeprom_client.send_eeprom_queue(write);
```
caila-marashaj: is this meant to be `write_iter`? if not, do you use `write_iter`?

include/sensors/core/mmr920.hpp
```diff
@@ -369,5 +369,5 @@ using RegisterSerializedType = uint8_t;
 // Command Registers are all 8 bits
 // Type definition to allow type aliasing for pointer dereferencing
 using RegisterSerializedTypeA = __attribute__((__may_alias__)) uint8_t;
-};  // namespace mmr920
```
caila-marashaj, line 372: there’s a bunch of these they don’t quite look intentional

caila-marashaj (changes requested):
Left a couple comments, pretty small things; as for naming this commit, what goes inside the parentheses would be the subsystem that you’re working on, rather than the name of the feature. So, this would be `feat(eeprom): …`


Aside from that, really great job. This pr is no joke!

include/common/core/message_utils.hpp
munanwugo-opentrons: yeah it was, my compiler kept shouting at me because of it

include/eeprom/core/addresses.hpp
munanwugo-opentrons: I used it because that's the pattern that I noticed accross the rest of the EEPROM stack for write iterators.

caila-marashaj (approved):
Nice work!

PR #839, opened 2026-03-19
feat(wear leveling): Cyclic Reduncancy Check for EEPROM data

munanwugo-opentrons:
# Overview

One of the core features of the OT-Library is the Cyclic Redundancy Check (CRC). This is a widely-used form of error checking and correction - though the OT-Library will only worry about the checking and not the correction. The plan is to include this in the header of every page as a means of verifying whether the data being read is accurate or not. If it isn't, the data will be invalidated and the system will simply find the last valid data partition and continue from there. This PR includes the logic for calculating the CRC, as well as the BookAccessor scaffolding. All of the actual checks will be implemented in the read/write portions of execution.

#### How it works:
CRC in general works by converting bits into a polynomial expression, then performing polynomial division over a "generator polynomial" to get a remainder that should be effectively unique to the data being stored (see this (Ben Eater)[https://www.youtube.com/watch?v=izG7qT0EpBw&t=1864s] video for more). My approach to this is as follows:
1. Convert byte array into a data bitset for easier bit manipulation
2. Left shift data to accommodate CRC
3. Step through each bit of the data bitset, performing polynomial division to reduce to final remainder
4. Convert final remainder back into byte array
5. return final byte array as CRC

### `book_accessor.hpp`
1. Creation of PageType struct: Provides a standardized way of organizing metadata for future read/writes.
6. BookAccessor scaffolding: Beginnings of the BookAccessor class. 3 functionalities: Read, Write, Create New Data Part
7. CRC implementation for future read/write implementations
8. Helper methods for converting byte arrays to bitsets and back

### `update_data_rev_task.hpp`
Imported book_accessor.hpp for compilation purposes and in preparation for final stage of migration (once BookAccessor is complete).

## Review requests

Does the logic work out? I'm not too used to substituting math for logic gates. I would just like to make sure that there aren't any super blatant gaps that I'm missing.

## Risk assessment

High. Deals with matters of data integrity. Faults here may result in over 2x the intended number of writes, negating the benefits of the wear-leveling solution.


sfoster1 (changes requested):
this chip actually has a hardware crc accelerator, which we should use - since we're reading and writing this data through i2c, we can dma it through the crc module to get this for free. see the g4 reference manual, section 16, and the hal code here: https://github.com/STMicroelectronics/stm32g4xx-hal-driver/blob/7186efe25f233560d6859c18b1c12f4e821d2ba7/Inc/stm32g4xx_hal_crc.h

munanwugo-opentrons:
> this chip actually has a hardware crc accelerator, which we should use - since we're reading and writing this data through i2c, we can dma it through the crc module to get this for free. see the g4 reference manual, section 16, and the hal code here: https://github.com/STMicroelectronics/stm32g4xx-hal-driver/blob/7186efe25f233560d6859c18b1c12f4e821d2ba7/Inc/stm32g4xx_hal_crc.h

I understand that it would be optimal to use the hardware accelerator for CRC, but the end of this project's intended lifetime is fast approaching. I would prefer to be able to get something that works and then come back and clean this up, especially since this is a critical part of the solution's design.

ryanthecoder (approved):
I think this looks good if we want to make the hardware-accelerated crc a separate PR so we can move on with the next part of the planned implantation and start testing something sooner rather than later.

sfoster1 (approved):
Okay, we can merge this before we do the hardware crc thing, but we definitely need to do that next.

munanwugo-opentrons:
> Okay, we can merge this before we do the hardware crc thing, but we definitely need to do that next.

Sounds good! I'll start working on it now!

PR #840, opened 2026-03-20
fix(state-manager): fix broken CI dependency for Pyro5

CaseyBatten:
Adding Pyro5 coverage on hardware modules to the monorepo with [this PR](https://github.com/Opentrons/opentrons/pull/21015) caused a dependency failure over in ot3-firmware during CI. This adds Pyro5 and the related serializer Serpent dependencies to the lock file.


sfoster1 (changes requested):
I don't think you should need to add pyro here, right? it should be a transitive dependency of the opentrons module, so relocking should be all that's needed?

sfoster1 (approved):
thank you!

PR #843, opened 2026-04-13
feat(wear-leveling): EEprom Reads

munanwugo-opentrons:
# Overview

This PR contains the new "read" mechanism for the EEproms. As a part of the wear-leveling design, reads will now read a full book (4 pages) of memory, and then select the most recent one that has a valid CRC to return to the user. 

3 test cases:
1. Valid read - Where the most recent page also has a valid CRC
2. Invalid read - Where the most recent page has an invalid CRC, but the next most recent page does (this is the page that will be selected)
3. All invalid reads - Where none of the pages have valid CRCs, in which case, the read returns all 0s to the user

### `eeprom/types`
New types to represent commonly used values (such as page size and book data size).

### `eeprom/messages`
New message types to support OT Library operations, with options for a full page and a full book.

### `eeprom/accessor`
added new methods to handle new message types

### `eeprom/task`
added new methods to handle new message types

### `eeprom/book_accessor`
Added Read functionality. The read flow is as follows:
1. user calls get_data() with relevant parameters
2. get data initiates a read to the data table to determine the book's starting address
3. table_action_callback() is called back and initiates a read to the book's address by calling accessors OT_start_read_at_offset method
4. accessor performs the read and moves the information into the book_accessor's buffer, then it calls read_complete on the book_accessor
5. read_complete paritions moves the data in to a temporary array of pages and calls read_final()
6. read_final determines the most recent valid page and moves its contents to the user's buffer

### `i2c/messages`
change the max i2c buffer size to perform larger reads

### `pipettes/tasks/move_group`
shrink sizes to give pipette more RAM to work with

## Review requests

I had an issue with segfaults while developing since I'm not used to manual memory management. Are there any other places where I'm managing memory incorrectly?

## Risk assessment

Medium


include/eeprom/core/accessor.hpp
```diff
@@ -181,6 +214,22 @@ class EEPromAccessor {
         }
     }
 
+    void OT_callback(const eeprom::message::EepromDataMessage& msg) {
+        std::visit(
```
caila-marashaj, line 218: why make this a lambda function?

include/eeprom/core/accessor.hpp
```diff
@@ -197,6 +246,20 @@ class EEPromAccessor {
...
     size_t bytes_recieved = 0;
     size_t bytes_to_read = 0;
+
+    /**
+     * Handle a completed read.
+     * @param msg The message
+     * @param param This pointer.
+     */
+    static void OT_callback(const eeprom::message::EepromDataMessage& msg,
+                            void* param) {
```
caila-marashaj, line 256: is `param` here meant to be the current `EEPromAccessor` object?

ryanthecoder (approved):
looks good. big chunks of this get replaced in further prs so merge this in so we can address things in the later prs.

PR #846, opened 2026-04-20
fix(cmake): Fix CI tests

ryanthecoder:
State manager doesn't like edge anymore so i locked it into 8.5.1 cause it was previously locked to 8.5.0a2 but the github action was checking out edge. 

Updated poetry for state manager

don't use new clang because it just creates a ton of issues that need to be gone through in a more elaborate update.

update some flags for codecov since it updated to gcc 13 based codecov
also added the codecov token to github so it actually works now.

don't force a gcc10 host build anymore, causes a bunch of other problems.

The new fetch content rule breaks downloading the correct version of clang. 
Set the old rule and we should go through and TODO upgrading compilers/clang etc.
https://cmake.org/cmake/help/latest/policy/CMP0169.html


vegano1 (approved):
🙏 Thank you!

PR #849, opened 2026-06-03
feat(wear-leveling): Hardware compatability

munanwugo-opentrons:
# Overview

This PR contains the trigger to migrate EEPROM data from the non-wear-leveled system to the wear-leveled one on the pipettes, gantry, head, gripper, and rear panel. It also allows all of those parts of the robot to talk to the EEPROM via the book_accessor to incorporate wear-leveling into both reads and writes. 

Please note:
Although the 96 channel pipette's eeprom migrates properly and has access to the accessor to update its EEPROM, RAM constraints mean that i2c message sizes weren't able to be increased large enough to accomodate messages coming in from the sensors and the EEPROM. No information will be written to the 96 channel's EEPROM after migration until RAM constraints are alleviated.

Tests from previous PRs were modified to provide a more accurate eeprom mock, as well as more realistic tests.

Hardware testing was done on Mrs. Met using the following protocol:
[test_eeprom.py](https://github.com/user-attachments/files/28901285/test_eeprom.py)

The length of this protocol was modified by alternating the number of iterations in the loop between 1, 10, and 100. After each run, the eeprom was dumped to verify that the result was expected. 

## Changelog
Changes in this PR fall into 3 major categories.
### Tests
1. `test_book_accessor.cpp` was overhauled to better resemble how the EEPROM works. Reads and Writes are now performed directly to the "accessor backing", and modified after the fact to trigger certain behaviours.
2. Message sizes in other tests files were changed to reflect the change in i2c message sizes. 
3. A "shadow" crc calculation was added. This is because the standard CRC uses hardware acceleration, which breaks tests.

### Trigger
1. Migration messsages were added to the `eeprom_keys.cpp` file in the head, gripper, gantry, and pipette. These messages trigger migration.

### Utility
1. Unnecessary methods were removed from `accessor.hpp`
2. the BookAccessor now takes in `all_reads` as a constructor argument to better handle concurrency issues.
3. Lengths are now stored in the book header. They are used to aid in CRC calculations.
4. BookAccessor now caches reads, allowing writes to bypass heavy reads if the relevant information already exists.
5. Unnecessary messages were removed

## Risk assessment

high


gantry/simulator/crc16.cpp
```diff
@@ -0,0 +1,61 @@
+#include <bitset>
+#include <cstdint>
+#include <variant>
+#include <vector>
+
+extern "C" {
+__attribute__((weak)) void crc16_init(void) {}
```
sfoster1: why do these need to be `__attribute__((weak))`?

gripper/firmware/main_rev1.cpp
```diff
@@ -69,6 +70,7 @@ static constexpr auto eeprom_chip =
 #else
 static constexpr auto eeprom_chip =
     eeprom::hardware_iface::EEPromChipType::ST_M24128_BF;
+// static eeprom::CRC16Accelerated crc16;
```
sfoster1: remove commented code

gantry/simulator/crc16.cpp
sfoster1: also there should be just one of these, in the eeprom simulator tree, to avoid having it in multiple places throughout the repo

head/core/tasks_rev1.cpp
```diff
@@ -269,7 +269,9 @@ void head_tasks::MotorQueueClient::send_move_status_reporter_queue(
 void head_tasks::HeadQueueClient::send_eeprom_queue(
     const eeprom::task::TaskMessage& m) {
 #if PCBA_PRIMARY_REVISION != 'b'
+
```
sfoster1: let's get rid of this, it just adds noise to the diff

include/eeprom/core/update_data_rev_task.hpp
```diff
@@ -72,55 +97,84 @@ class UpdateDataRevHandler : accessor::ReadListener {
...
-        // update the "tail" of ot_library (through ot_library_end address)
-        // to end of previous data
-        addresses::DataAddressWrapper::set_data_boundary(
-            table_creator.find_data_end(key, length), eeprom_client);
+    void migrate_message_helper() {
+        // create a new table entry at the same key
+        book_table_creator.create_data_part(key, length, accessor_backing,
+                                            true);
+
+        // while (!table_creator.table_ready()) {
```
sfoster1: remove commented code

include/eeprom/core/update_data_rev_task.hpp
```diff
@@ -72,55 +97,84 @@ class UpdateDataRevHandler : accessor::ReadListener {
...
+            current_data_rev, data_rev_backing.begin(), data_rev_backing.end());
+        data_rev_accessor.write(data_rev_backing, 0);
+
+        migrating = false;
     }
 
+    // method to set boundary of ot_library... not currently necessary, but may
+    // be useful in the future
+    //
+    // void set_ot_library_boundary(MigrateDataMessage&
```
sfoster1: remove commented code

sfoster1 (changes requested):
will have more to look at later but for now let's get rid of that commented out code and dedupe the crc c++ implementation

gantry/simulator/crc16.cpp
munanwugo-opentrons: Well I did it originally to account for the possibility that the real CRC functions and these fakes could exist at the same time, but come to think I don't think that would actually happen. 

Also putting this file in the eeprom simulator tree does make a lot more sense. Fixing now.

include/eeprom/core/update_data_rev_task.hpp
```diff
@@ -72,24 +97,43 @@ class UpdateDataRevHandler : accessor::ReadListener {
...
-    void visit(MigrateDataMessage& m) {
+    void visit(const MigrateDataMessage& m) {
+        // "finish data rev" before data rev is finished for migration to ensure
+        // that reads can properly occur
+        tail_accessor.finish_data_rev();
+
         if (m.data_rev == current_data_rev + 1) {
-            // set_ot_library_boundary(m);
-            //  TODO: Make an OTLibraryAccessor
+            migrating = true;
```
vegano1, line 111: I think this might be an issue

The `visit(const MigrateDataMessage&)` calls `tail_accessor.finish_data_rev()` unconditionally above before checking `m.data_rev == current_data_rev + 1`. The first migrate message in `table_updater` sets `data_rev_finished` globally while migrations are still in flight. `BookAccessor::read_write_ready()` then becomes true for `UsageStorageTaskHandler` and other accessors, allowing concurrent EEPROM traffic during migration.

Should we only call `finish_data_rev()` after all migration/update steps succeed, maybe at the end of `UpdateDataRevTask::operator()`?

include/eeprom/core/update_data_rev_task.hpp
```diff
@@ -72,24 +97,43 @@ class UpdateDataRevHandler : accessor::ReadListener {
...
-            data_rev_accessor.write(data_rev_backing, 0);
-            current_data_rev = m.data_rev;
+            // get data that was previously at key
+            table_creator.get_data(key, length, 0);
         }
     }
 
+    void migrate_message_helper() {
+        // create a new table entry at the same key
+        book_table_creator.create_data_part(key, length, accessor_backing,
```
vegano1, line 127: The `migrate_message_helper()` bumps `current_data_rev`, writes the data-revision to EEPROM, and clears `migrating` immediately after queueing `book_table_creator.create_data_part(...)`.

Since it does not wait for the asynchronous book-format write to finish, `busy_migrating()` returns false while migration I/O may still be pending. This might be VERY rare, but a power loss AFTER the rev bump but BEFORE the book write leaves the device marked as migrated with old-format data.

Maybe we can defer rev persistence and `migrating = false` until `book_table_creator` signals completion, maybe with a done callback. Then have `UpdateDataRevTask` wait on that completion, not only on the legacy read phase.

eeprom/firmware/crc16.c
```diff
@@ -0,0 +1,66 @@
...
+void crc16_init() { MX_CRC_Init(); }
+
+/**
+ * Compute the CRC
+ * @param data Data
+ * @param length Length of data
+ * @return Computed CRC
+ */
+uint16_t crc16_compute(const uint8_t* data, uint8_t length) {
+    return (uint16_t)~HAL_CRC_Calculate(&hcrc, (uint32_t*)data, length);
```
vegano1: The `HAL_CRC_Calculate` is fed via `(uint32_t*)data` without alignment guarantees. Payload buffers are often 1-byte aligned; on Cortex-M, unaligned 32-bit access can fault or produce incorrect CRCs. Is there any way we can make sure this is aligned before passing it down to `HAL_CRC_Calculate`?

include/eeprom/core/update_data_rev_task.hpp
```diff
@@ -129,25 +177,38 @@ class UpdateDataRevHandler : accessor::ReadListener {
...
+        // 2. during a migration read to read the data table for a key
+        if (migrating) {
+            migrate_message_helper();
+        } else {
+            if (std::equal(delivery_state.begin(), delivery_state.end(),
+                           data_rev_backing.begin())) {
+                data_rev_backing.fill(0x00);
+                data_rev_accessor.write(data_rev_backing, 0);
+            }
+            std::ignore = bit_utils::bytes_to_int(data_rev_backing.begin(),
```
vegano1, line 199: The `book_table_creator` and `table_creator` both register `UpdateDataRevHandler` as their `ReadListener`. Which means, any `BookAccessor` read completion invokes `UpdateDataRevHandler::read_complete()`. This sets `ready_for_new_message = true` even when the completion is unrelated to migration.

Maybe we can use separate listener objects per accessor?

eeprom/firmware/crc16.c
```diff
@@ -0,0 +1,66 @@
...
+ * @return Accumulated CRC
+ */
+uint16_t crc16_accumulate(const uint8_t* data, uint8_t length) {
+    return ~HAL_CRC_Accumulate(&hcrc, (uint32_t*)data, length);
+}
+
+/**
+ * Reset the accumulated CRC value.
+ */
+void crc16_reset_accumulator() { /*__HAL_CRC_DR_RESET(&hcrc);*/ }
```
vegano1: remove unused accumulate API

include/eeprom/core/book_accessor.hpp
```diff
@@ -250,77 +314,117 @@ class BookAccessor
...
+    std::array<std::array<uint8_t, types::page_length>, 4>& all_reads;
+    // cache the most recent key read to bypass reads wherever necesasry
+    int16_t cached_key = -1;
 
     template <size_t num_bytes>
-    auto calc_crc(std::array<uint8_t, num_bytes> data)
-        -> std::array<uint8_t, 2> {
+    auto calc_crc(std::array<uint8_t, num_bytes> data) -> uint16_t {
+        crc16_init();
+
```
vegano1, line 367: `crc16_init()` gets called every-time we call `calc_crc`, cant we initialize this once at boot? Maybe after creating the BookAccessor instance?

include/i2c/core/messages.hpp
```diff
@@ -9,7 +9,7 @@ namespace i2c {
 namespace messages {
 
 using std::size_t;
-static constexpr std::size_t MAX_BUFFER_SIZE = 16;
+static constexpr std::size_t MAX_BUFFER_SIZE = 10;
```
vegano1, line 12: This seems unrelated to the wear-leveling work. Could we move this to a separate PR?
Also, this may truncate I2C transactions on some boards, like the 96-channel.

include/eeprom/core/update_data_rev_task.hpp
```diff
@@ -72,24 +97,43 @@ class UpdateDataRevHandler : accessor::ReadListener {
...
+            length = m.data_table[0].second;
+            intermediate_data_rev = m.data_rev;
 
-            /*for (const auto& i : m.data_table) {
-                // TODO: add a table_creator method to migrate data
-                // 1. get value here
-                // 2. move to same index in new_location
-            }*/
+            // reset the accessor backing to make sure it's empty for the next
+            // read
```
vegano1, line 117: Could this vector be empty, and the migration indexes `m.data_table[0]` be checked without checking for an empty vector? Maybe we can guard with `if (m.data_table.empty()) { return; }` before indexing.

vegano1 (changes requested):
This is solid and should reduce wear-leveling significantly. Here are a few things I found that we should at least consider before moving forward.

include/i2c/core/messages.hpp
munanwugo-opentrons: 10 was the original value of the max_buffer_size. I increased it to 16 and then brought it back down since it wasn't compiling on the 96-channel. 

The max buffer size here was actually pretty important to the wear-leveling since each write now sends a lot more information that it used to, and we had an issue where a bunch of messages simply wouldn't make it through - or would partially make it through. This would have a cascading effect resulting in the CRC miscalculating the value stored on the EEPROM and writing the error code for worn pages into the read buffer.

In any case, this was the original value, and other changes were made to have the system fit neatly into it.

include/eeprom/core/update_data_rev_task.hpp
munanwugo-opentrons: Hmm... This is a real issue. The main reason I did this was because the DevDataAccessor refuses to read anything if the data rev isn't finished, so we can't actually receive any of that information. I think I might just make a seperate tail accessor whose data rev won't be finished until the migration is complete.

include/eeprom/core/update_data_rev_task.hpp
munanwugo-opentrons: I'm not entirely sure I understand.

the busy_migrating() function was created because the task would send so many migrate messages that not all of them would even make it to the "create" point of migration, and sometimes would migrate out of order. It was just meant to slow things down a bit.

It is true that migration isn't really fully done until the data is written back to the EEPROM. Power loss after the rev bump but before the book write is something that Ryan and I did consider and decided that it was rare enough that it would be okay - especially since most people know that turning of power during an update is a bad idea - but I'm definitely open to changing it.

The thing I don't understand is, even if migration isn't marked as finished in the `UpdateDataRevHandler` until it's been fully written to the eeprom, wouldn't the problem still persist? If power is lost after the data rev and before the data is written, when power is regained this flow is restarted. The system reads the data rev from the EEPROM to determine which one is the current one, so it would still think that that data revision occured, wouldn't it?

If the power loss issue is a concern, then wouldn't the solution be something along the lines of not updating the data rev itself until the data creation is completed? And if that's the case, then shouldn't that be the standard applied across all data creation, not just the migration?

eeprom/firmware/crc16.c
munanwugo-opentrons: Yes there is, and it was actually in there before. I'll add it back in.

include/eeprom/core/update_data_rev_task.hpp
munanwugo-opentrons: Data creation does not invoke reads, and update data rev task never asks for a read. The only reads that come through this read_complete are those from the legacy system. Would this be necessary?

include/eeprom/core/book_accessor.hpp
munanwugo-opentrons: I'm not sure why but for some reason when I tried that approach it didn't work. The CRC would return one value the first time, then a completely different value every other time.

include/eeprom/core/update_data_rev_task.hpp
munanwugo-opentrons: The vectors realistically should never be empty, especially since they are manually populated between firmware updates and not generated dynamically. Though it is a thing that can happen. Adding this.

include/eeprom/core/book_accessor.hpp
vegano1: hmm got it, not sure tbh

include/eeprom/core/update_data_rev_task.hpp
vegano1: no, just a thought

include/eeprom/core/update_data_rev_task.hpp
vegano1: yeah i agree, this is rare enough that honestly i would not worry about it.

vegano1 (approved):
Thanks for addressing the feedback, some of the issues i brung up are rare enough that we arent going to worry about them.

sfoster1 (approved):
ok, looks good to me, nice!

PR #851, opened 2026-06-12
fix(wear-leveling): Downgrades

munanwugo-opentrons:
#Overview
In order to support downgrades, the decision was made to make the ot library lookup table seperate from that of the old lookup table.

Tested using the following protocol on Mrs Met:
[test_eeprom.py](https://github.com/user-attachments/files/28901285/test_eeprom.py)

Made sure that what is written to the EEPROM is correct.

## Changelog
1. Changed location of data table
2. enlarged i2c messages (fixes regression)

## Risk assessment

low


vegano1 (approved):
This looks good!

